PR #15, opened 2014-10-23
Materialized path key traversal.

jacobstr:
Allows deep access of config settings via a dot-delimited materialized path E.g.

  viper.GetString("clothing.jacket") => "leather"

Environment variables may also be set to corresponding deep keys using double
underscores in place of the usual period delimiter. E.g:

  clothing.jacket => CLOTHING__JACKET


spf13:
closing in favor of #54.

PR #16, opened 2014-10-24
Attempt to resolve issue 579.

justin-calleja:
(Edit: it's issue 579 of hugo project)

Config file is not being picked up when running `hugo server --watch`
from within the hugo project directory (the site).

I know this is intended to solve an issue with hugo (I see it as an issue anyway)... but isn't it a good idea to have  "." in the configPaths?

I am not familiar with either Go, Hugo, or Viper so most likely there is a better solution.

Regards,
Justin

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -39,7 +39,7 @@ import (
 )
 
 // A set of paths to look for the config file in
-var configPaths []string
+var configPaths []string = []string{"."}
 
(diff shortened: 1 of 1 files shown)


spf13:
This should be set on an app specific basis. I believe it's been fixed in Hugo since.

PR #43, opened 2015-02-06
Support for cascading configuration files

wrobbins:
Added the ability to have sparse configuration files using cascading. See readme and viper_tests for more information and usage.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@
 # Folders
 _obj
 _test
+.idea
 
 # Architecture specific extensions/prefixes
 *.[568vq]
--- a/README.md
+++ b/README.md
@@ -86,7 +86,7 @@ the order that files are added  by AddConfigPath. For more see viper_test's casc
 Consider:
 
  * \etc\myapp\myapp.json
- * (%GOPATH)\src\myapp\myapp.json
+ * ($GOPATH)\src\myapp\myapp.json
 
 You can check in a default myapp.json for development and only override certain kvps in production
 
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-viper <!--[![Build Status](https://travis-ci.org/spf13/viper.svg)](https://travis-ci.org/spf13/viper)-->
+viper 
 =====
 
 Go configuration with fangs
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-viper [![Build Status](https://travis-ci.org/spf13/viper.svg)](https://travis-ci.org/spf13/viper)
+viper <!--[![Build Status](https://travis-ci.org/spf13/viper.svg)](https://travis-ci.org/spf13/viper)-->
 =====
 
 Go configuration with fangs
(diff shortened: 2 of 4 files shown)


spf13:
I love this contribution. Do later read files overwrite the earlier ones? That wasn't clear to me from the docs and I think that's what the source says, but didn't read it super carefully.

wrobbins:
Configuration is read from highest to lowest priority keys/values found in higher priority paths will override key/values in lower priority paths, if I understand your question.

Eg: this configuration setup:
    /tmp/high/test.yaml
    foobar: high
---
    /tmp/low/test.yaml
    foobar: low

The value of foobar will be "high" when /tmp/high/ is a higher priority path than /tmp/low

spf13:
@wrobbins how do you set the priority for a given path?

Also this is awesome. Could I get you to merge it up with head again and repush. I'm excited to get this into Viper. I think a lot of people will benefit from it.

bketelsen:
didn't look at the PR, does it also work with RemoteConfig?  This would be very valuable there too.  If not, I can look at extrapolating this work to the remote config providers to provide the same functionality.  Nice!

spf13:
It would be great if it worked with remote as well.

spf13:
@wrobbins Would you be able to merge this up with HEAD? I'd love to get it in.

wrobbins:
Been swamped - will try this week.

nishp1:
+1 for cascading.

spf13:
@wrobbins ping. Let me know if you have time to do this. I will give it a shot next week if you haven't gotten to it. 

I'm also curious how this will work with #54 which @kzvezdarov is working on.

wrobbins:
Looking like it will take a bit of time to merge in the current master so this fork is up to date for merge into the main baseline. 

Took a brief look at #54 (and #63). As of right now cascading won't work for nested configurations at all as it increases the amount of traversals needed. Nesting may help this but would need to be re-integrated into cascading.

To put it another way: cascading's current implementation supports sparse configurations only at the first level of the configuration hierarchy. It expects dense configurations when they are nested. If a key representing a dictionary isn't available at a higher priority then it will cascade down to find that key. If the dictionary is available but a key *inside* the dictionary isn't available then it won't attempt to cascade down to other configurations to find it.

kzvezdarov:
@wrobbins The nested search can be made to fall through to the lower priority levels for subkeys, but I feel like this is very undesirable. Currently it will fall through for the top-level key as usual.

Isn't cascading done during readin in of remote/local configs? My impression was that it produces a flattened config/key_value registry that contains union of the multiple files/sources.

endophage:
@spf13 I'm guessing this was the PR you were thinking about. Is there anything I can do to help get it merged?

spf13:
@endophage This is indeed the PR we discussed. I love what it's intent is. Unfortunately in it's current state it's not mergeable. It needs to be adjusted to work within the added support of nested keys as added in PR#63.

MattSurabian:
I tried rebasing this branch back into master with mixed results. If @wrobbins isn't planning to update this PR it might be easiest to rewrite into the latest version and open a new one.

Is that work that's currently underway?

endophage:
@MattSurabian I'm scheduling time to work on this as I need the feature for my next release. I haven't had a look at whether this PR can be updated or if work needs to start from scratch. Unless you need the feature sooner than the next 6 weeks, I'll dig in and work out how I want to proceed.

MattSurabian:
Cool, @endophage! I don't need it within 6 weeks, I was just starting to take a pass at this myself and didn't want to duplicate effort.

One thing I'd add is that I don't completely buy into the approach taken in this PR as it's introducing additional variables that I'm not sure are necessary. Though admittedly the code has changed quite a bit since this was first opened.

Regarding making it work with remotes my thought was to break out methods for `findInConfig` and `findInKV` where cascade behavior could be implemented: https://github.com/MattSurabian/viper/commit/0d0feebe45c04d5b784c9b8785ea9dbbf73599db Not sure if that makes sense from your perspective. But it seems an easy way to ensure existing precedence remained unchanged.

I'd love to be of help with this feature if possible as it's something I'd be very interested in using too. Let me know if there's anything I can do and whether you think it's worth me continuing to hack on it.

lightglitch:
What it's the status of this PR? It would be really useful for my project.

geoah:
+1 This is actually pretty awesome. Is this scheduled for merging at some point?

jojomi:
I would like to see this in viper too :+1:

four2five:
Jumping on the "would love to have this" bandwagon

DaBlitzStein:
Is this integrated?

endophage:
Not as far as I know. Also my apologies for not getting around to tackling it. This kept getting bumped down my priorities list and with the time that has passed, I can't really make a claim it's as useful to me as I originally thought. Not having it hasn't majorly impacted the project I would have used it for. Therefore I'm probably not going to find the time to look in to it :-/

PR #44, opened 2015-02-17
Allow Marshal to use env and flags

kzvezdarov:
This PR fixes the following issues:

1. Currently `Marshal` ignores `env` variables and `pflags` when marshaling the config data into a structure.

2. Currently `Marshal` uses the wrong order of precedence for the various config sources. The documentation states that it is defaults<k/v store<config<env<flags<overrides, but `Marshal` uses `defaults<config<k/v store<override`

Solution:

1. `Marshal` now uses the `AllSettings` method in order to get the appropriately overriden settings data.

2. Added `env` and `pflag` key checks in the `AllKeys` method in order to add their data to the return of `AllSettings` (which uses the keys to `Get` each value).

3. Switched to `WeakDecode` for marshaling the settings into the structure. This is necessary since `pflags`'values are always strings, even for numeric types. `WeakDecode` coerces those to the appropriate types for the corresponding fields.

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -313,23 +313,10 @@ func (v *viper) MarshalKey(key string, rawVal interface{}) error {
 // Marshals the config into a Struct
 func Marshal(rawVal interface{}) error { return v.Marshal(rawVal) }
 func (v *viper) Marshal(rawVal interface{}) error {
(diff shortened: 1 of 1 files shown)


spf13:
This is a much better approach. Thanks!

Merged as 2e2f3b2

PR #45, opened 2015-02-17
Exported the Viper type in order to better support multiple Vipers

kzvezdarov:
This PR exports the `Viper` type to make multiple vipers more convenient.

Specific use case that necessitated this:

1. Passing the viper config down a line of function calls, basically using it as a context.

2. Dynamically generating new configs for sub commands - such that I can automate it. E.g. I can store a mapping of all vipers to command names, and automatically create a new viper when a new command  is added, enclosing that command's configuration scope.

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -37,7 +37,7 @@ import (
 	crypt "github.com/xordataexchange/crypt/config"
 )
 
-var v *viper
+var v *Viper
 
 func init() {
 	v = New()
@@ -61,9 +61,9 @@ func (rce RemoteConfigError) Error() string {
 	return fmt.Sprintf("Remote Configurations Error: %s", string(rce))
 }
 
-// A viper is a unexported struct. Use New() to create a new instance of viper
+// A Viper is a unexported struct. Use New() to create a new instance of viper
 // or use the functions for a "global instance"
-type viper struct {
+type Viper struct {
 	// A set of paths to look for the config file in
 	configPaths []string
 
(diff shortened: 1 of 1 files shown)


spf13:
I'm curious about your use case. Do you have code I can see doing this?

kzvezdarov:
Certainly. I am writing a (rather simple) utility that ingests data from the Dota 2 web API. The configuration can be seen in the `main` module here:
https://github.com/kzvezdarov/trencher/blob/master/trencher.go

Making a call to a specific endpoint with specific parameters is done by issuing the appropriate subcommand and giving it any combination of config files and command line flags.
The config data is used to build the query by marshaling the config into the query struct (which relates to my other PR #44 ).
By having one viper per command per query type things remain clean and simple. I like the fact that this way I can isolate each command's configuration context and avoid any accidental spillage or other problems (even though that should not be possible, technically).
Lastly, I will need to eventually pass the configuration down the tree of functions. This is because I want to support changes in the data store resource - as in, in case one data store instance goes down and the process needs to dump its data to another.

ethervoid:
Another use case is to inject the configuration to your components as a dependency to avoid create it in every step of your application. For example:

```
func loadConfig() *viper.Viper {
    config := viper.New()
    config.SetConfigName("config_file")
    config.AddConfigPath("/etc/app")
    config.AddConfigPath("config")
    err := config.ReadInConfig()
    if err != nil {
        log.Fatal("Can't find the configuration")
    }

    return config
}

func main() {
    config := loadConfig()
    component := Component{config}
    component2 := Component2{config}
}
```

spf13:
merged as 700eefa

spf13:
Thanks for this guys. I'm glad that you pushed for it. I think it will make Viper a lot easier for a lot of people to use.

PR #47, opened 2015-02-19
Add viper.GetDuration

ecnahc515:
Depends on spf13/cast#10

Would be really handy for those of us who deal with duration's a lot.


divan:
Would love having this as well.

ecnahc515:
Merged as ededa04e0bbab359f4dfd31352e39d7c8b200950

PR #48, opened 2015-02-19
Support rewriting envvar keys

ecnahc515:
Use case is replacing `-` in my command line/viper keys with `_` when retrieving env vars.

```
viper.AutomaticEnv()
viper.SetEnvKeyReplacer(strings.NewReplacer("-", "_"))
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@ Go configuration with fangs
 
 Viper is a complete configuration solution for go applications. It has
 been designed to work within an application to handle all types of
-configuration. It supports 
+configuration. It supports
 
 * setting defaults
 * reading from yaml, toml and json config files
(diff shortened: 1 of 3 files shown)


ecnahc515:
@spf13 How would you feel about this?

spf13:
Hoping to review in the next day or so. 

Steve Francia
spf13.com
@spf13

spf13:
Or now. ;)

I think it looks ok, but I can't think of another use case for the feature other than the one you defined. Perhaps there are many more, I haven't given it much though yet. I feel like there's an easier way for a user to use the use case you suggest. A setting perhaps.

ecnahc515:
Yeah, the main thing is, that this adds base functionality for supporting different keys in your config from keys outside your config. I can see this being applied to remote kv stores and config files where you might want to use `-` for delimiters in your `viper.Get` calls, but you might want `_` for the key names in your JSON (the typical way to do json is snakecase).

For example, you might use:
`viper.Get("someThingCamelCase")` 
And then your JSON config might have:
`{ "some_thing_camel_case": "foo"}`

PR #49, opened 2015-02-19
Move viper.Reset() back to the public interface

### LINKED ISSUE #46: viper.Reset() can longer be called from test code
We have a number of tests that depend on `viper.Reset()` to restore viper's state between tests.  Since `Reset()` got moved to viper_test.go, though, we are no longer able to do that.  I thought I remembered some documentation somewhere even suggesting that `Reset()` could be used in tests but it's possible I'm imagining that.  Anyway, having access to that function in our tests would be super helpful.

rosscooperman:
It is helpful for applications that use viper to be able to
reset their configurations between test runs.

This will resolve #46

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -102,6 +102,15 @@ func New() *viper {
 	return v
 }
 
+// Intended for testing, will reset all to default settings.
+// In the public interface for the viper package so applications
+// can use it in their testing as well.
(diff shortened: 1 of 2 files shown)


spf13:
Thanks @rosscooperman. Merged as 9073483

PR #51, opened 2015-02-28
Add GetSizeInBytes

Eloff:
Useful to parse strings like 1GB or 12 mb into an unsigned integer number of bytes. I was going to add this just to my own codebase, but realized that others can benefit.

I originally thought GetBytes() would be a good name, but it would conflict if you ever wanted to add a GetBytes() []byte to viper for some reason. GetSizeInBytes() is pretty clear.

### DIFF
--- a/util.go
+++ b/util.go
@@ -19,8 +19,10 @@ import (
 	"path/filepath"
 	"runtime"
 	"strings"
+	"unicode"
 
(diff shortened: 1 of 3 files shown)


spf13:
This is super helpful. Thanks!

spf13:
Merged as e133904

PR #53, opened 2015-03-26
Change small syntax error in Env Example

eranchetz:


kzvezdarov:
Merged, thanks for noticing that!

Issue #52, opened 2015-03-22
Proposal for materialized paths


kzvezdarov:
I've been working on reconciling #15 with the tip of master. The approach can work, but it has an issue which might lead to performance degradation for large configurations and complicate feature implementation. The index has to be rebuilt on each call to `find()` since there is no way to keep it up to date and preserve the config source hierarchy. For example, when a value is set for whatever reason (reading in file, pflag, env.), there is no clean way to check if the key to that value should become the new value of the index since all sources share the same index map. So indexing happens for every `find()` call and involves layering the configs via `AllSettings()` in order to get the proper, prioritized, structure.

I think that this can be improved in the following way:

1. Create a config source type which holds both the map of values and the flat index to that map.

2. Extract all source specific logic from the `Viper` type and attach it to a specialized config source type (e.g. `RemoteConfig`, `EnvConfig`, etc).
3. In the `Viper` type, maintain a slice of config sources, sorted for priority. 

This allows for incremental index update since each specific index can be set as appropriate and then traversed in priority as needed.
Furthermore, the refactoring of the config sources should help avoid mistakes in config priority as this will be set only once (similar to the issue that #44 fixed).
Lastly, this should make new feature implementation easier in the future - e.g., setting custom priority for config sources, for example making `Consul` the primary source, and local environment the fallback.

Here is a pseudocode to illustrate:

```
type Viper struct {
    [...]
    priority []string
    sources map[string]ConfigSource
    [...]
}

func New() *Viper {
    v.sources = map[string]{
        "flags": PFlagSource{},
        "env": EnvSource{},
        "config": FileSource{},
        [...]
    }
}

type ConfigSource interface {
    Find(key string) interface{}
    Set(key string, val interface{})
    [...]
}

type FileSource struct {
    index map[string]interface{}
    config map[string]interface{}
}
```

This is a fairly large refactor, but it should be completely backwards-compatible, fix the issues with using a shared index on the configs, and make maintenance and future development easier. Working on a prototype, will update this issue with the PR when ready.

kzvezdarov:
Closing this for now, as #63 resolves the issue, though a general refactoring in that direction may still be a good idea.

PR #54, opened 2015-03-26
Materialized Paths

kzvezdarov:
Still working on this (specifically need to add tests and documentation), but I am hoping to get some feedback on the implementation.

This is the straightforward approach that I described in #52 , with all of the drawbacks listed there.
This PR defines a new map on `Viper` which serves as an index mapping to specific values within the rest of the configuration structures. Right now the index has to be rebuilt before each `Get`, which involves traversing all configuration structures - I am looking into improving that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -235,6 +235,12 @@ Example:
         fmt.Println("verbose enabled")
     }
 
+You can get deeply nested values by providing the expected key path.
+
+Example:
+
+    viper.Get("log.verbose") // case-insensitive path <root key>.<subkey1>.<subkey2>...etc
+
 ### Marshaling
 
 You also have the option of Marshaling all or a specific value to a struct, map, etc.
--- a/util.go
+++ b/util.go
@@ -17,6 +17,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"reflect"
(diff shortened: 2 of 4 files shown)


kzvezdarov:
@spf13 Added some basic tests; working on getting some specific corner cases covered. At this point I don't see how i can avoid rebuilding the index on every request, but unless every key requested is a deeply nested key, it should not degrade performance too much.

leolara:
+1

I was thinking in using Confer, because it does have this, but I think viper is mantained more often, so it would be amazing if this functionality is included.

kzvezdarov:
I actually figured out a better way to do it over the weekend, which is why this is not merged yet. I'm probably going to get that in today or tomorrow.

spf13:
@kzvezdarov I'm really excited to see what you are working on here. Let me know if you need any help.

kzvezdarov:
@spf13 This is mostly a reference implementation on how accessing deeply nested values can work using the indexing approach. It has some drawbacks (regarding performance on very large configs) as I explained, which cannot be addressed without some extensive refactoring. 
See #63 for an alternative implementation.

kzvezdarov:
Closing this for now.

PR #60, opened 2015-04-14
59 - add properties file support to viper

wwalker:
Added support for properties file format configs (via the magiconair/properties) library


kzvezdarov:
Looks good, thanks!

PR #63, opened 2015-05-01
Access to keys nested at an arbitrary depth

kzvezdarov:
This accomplishes the same as #54 , but in a better way given the current Viper implementation since it does not rebuild the indexes on every operation, but rather uses recursive search though the registers.

This adds the ability to access keys nested in an arbitrary number of maps by passing a delimited string such as:
`databases.metrics.host`

It works in the following manner:

1. First, checks for the existence of the key as is, e.g. looks up the value of `databases.metrics.host`.

2. If nothing is found, the key is split on the delimiter (`.`). The root source is looked up by finding the first part of the path (`databases`) in this case. 

3. If the root source is a map, recursively looks for each part of the path. So first the root source will be searched for the value of the `metrics` key, and if that is a map, it will be searched for the `host` key. If all of these succeed, the value of `host` is returned.

Notes:

1. Only the first encountered source is used. So if the following situation occurs:
```
configs = {
    "databases": {}
},
defaults = {
    "databases": {
        "metrics": {
            "host": "localhost"
         }
    }
}
```
searching for `databases.metrics.host` will return `nil`, even though the `defaults` registry contains a value for the key.

2. Only the first part of the path is case insensitive. This is due to how Viper has behaved up to this point: when a config is loaded, only the surface keys are insensitivized, without recursing deeper. So the 2nd and 3rd, etc, level keys might be in mixed case. @spf13 what is your opinion on this - keep the current solution which maintains the established behavior, or recursively insensitivize everything?

3. Obviously, keys that include the delimiter will cause issues. Maybe the delimiter should be configurable/escapable?

### DIFF
--- a/util.go
+++ b/util.go
@@ -17,6 +17,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"reflect"
 	"runtime"
 	"strings"
 	"unicode"
@@ -34,6 +35,10 @@ func insensitiviseMap(m map[string]interface{}) {
 		if key != lower {
 			delete(m, key)
 			m[lower] = val
+		}
+
+		if val != nil && reflect.TypeOf(val).Kind() == reflect.Map {
+			insensitiviseMap(cast.ToStringMap(val))
 		}
 	}
 }
--- a/viper.go
+++ b/viper.go
@@ -107,6 +107,10 @@ func (rce RemoteConfigError) Error() string {
 //		"endpoint": "https://localhost"
 //	}
 type Viper struct {
+	// Delimiter that separates a list of keys
+	// used to access a nested value in one go
+	keyDelim string
+
 	// A set of paths to look for the config file in
 	configPaths []string
 
@@ -134,6 +138,7 @@ type Viper struct {
 // Returns an initialized Viper instance.
 func New() *Viper {
 	v := new(Viper)
+	v.keyDelim = "."
 	v.configName = "config"
 	v.config = make(map[string]interface{})
 	v.override = make(map[string]interface{})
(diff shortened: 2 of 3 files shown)


spf13:
@kzvezdarov I think it should be case insensitive all the way down.

kzvezdarov:
@spf13 The last commit should do the trick. Updated the tests to expect lowercase only keys as well.

spf13:
I'm a bit worried about the approach taken of matching the earliest match, but I do think it makes sense. I see this as a feature which enables a later configuration to replace an entire subtree without needing to overwrite each subkey.

kzvezdarov:
There is an argument to be made for falling through to the lower priority configurations in case a nested key is not found in the initial source, but I feel that the convenience of that is outweighed by the pitfalls that it creates. E.g. forgetting to set a sub-parameter in a config file will not be caught because of its absence, but the default value will sneak in instead.

spf13:
I like this one, but since you submitted two Prs in what ways is the other one superior to this one?

kzvezdarov:
This one is superior to #54 because it should perform better. 

#54 works by building a flat index for each registry, searching through those in priority order. The problem with it is that it has to rebuild the registries too often (mostly because of pflags/environment variables) and each rebuild recursively goes over each map.

The approach in this PR on the other hand simply searches through only the maps which are matched by the path, as opposed to everything.

spf13:
LGTM. Feel free to merge it when you can, or I will early next week if you want.

kzvezdarov:
Merged.

leolara:
:smiley:

PR #66, opened 2015-05-08
ReadBufConfig & WatchRemoteConfig

oliveagle:
ReadBufConfig
````go
func TestReadBufConfig(t *testing.T) {
	v := New()
	v.SetConfigType("yaml")
	v.ReadBufConfig(bytes.NewBuffer(yamlExample))
	t.Log(v.AllKeys())

	assert.True(t, v.InConfig("name"))
	assert.False(t, v.InConfig("state"))
	assert.Equal(t, "steve", v.Get("name"))
	assert.Equal(t, []interface{}{"skateboarding", "snowboarding", "go"}, v.Get("hobbies"))
	assert.Equal(t, map[interface{}]interface{}{"jacket": "leather", "trousers": "denim"}, v.Get("clothing"))
	assert.Equal(t, 35, v.Get("age"))
}
````

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,8 +14,9 @@ configuration. It supports
 * setting defaults
 * reading from json, toml and yaml config files
 * reading from environment variables
-* reading from remote config systems (Etcd or Consul)
+* reading from remote config systems (Etcd or Consul), watching changes
 * reading from command line flags
+* reading from buffer
 * setting explicit values
 
 It can be thought of as a registry for all of your applications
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by oliveagle** [comment]
WatchRemoteConfig

````go
runtime_viper.ReadRemoteConfig()

for {
	err := runtime_viper.WatchRemoteConfig()
	if err != nil {
		log.Errorf("unable to read remote config: %v", err)
		time.Sleep(time.Second * 5)
		continue
	}
    runtime_viper.Marshal(&runtime_conf)
    time.Sleep(time.Second * 5)   
}
```

spf13:
@oliveagle This looks really well done. Can you follow up with adding some documentation to the readme so users can know the best way to use it.

ecnahc515:
Looks awesome. One nit pick. Why cant we just make this work on a reader, rather than using a buffer specifically? This would let users use any kind of reader instead of just a bytes.Buffer, and would not restrict usage of a bytes.Buffer at all.

oliveagle:
good idea.  :+1: 

replaced `bytes.buffer` with `io.Writer`

ecnahc515:
Awesome. Could you fix up the documentation to reflect the changes, and this would be ready. Wouldn't hurt to rebase either if you dont mind.

PR #72, opened 2015-05-22
Removed CWD from default search path

kzvezdarov:


kzvezdarov:
@bep nitpick: isn't the workflow for this repository supposed to be `ff-only`, to avoid merge commits creating noise in the history?

bep:
Looking at the commit log I would say no.

bep:
That said, I prefer cherry-picking and avoid the "merge-button" in GH ... usually.

Issue #73, opened 2015-05-22
Current directory used for config if no other specified directories exist


didenko:
It does not seem to be a problem if no directories were added with `AddConfigPath` - CWD is probably a good assumption.

However, if config directories are misconfigured, or, more importantly, get accidentally erased later, then loading configuration from CWD is a surprise behaviour. I would expect a failure at that point.

Code example:

```Go
package main

import (
	"io/ioutil"
	"os"

	"github.com/spf13/viper"
)

const configFile = "surprise_inside"
const configFileFull = configFile + ".toml"

func main() {

	defer os.Remove(configFileFull)

	// Although probably distinct use cases, viper
	// reads the config in cwd regardless of AddConfigPath
	// being commented out or not
	viper.AddConfigPath("this_junk_does_not_exist")

	viper.SetConfigName(configFile)
	viper.SetDefault("key", "Feels good!")

	err := viper.ReadInConfig()
	if err != nil {
		println("Fatal error reading config file - good")
	} else {
		println(viper.GetString("key"))
	}

	println("Writing " + configFileFull)

	err = ioutil.WriteFile(
		configFileFull,
		[]byte("key = \"Surprise!\"\n"),
		0600)
	if err != nil {
		println("Fatal error writing config file - bad")
	} else {

		err = viper.ReadInConfig()
		if err != nil {
			println("Fatal error reading config file - bad")
		} else {
			println(viper.GetString("key"))
		}
	}
}
```

didenko:
The two primary scenarios I see are:

0. A config file not found in specified locations
0. No locations specified and a config file not found in CWD

It seems to be a separate issue, how to treat the lack of a config file - as an error or as a suppressed condition.

This issue deals with that CWD should not be checked for config file unless either explicitly included in search locations. If no locations specified to begin with it may be a reasonable behaviour to check CWD. Although still a debatable use case it is not worth changing today, but would be great to hear opinions about.

didenko:
Suggested fix is in the #78 pull request

didenko:
After fixing this issue the code which intentionally or not (like #104) relied on the default search in the current directory will break. To get the pre-fix functionality back in your application, add the following call the proper viper callee after adding all other search paths to it:

```go
viper.AddConfigPath(".")
```

PR #74, opened 2015-05-23
Avoid searching CWD if no config in provided dirs

didenko:
Address #73. Clarify specific error messages. Propagating errors properly up to the public function caller.


didenko:
The suggested code fails to apply defaults when configuration not found in specified directories. Will re-test and re-submit.

PR #76, opened 2015-05-27
More than 2 levels of nesting fails the searchMap search. Submitting a failing test and a fix.

jondot:


viper.go
```diff
@@ -313,6 +313,8 @@ func (v *Viper) searchMap(source map[string]interface{}, path []string) interfac
 
 	if next, ok := source[path[0]]; ok {
 		switch next.(type) {
+		case map[interface{}]interface{}:
```
kzvezdarov: Ironically I had taken care of this edge case in the previous (unmerged) implementation. Good catch.

jondot: Good to hear! I was not so confident since this is quite a trivial case.

kzvezdarov: I believe another way in which you get a map with `interface{}` as a key is when a YAML config is parsed in. In any case looks good, will try to get it merged tomorrow.

JeffMerlet: Would be good to add more deep nested properties in the viper_test.go samples to be sure that such deep nested cases are always tested. Thanks!

kzvezdarov:
Merged https://github.com/spf13/viper/commit/28ada1e5b0d3edf3d0402f2d425044ee9faaffc1

PR #77, opened 2015-05-27
AddSecureRemoteProvider missing secretKeyring to remoteProvider

chalupaul:
I set up viper with a encrypted crypt storage and couldn't decrypt anything. The error was that it was unable to wrangle it to json (I can't remember the exact). When I dug in the code, I found that when AddSecureRemoteProvider added the remoteProvider struct, it wasn't passing in the secretKeyring field so nothing was getting decrypted and the raw ascii armor'd gpg stuff was getting returned instead (hence the json marshaling failure).

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -288,6 +288,7 @@ func (v *Viper) AddSecureRemoteProvider(provider, endpoint, path, secretkeyring 
 			endpoint: endpoint,
 			provider: provider,
 			path:     path,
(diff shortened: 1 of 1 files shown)


kzvezdarov:
Merged your PR: https://github.com/spf13/viper/commit/db7ff930a189b98d602179d9001d33345f42b8c7
(rebasing and ff-merging to master was applying the changes weirdly so added it manually)

PR #78, opened 2015-05-27
Fixes #73 Current directory used for config if no other specified directories exist

### LINKED ISSUE #73: Current directory used for config if no other specified directories exist

didenko:
Added five tests for old and new behavior. Split the `findConfigFile()` code to make the control flow intent clear. Added an error type specific to missing config file for better diagnostic.

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -72,6 +72,16 @@ func (rce RemoteConfigError) Error() string {
 	return fmt.Sprintf("Remote Configurations Error: %s", string(rce))
 }
 
+// Denotes failing to find configuration file.
+type ConfigFileNotFoundError struct {
+	name, locations string
+}
+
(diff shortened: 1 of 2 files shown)


didenko:
Ping @kzvezdarov, @bep - any feedback if I can improve the code?

PR #79, opened 2015-05-28
Make the remote features optional

bep:
@spf13 this is a little bit quick and dirty from me -- I will clean up and make sure it works (I do not see why it shouldn't) when we agree that this is a good idea.

### DIFF
--- a/README.md
+++ b/README.md
@@ -201,6 +201,11 @@ Example:
 
 
 ### Remote Key/Value Store Support
+
+To enable remote support in Viper, do a blank import of the `viper/remote` package:
+
+`import _ github.com/spf13/viper/remote`
+
 Viper will read a config string (as JSON, TOML, or YAML) retrieved from a
 path in a Key/Value store such as Etcd or Consul.  These values take precedence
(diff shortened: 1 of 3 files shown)


spf13:
It looks solid to me, but I'd love to get others input on it.

spf13:
@bketelsen Do you see anything wrong with this approach?

bketelsen:
will look this evening.

bep:
For reference (and some alt solutions), see https://groups.google.com/forum/#!topic/golang-nuts/IG7Pref7kkk

The motivation is clear enough in this case. 90%+ reads config from files. And this change shaves 1,4MB off the Hugo binary.

bketelsen:
LGTM - just means I need to change my uses of this.

bep:
Good. You should get a pretty clear and instructive error message if you try to use the remote features without the necessary import.

Will wrap this up tomorrow.

leolara:
:+1:

bep:
I tested it OK with etcd. Merged in be5ff3e

PR #81, opened 2015-05-29
we should be able to recover from panic other than os.Exit(1)

oliveagle:
if you have a config file which is not UTF-8 encoding.  viper will crash and exit the program. 

    ERROR: 2015/05/29 Error parsing config: yaml: invalid leading UTF-8 octet

if our requirement is to  load config once while program init process and any error could stop it, then this won't be a problem .

but if we need to load config multiple times at runtime. this will be a problem.

### DIFF
--- a/README.md
+++ b/README.md
@@ -201,6 +201,11 @@ Example:
 
 
 ### Remote Key/Value Store Support
+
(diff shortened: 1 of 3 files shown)


bep:
Can you squash these commits into one?

Issue #83, opened 2015-05-31
Struct method BindPFlag calls package function SetDefault instead of method of the same name


leolara:
I think method `v.BindPFlag` should call `v.SetDefault` instead of just `SetDefault`

leolara:
https://github.com/spf13/viper/blob/master/viper.go#L522

didenko:
Nice find!

leolara:
just happen to be reading the code and found it. I feel it is clear enough
in README.md how to use pflags with viper.

didenko:
Fixed in SHA: 1241d2e04b9d3a9d345bf62dd3f60ca40c7dfa0a

leolara:
:+1:

didenko:
FYI, the fix is in my development, not the "official" `spf13/viper`, branch yet. If not merged in a week-two, will ask to reopen this issue.

PR #84, opened 2015-06-01
we should let uesr decide when to exit the program

oliveagle:
if the program reload configuration multiple times, it's better to let the user decide when to exit the program.

### DIFF
--- a/util.go
+++ b/util.go
@@ -126,24 +126,24 @@ func marshallConfigReader(in io.Reader, c map[string]interface{}, configType str
 	switch strings.ToLower(configType) {
 	case "yaml", "yml":
 		if err := yaml.Unmarshal(buf.Bytes(), &c); err != nil {
-			jww.ERROR.Fatalf("Error parsing config: %s", err)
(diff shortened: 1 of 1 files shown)


didenko:
Would it be preferable to have error propagated, rather than panic in general case? It is preferable for my development branch, which is at the SHA: ad82755e41a287b1bd99eb88665c904cc52e188d

It is part of the https://github.com/didenko/viper/commits/pulls_from_main branch with other fixes for my development purposes, including closing #68, #73, #78, #83, #84. @bep,@kzvezdarov, - let me know if interested and if you want a pull request, will pull yourself, or want some adjustments - or not interested. Pull requests started to accumulate, so I packaged the three issues.

didenko:
I have put the above-mentioned fixes into a pull request #86.

didenko:
Based on a feedback, the fix is put in a separate PR #98

didenko:
The issue was addressed via the PR #98, giving users an ability to analyse errors in case reading config file failed.

PR #92, opened 2015-07-04
nested config alias -test included

WnP:
hope this help, for me it's realy usefull, especialy when using with cobra pflags

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -676,6 +676,14 @@ func (v *Viper) registerAlias(alias string, key string) {
 				delete(v.override, alias)
 				v.override[key] = val
 			}
+			if strings.Contains(alias, v.keyDelim) {
+				path := strings.Split(alias, v.keyDelim)
+				source := v.find(path[0])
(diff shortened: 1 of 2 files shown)


viper.go
```diff
@@ -676,6 +676,14 @@ func (v *Viper) registerAlias(alias string, key string) {
 				delete(v.override, alias)
 				v.override[key] = val
 			}
+			if strings.Contains(alias, v.keyDelim) {
+				path := strings.Split(alias, v.keyDelim)
+				source := v.find(path[0])
+				if source != nil && reflect.TypeOf(source).Kind() == reflect.Map {
+					val := v.searchMap(cast.ToStringMap(source), path[1:])
+					v.config[key] = val
```
WnP: I do not delete the map key as done in other case, let me know if it's a problem, by the way I didn't see any side effect

WnP:
I add this commit because in case an option default value is provided by a pflag -for example- if the option is not set in config file `nil` is return

this commit fix this issue and preserve the default value

PR #94, opened 2015-07-15
Support for loading from a directory tree where each file represents one setting

niccaluim:
I've added a ReadDir method that loads configuration from a directory tree. Each file in the tree is treated as a single setting: the filename is the key, and the file contents are the value. E.g.:

```
/config -- name (file contents: Skadi)
      |
      +--- debug (file contents: true)
      |
      +--- service -- port (file contents: 1234)
                 |
                 +--- ip (file contents: 1.2.3.4)
```

Calling `ReadDir("/config")` would produce the following configuration:

```
name=Skadi
debug=true
service.port=1234
service.ip=1.2.3.4
```

No idea if you'll find this useful, but I thought I'd send it your way nonetheless.

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -749,6 +749,42 @@ func (v *Viper) ReadConfig(in io.Reader) error {
 	return nil
 }
 
+// Loads configuration from a directory tree where filenames are keys
+// and file contents are values.
+func ReadDir(dirname string) error { return v.ReadDir(dirname) }
+func (v *Viper) ReadDir(dirname string) error {
+	config, err := v.readDir(dirname, "")
+	if err == nil {
+		v.config = config
+	}
+	return err
+}
+
+func (v *Viper) readDir(dirname, keyPrefix string) (map[string]interface{}, error) {
+	entries, err := ioutil.ReadDir(dirname)
+	if err != nil {
+		return nil, err
+	}
+
+	node := make(map[string]interface{})
+	for _, entry := range entries {
+		path := filepath.Join(dirname, entry.Name())
+		if entry.IsDir() {
+			node[entry.Name()], err = v.readDir(path, keyPrefix+v.keyDelim+entry.Name())
+			if err != nil {
+				return nil, err
+			}
+		} else {
+			value, err := ioutil.ReadFile(path)
+			if err != nil {
+				return nil, err
+			}
+			node[entry.Name()] = strings.TrimRight(string(value), "\r\n")
+		}
+	}
+	return node, nil
+}
+
 // func ReadBufConfig(buf *bytes.Buffer) error { return v.ReadBufConfig(buf) }
 // func (v *Viper) ReadBufConfig(buf *bytes.Buffer) error {
 // 	v.config = make(map[string]interface{})
--- a/viper_test.go
+++ b/viper_test.go
@@ -8,7 +8,9 @@ package viper
 import (
 	"bytes"
 	"fmt"
+	"io/ioutil"
 	"os"
+	"path/filepath"
 	"sort"
 	"strings"
 	"testing"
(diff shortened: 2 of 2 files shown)


niccaluim:
Actually this should use discovery instead of reading from a given directory. Instead of `ReadDir("/config")` you'd do:

```
AddConfigPath("/config")
ReadInConfigDir()
```

kzvezdarov:
What use case does this satisfy?

philk:
We're looking for this because of the way kubernetes [handles secret distribution](https://github.com/GoogleCloudPlatform/kubernetes/blob/master/docs/design/secrets.md). Basically it mounts a volume into the container with the secrets [exposed as files](https://github.com/GoogleCloudPlatform/kubernetes/blob/master/docs/design/secrets.md#examples).

kzvezdarov:
Sounds good. There is an in-progress PR for supporting cascading config files, it might be good to coordinate the two though I am not sure yet how (and if ) it would make sense to combine them.

niccaluim:
Support for cascading configs will be great, especially since it means we could use kubernetes volumes for secrets, and JSON/YAML for other config. I'll take a look at that diff and see the two might interact.

spf13:
If we can marry these two together I think we’ve got an extremely powerful and useful tool. Just need to be mindful of how this interacts with nesting.

niccaluim:
I've merged the changes from wrobbins cascading configs branch. I'd like to take a closer look before claiming any of this still works though.

spf13:
@niccaluim Do you think this is ready for review/merge?

spf13:
@niccaluim PING :)

cynipe:
any progress on this?

spf13:
Until the CLA is signed, this is stalled.

@niccaluim would need to sign it at https://cla-assistant.io/spf13/viper?pullRequest=94

Then someone can try to merge this in and test it.

spf13:
@niccaluim can you sign the CLA https://cla-assistant.io/spf13/viper?pullRequest=94 so we can take over this PR and incorporate it into Viper?

niccaluim:
Hey there, sorry. I'll sign it just as soon as I get the ok from Legal.

niccaluim:
Alright, it's signed! Does it normally take a while for the check to update?

spf13:
Not usually. Try this link instead https://cla-assistant.io/spf13/viper . If it says you are good then you are already signed.

niccaluim:
Hmm, that doesn't seem to be working either. It shows me the CLA, I click to sign, it says I've signed and redirects me.

spf13:
Thanks for being patient. I've checked in the CLA backend and you have indeed signed it. Not sure why the PR isn't updating, but it's all good. Thanks!

vinodmut:
Are there any updates on this? As mentioned earlier in the thread, it's useful to reading Kubernetes [secrets](https://kubernetes.io/docs/user-guide/secrets/).

titpetric:
Docker introspection feature is being implemented in the same way, currently it's slated as experimental, but might land in the following months. Related issue [here](https://github.com/moby/moby/pull/26331)

kolaente:
Any updates on this considering it's been open since quite a while? There's now even an issue about this: https://github.com/spf13/viper/issues/1022

jedahan:
Would love to see this implemented to support the same use case - docker secrets

jeroenhabets:
> Would love to see this implemented to support the same use case - docker secrets

Exactly, a shame that despite all the contributions over the years this was not added. It would instantly give all depending projects more secure configuration options.

titpetric:
> Exactly, a shame that despite all the contributions over the years this was not added. It would instantly give all depending projects more secure configuration options.

https://github.com/golang/go/commit/a3fb92a7100f3f2824d483ee0cbcf1264584b3e4

Forward secrecy concerns lead people to design other abstractions which I'm more aligned with these days. Keeping env vars and config secrets in process memory is perhaps a measure of convenience, rather than security. After you've established a database connection or whatever, the secret becomes unavailable for introspection until the next connection is required. I've observed security posture like clearing the set environment for the process after the secrets have been consumed in the right places, but that still leaves config in memory.

I think this whole opinion of mine is really centered around zero trust of the runtime environment. I'd probably be more relaxed if running applications on my own hardware rather than some cloud provider, and I find it reasonable to address the concern from a design standpoint (outside of viper).

PR #96, opened 2015-07-19
Fixes #83

### LINKED ISSUE #83: Struct method BindPFlag calls package function SetDefault instead of method of the same name

didenko:
Call methods instead of package functions


spf13:
Merged as f14e1ba

PR #97, opened 2015-07-19
Readme fixes and small edits

jackspirou:
This pull request fixes a couple small README spelling typos.  It also formats the file to not hit the end of a printed page which can be seen in a text editor like atom.  Some tweaks to a couple phrases here and there have also been made to be (I think) more readable.  

Many tweaks are strictly my own opinion, and so are not critical issues that need to be fixed.

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,41 +7,41 @@ Go configuration with fangs
 
 ## What is Viper?
 
-Viper is a complete configuration solution for go applications. It has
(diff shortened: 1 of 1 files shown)


spf13:
Merged as 2abb1bebfde865b0bb6bb7ada5be63ec78527fa6. Thanks for the improvements!

PR #99, opened 2015-07-22
Fix #73 - BREAKING

### LINKED ISSUE #73: Current directory used for config if no other specified directories exist

didenko:
@spf13 at GopherCon agreed that automatically searching for a config file in CWD is a mistake in the package. It may be a valid choice for a final binary, but not a default behavior in the library.

Accordingly, this PR removed the CWD search code and adds a couple tests to searching for the config file in explicitly requested directories.

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -79,6 +79,16 @@ func (rce RemoteConfigError) Error() string {
 	return fmt.Sprintf("Remote Configurations Error: %s", string(rce))
 }
 
+// Denotes failing to find configuration file.
+type ConfigFileNotFoundError struct {
+	name, locations string
+}
+
+// Returns the formatted configuration error.
+func (fnfe ConfigFileNotFoundError) Error() string {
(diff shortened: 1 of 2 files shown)


didenko:
This PR is ready to merge after some cleanup as requested by spf13.

The code which intentionally or not (like #104) relied on the default search in the current directory will break. To get the pre-fix functionality back in your application, add the following call the proper viper callee after adding all other search paths to it:

```go
viper.AddConfigPath(".")
```

didenko:
Merged in SHA: 9fca10189b1307bba68b2cd487dd93da0bfbda06

PR #108, opened 2015-08-24
Fix #36: Marshal to Unmarshal

### LINKED ISSUE #36: Unmarshal vs. Marshal
Hey spf13, we love your package.

Seems to me you use "Marshal" to mean what most libraries I've used call "Unmarshal".  (Compare with the [Go JSON package](https://golang.org/pkg/encoding/json/#Marshal).

What's your thinking on this?

MattSurabian:
There was some concern around #36 being a breaking change. Since breaking changes were just released it seemed an opportune time to try and get this in too, even though it's arguably a more destructive breaking change.

Going forward I think this should be addressed eventually to avoid confusion for adopters.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ Viper is here to help with that.
 
 Viper does the following for you:
 
-1. Find, load, and marshal a configuration file in JSON, TOML, or YAML.
+1. Find, load, and unmarshal a configuration file in JSON, TOML, or YAML.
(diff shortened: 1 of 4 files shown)


spf13:
Ugh.. I hate to do this breaking change, but the sooner we merge this the better. 

Merged as 1967d93

MattSurabian:
Yeah this kind of change is never fun, at least the band-aid is ripped off now.

Issue #110, opened 2015-08-29
If set, default values should always imply type


akutz:
I will be submitting a pull request in accompaniment with this issue. The gist of it is this - I believe a key's default value, if set, should imply the value's type no matter from where else the value is set. This is particularly important when working with environment variables. For example:

```go
package main

import (
	"fmt"
	"os"

	"github.com/spf13/viper"
)

func print(name string, val interface{}) {
	fmt.Printf("%-15[1]s%-15[2]T%[2]v\n", name, val)
}

func main() {
	viper.BindEnv("mykey", "MYPREFIX_MYKEY")
	viper.SetDefault("mykey", []string{})
	os.Setenv("MYPREFIX_MYKEY", "a b c")

	v1 := viper.GetStringSlice("mykey")
	v2 := viper.Get("mykey")

	print("v1", v1)
	print("v2", v2)
}
```

When this program is executed the following is emitted:

```bash
[0]akutz@pax:ex$ ./ex1 
v1             []string       [a b c]
v2             string         a b c
[0]akutz@pax:ex$
```

You may wonder, why is this important? Just use the `GetStringSlice` function. Well, it *becomes* important when dealing with marshaling. If we update the above program to this:

```go
package main

import (
	"fmt"
	"os"

	"github.com/spf13/viper"
)

type Data struct {
	MyKey []string
}

func print(name string, val interface{}) {
	fmt.Printf("%-15[1]s%-15[2]T%[2]v\n", name, val)
}

func main() {
	viper.BindEnv("mykey", "MYPREFIX_MYKEY")
	viper.SetDefault("mykey", []string{})
	os.Setenv("MYPREFIX_MYKEY", "a b c")

	v1 := viper.GetStringSlice("mykey")
	v2 := viper.Get("mykey")

	print("v1", v1)
	print("v2", v2)

	d := &Data{}
	viper.Marshal(d)
	print("d.MyKey", d.MyKey)
}
```

Now we can see the issue when we execute the updated program:

```bash
[0]akutz@pax:ex$ ./ex2
v1             []string       [a b c]
v2             string         a b c
d.MyKey        []string       []
[0]akutz@pax:ex$
```

The marshalled data structure's field `MyKey` is empty when in fact it should have a string slice equal to, in value, `[]string {"a", "b", "c"}`.

[...]

PR #111, opened 2015-08-29
[110] Default Values Specify Type

akutz:
(Issue #110)

This patch adds a feature, if enabled, will infer a value's type from
its default value no matter from where else the value is set. This is
particularly important when working with environment variables. For
example:

    package main

    import (
      "fmt"
      "os"

      "github.com/spf13/viper"
    )

    func print(name string, val interface{}) {
      fmt.Printf("%-15[1]s%-15[2]T%[2]v\n", name, val)
    }

    func main() {
      viper.BindEnv("mykey", "MYPREFIX_MYKEY")
      viper.SetDefault("mykey", []string{})
      os.Setenv("MYPREFIX_MYKEY", "a b c")

      v1 := viper.GetStringSlice("mykey")
      v2 := viper.Get("mykey")

      print("v1", v1)
      print("v2", v2)
    }

When this program is executed the following is emitted:

    [0]akutz@pax:ex$ ./ex1
    v1             []string       [a b c]
    v2             string         a b c
    [0]akutz@pax:ex$

You may wonder, why is this important? Just use the GetStringSlice
function. Well, it *becomes* important when dealing with marshaling.
If we update the above program to this:

    package main

    import (
      "fmt"
      "os"

      "github.com/spf13/viper"
    )

    type Data struct {
      MyKey []string
    }

    func print(name string, val interface{}) {
      fmt.Printf("%-15[1]s%-15[2]T%[2]v\n", name, val)
    }

    func main() {
      viper.BindEnv("mykey", "MYPREFIX_MYKEY")
      viper.SetDefault("mykey", []string{})
      os.Setenv("MYPREFIX_MYKEY", "a b c")

      v1 := viper.GetStringSlice("mykey")
      v2 := viper.Get("mykey")

      print("v1", v1)
      print("v2", v2)

      d := &Data{}
      viper.Marshal(d)
      print("d.MyKey", d.MyKey)
    }

Now we can see the issue when we execute the updated program:

    [0]akutz@pax:ex$ ./ex2
    v1             []string       [a b c]
    v2             string         a b c
    d.MyKey        []string       []
    [0]akutz@pax:ex$

The marshalled data structure's field MyKey is empty when in fact it
should have a string slice equal to, in value, []string {"a", "b",
"c"}.

The problem is that viper's Marshal function calls AllSettings which
ultimately uses the Get function. The Get function does try to infer
the value's type, but it does so using the type of the value retrieved
using this logic:

    Get has the behavior of returning the value associated with the
    first place from where it is set. Viper will check in the
    following order:

      * override
      * flag
      * env
      * config file
      * key/value store
      * default

While the above order is the one we want when retrieving the values,
this patch enables users to decide if it's the order they want to be
used when inferring a value's type. To that end the function
SetTypeByDefaultValue is introduced. When SetTypeByDefaultValue(true)
is called, a call to the Get function will now first check a key's
default value, if set, when inferring a value's type. This is
demonstrated using a modified version of the same program above:

    package main

    import (
      "fmt"
      "os"

      "github.com/spf13/viper"
    )

    type Data struct {
      MyKey []string
    }

    func print(name string, val interface{}) {
      fmt.Printf("%-15[1]s%-15[2]T%[2]v\n", name, val)
    }

    func main() {
      viper.BindEnv("mykey", "MYPREFIX_MYKEY")
      viper.SetDefault("mykey", []string{})
      os.Setenv("MYPREFIX_MYKEY", "a b c")

      v1 := viper.GetStringSlice("mykey")
      v2 := viper.Get("mykey")

      print("v1", v1)
      print("v2", v2)

      d1 := &Data{}
      viper.Marshal(d1)
      print("d1.MyKey", d1.MyKey)

      viper.SetTypeByDefaultValue(true)

      d2 := &Data{}
      viper.Marshal(d2)
      print("d2.MyKey", d2.MyKey)
    }

Now the following is emitted:

    [0]akutz@pax:ex$ ./ex3
    v1             []string       [a b c]
    v2             string         a b c
    d1.MyKey       []string       []
    d2.MyKey       []string       [a b c]
    [0]akutz@pax:ex$

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -143,13 +143,14 @@ type Viper struct {
 	automaticEnvApplied bool
 	envKeyReplacer      *strings.Replacer
 
-	config   map[string]interface{}
-	override map[string]interface{}
-	defaults map[string]interface{}
-	kvstore  map[string]interface{}
-	pflags   map[string]*pflag.Flag
-	env      map[string]string
-	aliases  map[string]string
+	config         map[string]interface{}
+	override       map[string]interface{}
+	defaults       map[string]interface{}
+	kvstore        map[string]interface{}
+	pflags         map[string]*pflag.Flag
+	env            map[string]string
+	aliases        map[string]string
+	typeByDefValue bool
 }
 
 // Returns an initialized Viper instance.
@@ -164,6 +165,7 @@ func New() *Viper {
 	v.pflags = make(map[string]*pflag.Flag)
 	v.env = make(map[string]string)
 	v.aliases = make(map[string]string)
+	v.typeByDefValue = false
 
 	return v
 }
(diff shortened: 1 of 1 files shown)


spf13:
Merged as 0a4a93b

PR #114, opened 2015-09-29
Added method ReadConfigNoNil

akutz:
This patch adds a new method called ReadConfigNoNil that will read in a configuration file, but only set existing keys if they exist in the provided file.


spf13:
I think I understand a valid use case for this, but the name of the function is really confusing to me. It also would be best to have accompanying documentation.

akutz:
Hi @spf13,

I'm closing this as it's superseded by #132.

PR #118, opened 2015-10-13
Fixed #115: Added code in the find method to search for nested configuration parameters

### LINKED ISSUE #115: Defaults take precedence over actual values in nested configs
So I changed the TestDefault in viper_test.go to create this effect

```go
func TestDefault(t *testing.T) {
        SetDefault("clothing.jacket", "slacks")
        assert.Equal(t, "slacks", Get("clothing.jacket"))
        err := ReadConfig(bytes.NewBuffer(yamlExample))
        assert.NoError(t, err)
        assert.Equal(t, "denim", Get("clothing.jacket"))
}
```
output
```
--- FAIL: TestDefault (0.00s)
        Error Trace:    viper_test.go:204
        Error:          Not equal: "denim" (expected)
                                != "slacks" (actual)
```

Omegaice:
As described in Issue #115, nested configuration defaults do not work. It was caused because the find method tried to search the config for the value which failed causing it to always return the default value to the Get function.

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -657,6 +657,20 @@ func (v *Viper) find(key string) interface{} {
 		return val
 	}
 
+	// Test for nested config parameter
+	if strings.Contains(key, v.keyDelim) {
+		path := strings.Split(key, v.keyDelim)
+
+		source := v.find(path[0])
+		if source != nil {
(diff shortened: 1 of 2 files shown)


spf13:
Please sign the CLA (added recently). 

I've merged this as 87b94ba

PR #122, opened 2015-10-26
Bugfix for Nested Key Casing

akutz:
This patch fixes a bug with how Viper handle's key casing when keys are nested. Previously Viper would not lcase the first part of a key's path when looking that key up, but said key is in fact stored in lower case, so it must be cased as such.


spf13:
Merged as 110492b

PR #128, opened 2015-11-03
Clarify lack of working directory lookup by default, fix #32

### LINKED ISSUE #32: Clarify error handling and config file search techniques in documentation
I included a few specific issues I had to figure out in issue #6 that has been closed, but I hadn't seen any specific fixes in the docs for the issues listed. I'm fine with the bug being 'wont fixed' but I just wanted to be sure the specific issues I had were bubbled up in case they were seen as just a 'me too' post on the thread. From #6:

I couldn't get this working on my first try out of the box and no feedback/ detailed errors (beyond generic io errors) from the apis makes figure this stuff out even harder

In my particular case, I assumed two things that were not obvious to me without reading the code:
* the current working directory is not included in `configPaths` by default.
* the config file is required to have the extension of the file type even though you can specify through SetConfigType. I assumed the search code would have included looking at `configName` rather than just `configName+"."+ext` for each known extension

In addition, it would have been helpful for the example that does exist in the readme to properly check for errors loading the file to make it clear that thats how the API works without needing to dig into the source (as is the norm in go) e.g:

    viper.AddConfigPath("$HOME/.appname")  // call multiple times to add many search paths
    err := viper.ReadInConfig() // Find and read the config file
    if err != nil {
      fmt.Println("Failed to read config", err)
    }

didenko:


spf13:
I think that this is bad behavior to read the config in the CWD by default. I don't think it's normal behavior for most applications and could lead to some very awful things happening.

spf13:
Vlad,
I'm confused about this. In reading your prior comments, I thought you agreed that this was a bad idea. I like the clarifying language, but am confused about the additional directory to search.

didenko:
Steve, sorry if I did not communicate clearly.

Th #32 issue was already partially addressed, except for the CWD part. From what I understand the #32 does not actually request to add CWD by default, but rather to document what the code does in regard to CWD.

There is no change in my opinion that it is not the library-level decision to search CWD, but rather an application level decision - like in Hugo ( spf13/hugo#1363 ). Because of such applications I feel compelled to give an example of `viper.AddConfigPath(".")` in the README.md to show hoe Viper can _optionally_ meet that application requirement.

Would it be more clear if the section reads as following (added the "example" and "required" clarifications):

### Reading Config Files

Viper requires minimal configuration so it knows where to look for config files.
Viper supports JSON, TOML and YAML files. Viper can search multiple paths, but
currently a single Viper instance only supports a single configuration file.
Viper does not default to any configuration search paths leaving defaults decision
to an application.

Here is an example of how to use Viper to search for and read a configuration file.
None of the specific paths are required, but at least one path should be provided
where a configuration file is expected. 

```go
viper.AddConfigPath("$HOME/.appname")  // call multiple times to add many search paths
viper.AddConfigPath(".")               // optionally look for config in the working directory
err := viper.ReadInConfig() // Find and read the config file
if err != nil { // Handle errors reading the config file
	panic(fmt.Errorf("Fatal error config file: %s \n", err))
}
```

spf13:
@didenko I think this revised language is perfect.

didenko:
Will merge through the cleaned up PR #130.

variab1e:
Yes, I think this is perfectly clear.

PR #130, opened 2015-11-06
Clarify CWD search. Fixes #32

### LINKED ISSUE #32: Clarify error handling and config file search techniques in documentation

didenko:


spf13:
merged as a7ef020a9a8138a7d40d39796b1a4023d46328cb

PR #132, opened 2015-11-12
Adds MergeConfig functionality

akutz:
This patch adds the `MergeConfig` and `MergeInConfig` functions to enable reading new configuration files via a merge strategy rather than replace. For example, take the following as the base YAML for a configuration:

    hello:
        pop: 37890
        world:
        - us
        - uk
        - fr
        - de

Now imagine we want to read the following, new configuration data:

    hello:
        pop: 45000
        universe:
        - mw
        - ad
    fu: bar

Using the standard `ReadConfig` function the value returned by the nested key `hello.world` would no longer be present after the second configuration is read. This is because the `ReadConfig` function and its relatives replace nested structures entirely.

The new `MergeConfig` function would produce the following config after the second YAML snippet was merged with the first:

    hello:
        pop: 45000
        world:
        - us
        - uk
        - fr
        - de
        universe:
        - mw
        - ad
    fu: bar

Examples showing how this works can be found in the two unit tests named `TestMergeConfig` and `TestMergeConfigNoMerge`.

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -858,10 +858,122 @@ func (v *Viper) ReadInConfig() error {
 	return v.unmarshalReader(bytes.NewReader(file), v.config)
 }
 
+// MergeInConfig merges a new configuration with an existing config.
+func MergeInConfig() error { return v.MergeInConfig() }
+func (v *Viper) MergeInConfig() error {
+	jww.INFO.Println("Attempting to merge in config file")
+	if !stringInSlice(v.getConfigType(), SupportedExts) {
+		return UnsupportedConfigError(v.getConfigType())
+	}
+
+	file, err := ioutil.ReadFile(v.getConfigFile())
+	if err != nil {
+		return err
+	}
+
+	return v.MergeConfig(bytes.NewReader(file))
+}
+
+// Viper will read a configuration file, setting existing keys to nil if the
+// key does not exist in the file.
 func ReadConfig(in io.Reader) error { return v.ReadConfig(in) }
 func (v *Viper) ReadConfig(in io.Reader) error {
 	v.config = make(map[string]interface{})
 	return v.unmarshalReader(in, v.config)
+}
+
+// MergeConfig merges a new configuration with an existing config.
+func MergeConfig(in io.Reader) error { return v.MergeConfig(in) }
+func (v *Viper) MergeConfig(in io.Reader) error {
+	if v.config == nil {
+		v.config = make(map[string]interface{})
+	}
+	cfg := make(map[string]interface{})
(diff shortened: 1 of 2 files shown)


spf13:
I love this. Where should we implement it internally? Should it be the default behavior for things?

This is how I've intended the functionality to work across things. 

The only problem here is we need a mechanism to clear out a key.

akutz:
Hi @spf13,

As for clearing out a key that's difficult as even `nil` could be considered a string value. Here's my suggestion - a key can be cleared by setting it to nothing. This should result in an empty string (I believe) and thus anything set to an empty string in the source map is cleared from the target.

Also, on another note please see the compromise I had to make in the internal `mergeMaps` method. In the project on which I'm working where I want to transform a Viper instance to JSON I had to patch the underlying `yaml` project with this https://github.com/akutz/yaml/commit/b4a9f8c4b84c6c4256d669c649837f1441e4b050. It's because `go-yaml` unmarshals any nested structures below the first level as `map[interface{}]interface{}` and golang's stdlib `encoding/json` barfs at marshaling a `map[interface{}]interface{}`. It's simply not supported.

akutz:
Hi @spf13,

FYI, I'm doing a lot of work on Viper recently for something that I've recently made into it's own project, [Gofig](https://github.com/akutz/gofig). However, maybe you see some functionality in Gofig you'd rather I build into Viper?

akutz:
Hi @spf13,

Any more thoughts on this?

dcelasun:
This would be an absolutely amazing addition to viper. @spf13, any chance it could get merged soon?

spf13:
I just need time to test this further and make sure we've thought of all edge cases. I think it's one of the best additions to Viper we've seen and would be amazing to have. Any help that can be provided would be certainly appreciated and would help move it along faster. 

I'm concerned we don't have enough tests and we need to test it against real world usage of Viper (Hugo at the least) to make sure it behaves as expected.

akutz:
Hi @spf13,

If it helps, [Gofig](https://github.com/akutz/gofig), [REX-Ray](https://github.com/emccode/rexray), and [libStorage](https://github.com/emccode/libstorage) all make heavy use of MergeConfig. 

GoFig even has [several tests](https://github.com/akutz/gofig/blob/master/gofig_test.go) that confirm MergeConfig's functionality.

spf13:
@akutz That helps quite a bit actually. I'd love to get it merged. I'll try to make time this week on it.

spf13:
Merged as 991d18a. Thanks for the great contribution!

PR #134, opened 2015-11-29
Refactored IsSet to examine keys

akutz:
This patch refactors the IsSet function to examine the keys in order to see if a key is set instead of simply checking if a value is nil. This change is necessary due to the fact that default values via flag bindings will result in the old logic always being true for the IsSet function due to a type's default value such as 0 for an integer or an empty string for a string. While a type's default value may be preferable when getting the value for a key, it results in a false positive when determining if a key is actually set. This change enables users to detect whether a key is set by only returning a flag's value if it has changed.


spf13:
merged as e3bc06f

PR #143, opened 2015-12-10
Add FlagValue interface to support other flag systems.

calavera:
Using an interface allows people to use their favourite flag system
with viper without being restricted to the semantics of pflag or the
standard library.

This change introduce two new functions `BindFlagValues` and
`BindFlagValue` that behave like `BindFlags` and `BindFlag` but using
the new interface as values.

This change also introduces two internal structures to transform
`*pflag.FlagSet` and `*pflag.Flag` into the new interface. This way,
viper keeps working as expected for people that are currently using the
pflag package without breaking backwards compatibility.

Signed-off-by: David Calavera <david.calavera@gmail.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -259,6 +259,49 @@ func main() {
 }
 ```
 
+#### Flag interfaces
+
+Viper provides two Go interfaces to bind other flag systems if you don't use `Pflags`.
+
+`FlagValue` represents a single flag. This is a very simple example on how to implement this interface:
+
+```go
+type myFlag struct {}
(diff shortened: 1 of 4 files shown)


spf13:
@calavera Good to hear from you, this patch is great! Would you mind adding a bit to the readme so others can benefit from this?

calavera:
@spf13 I've added some docs to the readme. Let me know what you think.

spf13:
Looks good. Thanks!

spf13:
Merged into master c8c6312

PR #146, opened 2015-12-24
Add Sub() for Viper, which returns a branch of a Viper instance.

flei2000:
Suppose I have a component which accepts a Viper object as config when it's initialized. Now I embed the config information of the component in the main config. But the component does not know where the root of its config is. In this case, with the Sub() function, the main program can extract exactly the sub-tree for this component and transfer the sub-tree to the component.

### DIFF
--- a/README.md
+++ b/README.md
@@ -441,6 +441,52 @@ will be returned instead. E.g.
 GetString("datastore.metric.host") //returns "0.0.0.0"
 ```
 
+### Extract sub-tree
+
+Extract sub-tree from Viper.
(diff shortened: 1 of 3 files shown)


spf13:
Can you please add some documentation to the readme on using this?

Is there a reason it goes to yaml and back? I would imagine there's a better way to do this, but I see how this is the simple way.

flei2000:
Updated. Please review again.

spf13:
I think this is much better. Thanks!

PR #147, opened 2015-12-25
Handle the case Get() returns map[interface{}]interface{}

flei2000:
Sometime Get()returns type map[interface{}]interface{}. This patch is able to handle it as well as map[string]interface{}.

### DIFF
--- a/viper_test.go
+++ b/viper_test.go
@@ -730,7 +730,10 @@ func TestSub(t *testing.T) {
 	v.SetConfigType("yaml")
 	v.ReadConfig(bytes.NewBuffer(yamlExample))
 
-	subv := v.Sub("clothing.pants")
+	subv := v.Sub("clothing")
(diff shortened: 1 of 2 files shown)


akutz:
Hi @flei2000,

Could you possibly add a test to demonstrate the two cases when `Get()` returns `map[interface{}]interface{}` as well as `map[string]interface{}` and how the patch returns both successfully? Thank you!

viper_test.go
```diff
@@ -730,7 +730,10 @@ func TestSub(t *testing.T) {
 	v.SetConfigType("yaml")
 	v.ReadConfig(bytes.NewBuffer(yamlExample))
 
-	subv := v.Sub("clothing.pants")
+	subv := v.Sub("clothing")
```
flei2000: Get("clothing") returns type map[interface{}]interface{}.
Get("clothing.pants") returns type map[string]interface{}.

PR #149, opened 2015-12-30
Adding an Unmarshal Function that errors when a field in the conf is not present in the destination struct

ramrunner:
This functions protects from configs that might have unused or mistyped fields when unmarshalling into a struct. The relevant test passes.

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -623,6 +623,37 @@ func (v *Viper) Unmarshal(rawVal interface{}) error {
 	return nil
 }
 
+// A wrapper around mapstructure.Decode that mimics the WeakDecode functionality
+// while erroring on non existing vals in the destination struct
+func weakDecodeExact(input, output interface{}) error {
+	config := &mapstructure.DecoderConfig{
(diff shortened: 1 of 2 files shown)


doublerebel:
LGTM :+1:

spf13:
Thanks for the contribution. I like the feature but I find the names confusing. I don't think of Non-Existing when I read NX. 

Maybe call it UnmarshalExact or something along those lines?

Additionally the PR now has includes to your fork.

ramrunner:
sure thing. i agree with your naming suggestion. hope these commits are ok with you.

spf13:
The commits should be squashed via rebase into a single commit. Then git push --force it and it will update this PR. 

If you are not familiar with the process there is a thorough guide at https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request

If you have questions, please just ask.

ramrunner:
hope it's ok. thanks for the tutorial.

spf13:
Looks great. Merged as c975dc1

PR #152, opened 2016-01-09
Expose the default Viper instance through methods.

abourget:
To open up programming switching of configs, related to
https://github.com/spf13/hugo/pull/1734

The goal is to be able to load multiple configs from the same file, by splitting along the "---" or "+++" or "}\n{" markers, and swap the config when rendering each language, one after the other.

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -174,6 +174,18 @@ func New() *Viper {
 	return v
 }
 
+// DefaultConfig retrieves the default Viper instance. To be loaded
+// later with SetDefaultConfig.
+func DefaultConfig() *Viper {
+	return v
+}
+
+// SetDefaultConfig applies this Viper instance as the default. Used
+// when calling the functions defined in the viper package.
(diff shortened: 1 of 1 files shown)


spf13:
I think this is a good idea but I think the names will confuse people considering Default and Config already mean different things in Viper. I'm going to merge your code and then follow up with a rename.

spf13:
Actually thinking through this more. I'm not clear why this would be needed and I'm a bit worried that being able to change the default on the fly could result in strange behaviors. Let's talk about this a bit more and see if there's a way to get done what you need to do without this change.

abourget:
This was for this PR: https://github.com/spf13/hugo/pull/1764 but I think we're not going to use the code that required this PR.. as I'm going to rework the multliingual there to use some sort of inheritance, so we'll be creating an abstraction on _top_ of viper rather..

Let's close this..

PR #153, opened 2016-01-11
Adding SaveConfig().

g3rk6:
It would be nice to have a way to store configuration into a file. There could be better way of doing it, but this is the version that works for me.

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -888,11 +888,10 @@ func (v *Viper) SaveConfig() error {
 	}
 
 	f, err := os.Create(v.getConfigFile())
-	defer f.Close()
-
 	if err != nil {
 		return err
 	}
+	defer f.Close()
 
 	switch v.getConfigType() {
 	case "json":
--- a/viper.go
+++ b/viper.go
@@ -22,6 +22,7 @@ package viper
 import (
 	"bufio"
 	"bytes"
+	"encoding/json"
 	"fmt"
 	"io"
 	"io/ioutil"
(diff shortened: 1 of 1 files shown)


viper.go
```diff
@@ -874,6 +878,51 @@ func (v *Viper) InConfig(key string) bool {
...
+func SaveConfig() error { return v.SaveConfig() }
+func (v *Viper) SaveConfig() error {
+
+	jww.INFO.Println("Attempting to write config into the file.")
+	if !stringInSlice(v.getConfigType(), SupportedExts) {
+		return UnsupportedConfigError(v.getConfigType())
+	}
+
+	f, err := os.Create(v.getConfigFile())
+	defer f.Close()
```
s7v7nislands: defer should after check error?

g3rk6: Fixed it. Thanks for pointing it out!

doublerebel:
LGTM :+1: .  Just needs docs.

spf13:
Thanks for the contribution!
I love this idea, and the approach, but I think the PR needs some tweaking.

A few suggestions:

* Remove the debug line you left in
* Create a function in util.go that is called marshallConfigWriter
It would do the opposite of the unmarshallConfigReader function presently there
* Expand your current code to handle the additional types (hcl & prop)
* rename SaveConfig to WriteConfig()
* perhaps add another function called SafeWriteConfig() which only writes when the file doesn't currently exist.
* Add some tests. [Afero](https://github.com/spf13/afero) would be very helpful here. 
* Add some docs. It's a great feature, people should know about it.

I think you've done most of the work already, it's just mostly a question of reorganizing things.

Let me know if you want help or have questions. Happy to help.

bryanl:
This idea is six months old, so I'd assume this PR is dead?

AntonioMeireles:
@g3rk6 hi, anything (other than time) blocking you from progressing in this front ?

g3rk6:
Sorry for the long delay. At the moment the only issue is time and currently super busy. I'll try to make some progress as soon as I get some time. Thanks.

theherk:
@g3rk6 I'd love to see this fully implemented. I think it is looking good.

spf13:
this is stale and it appears #287 has carried it forward. Closing this.

PR #154, opened 2016-01-11
add func:GetConfig

seantao:
hi
this is a way to provide to get v.config,so I can do something about v.config,such as delete(c,key).
sorry for bad Endlish!
thank you!


spf13:
This would expose functionality that should be left private. I can't think of a good reason to do this. Please provide more information if you feel that there is a good reason and re-open.

PR #155, opened 2016-01-20
use aliases in all keys list to enable proper Unmarshaling

unilama:
I added support for aliasing in Unmarashalling, of course I can use StructTag "mapstructure" like here: https://github.com/spf13/viper#unmarshaling but in this case we remap it not alias it,  so you cannot use original key, only new remaped one. That will be helpful in config evolution.

In addition it looks like inconsistency in alias behavior across whole lib.


spf13:
Merged as e072d51

Thanks!

PR #156, opened 2016-01-24
Prevent shadowning of keys when a nested key's value is nil.

laszlo-kiss:
The change enables setting defaults for nested keys w/o those keys being shadowed by non-existent configuration values (nil) when read from a file.

For example:
```go
viper.SetDefault("nested.one","foo")
...
viper.ReadInConfig()     // { "nested" : { "two" : "bar" } }
...
```
Following the configuration read the "nested.one" will become inaccessible w/o this change.

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -757,8 +757,10 @@ func (v *Viper) find(key string) interface{} {
 		if source != nil {
 			if reflect.TypeOf(source).Kind() == reflect.Map {
 				val := v.searchMap(cast.ToStringMap(source), path[1:])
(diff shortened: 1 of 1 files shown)


laszlo-kiss:
CLA signed.

spf13:
@laszlo-kiss please add a test to confirm this behavior and do a rebase and git push. There's no reason to have the extra commit and then revert in the PR. Let me know if you need help with this.

awishformore:
Closed in favour of #165

PR #161, opened 2016-02-09
Allow errors to propagate through getConfigFile()

trombonehero:
Currently, when I run a Viper-enabled application without a configuration file, expecting it to just use the default values, I see a slightly unintuitive error: `Unsupported Config Type ""`. When I *write* an application with Viper, I have no way to distinguish between configuration file errors and users who have chosen not to write an explicit config file. Although `findConfigFile()` can return a `ConfigFileNotFoundError`, some of its callers (e.g., `getConfigFile()`) silently drop the error, resulting in lots of `""` strings with no explanation. Code like `ReadInConfig()` must then re-construct the error without information about why it occurred, so "there is no config file" is conflated with "the config file is not a valid format".

This patch changes `getConfigFile()` to propagate the error to its caller, allowing applications to handle the "no config file" case differently from other file-reading errors. It also means that, when no configuration file is found, `ReadInConfig()` can return a `ConfigFileNotFoundError` rather than an `UnsupportedConfigError`.

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -238,7 +238,13 @@ func (v *Viper) WatchConfig() {
 		defer watcher.Close()
 
 		// we have to watch the entire directory to pick up renames/atomic saves in a cross-platform way
-		configFile := filepath.Clean(v.getConfigFile())
+		filename, err := v.getConfigFile()
+		if err != nil {
+			log.Println("error:", err)
+			return
+		}
+
+		configFile := filepath.Clean(filename)
 		configDir, _ := filepath.Split(configFile)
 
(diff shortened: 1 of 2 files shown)


fatih:
Is there any chance we can get merged this? Right now I rely on this and I can't get the underlying error as it's not returned to the user.

awishformore:
I will look at this PR soon. In the meantime, it would help if you could rebase it?

awishformore:
@trombonehero are you going to rebase it?

trombonehero:
@awishformore: yup :)

PR #163, opened 2016-02-19
Replace "." characters with "_" when parsing environment variables

four2five:
This change allows the parsing of environment variables that correspond to nested keys. This seems like a safe conversion since "." is not allowed in environment variables.

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -305,6 +305,10 @@ func (v *Viper) getEnv(key string) string {
 	if v.envKeyReplacer != nil {
 		key = v.envKeyReplacer.Replace(key)
 	}
+
+	// Replace dots with underscores, since that is the sensible mapping
+	// for environment variables. The -1 means replace all insstances of "." with "_"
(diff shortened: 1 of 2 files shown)


four2five:
I'm waiting on clearance from my work to sign the CLA. Should have it today or tomorrow. Would love some comments on the patch in the interim.

chancez:
Why not just use `SetEnvReplacer(string...)`?

four2five:
Yeah, that seems to be the route forward. I need to rework this when I have a bit more time. Thanks for the suggestion.

four2five:
I intend to rework this patch to remove my code change and to update the test cases to use SetEnvReplacer , as suggested. That would (IMHO) add value as an example of how to solve this issue and as a valuable test case.

awishformore:
If you still want to write test cases for the `SetEnvReplacer`, feel free to create a new pull request. I will close this PR for the time being.

Issue #145, opened 2015-12-24
idea / question: generation of config from a strut / json spec


joeblew99:
i have been using runc, and it has the ability to generate config based on a spec.
Its really useful when you have lots of configs.
Its uses codeganstra/cli package.

See here:
https://github.com/opencontainers/runc/blob/d97d5e8b007e4657316eed76ea30bc0f690230cf/spec.go

The specs are imported from: "github.com/opencontainers/specs"
Then it just builds configs for you.  
This is where viper and cobra integrate.

SO i am wondering if there is an example of this usage example with viper / cobra ?

doublerebel:
Would #132 and #153 together create this functionality, or is there more to it?

PR #164, opened 2016-03-02
flatten defaults to ensure env var load; expand flat settings to ensure override precedence [Fixes #71] [Fixes #160]

### LINKED ISSUE #71: Default value for nested key
To support yaml and json config files, it would be nice if it is possible to set default values for nested key.

Currently when I set a default value for.

```golang
viper.SetDefault("datastore.metric.host", "0.0.0.0")
```

The key would not be interpreted as a nested key and it is not possible to overwrite it with a yaml or json config file

### LINKED ISSUE #160: Nested keys don't work with AutomaticEnv
Periods are not appropriate for environment variables, so for a nested key `a.b` with an environment prefix `example`, you might run your program like this:

```bash
EXAMPLE_A_B=3 ./program
```

But this example code:

```go
package main

import (
	. "fmt"
	"github.com/spf13/viper"
)

func main() {
	v := viper.New()

	v.SetEnvPrefix("example")
	v.AutomaticEnv()

	Println(v.Get("a.b"))
	Println(v.Get("a_b"))
}
```

Will print:

```
<nil>
3
```

This is inconsistent, as [nested keys](https://github.com/spf13/viper#accessing-nested-keys) in viper are accessed with a `.` separator, not an underscore.
Tangential relation to #71.

doublerebel:
There are some issues with nested settings precedence, due to some settings arriving from a flat map (flags, env vars) and some settings arriving from a nested map (existing structs, config files).

Right now `AutomaticEnv` only notices env vars if the flat path has been declared as a key.  The best place to ensure a key is declared is in the defaults.  Therefore we create a flatmap when setting the defaults.

However, if any flat keys are unexpanded when `Unmarshal`ing to a Struct or Map, the flat keys will be ignored in favor of the existing expanded structure.  Therefore we expand all keys before unmarshaling to the final config.

This should solve all of the issues in #160 but we probably need a test for nested config and precedence, as proposed in #163.

Also related: #145, since with this change `Unmarshal` successfully merges config from env vars, pflags, defaults, and existing structures.  I used this change [to implement Viper config in ebay/fabio](https://github.com/doublerebel/fabio/tree/viper-config-loader).

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -37,6 +37,7 @@ import (
 	jww "github.com/spf13/jwalterweatherman"
 	"github.com/spf13/pflag"
 	"gopkg.in/fsnotify.v1"
+	"github.com/doublerebel/bellows"
 )
 
 var v *Viper
@@ -612,7 +613,8 @@ func (v *Viper) UnmarshalKey(key string, rawVal interface{}) error {
 // on the fields of the structure are properly set.
 func Unmarshal(rawVal interface{}) error { return v.Unmarshal(rawVal) }
 func (v *Viper) Unmarshal(rawVal interface{}) error {
-	err := mapstructure.WeakDecode(v.AllSettings(), rawVal)
+	expanded := bellows.Expand(v.AllSettings())
+	err := mapstructure.WeakDecode(expanded, rawVal)
 
 	if err != nil {
 		return err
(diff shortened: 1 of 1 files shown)


viper.go
```diff
@@ -908,8 +910,11 @@ func (v *Viper) InConfig(key string) bool {
 func SetDefault(key string, value interface{}) { v.SetDefault(key, value) }
 func (v *Viper) SetDefault(key string, value interface{}) {
 	// If alias passed in, then set the proper default
-	key = v.realKey(strings.ToLower(key))
-	v.defaults[key] = value
+	flat := bellows.FlattenPrefixed(value, key)
```
doublerebel: We flatten the `value` if it is Struct or Map.  Therefore, if we have [a Config defined as a struct](https://github.com/doublerebel/fabio/blob/5a05baa4418abdb74c8cbc39d465858b18f0e1db/config/config.go) and [define defaults as an instance of the Config struct](https://github.com/doublerebel/fabio/blob/5a05baa4418abdb74c8cbc39d465858b18f0e1db/config/default.go), then we can [set any or all defaults in a single call](https://github.com/doublerebel/fabio/blob/5a05baa4418abdb74c8cbc39d465858b18f0e1db/config/load.go#L45).

awishformore:
This seems like an important improvement, I'll try to take a look soon. In the meantime, if you could rebase it, it would be appreciated.

awishformore:
I think this PR has a different approach to what was merged in #195 . It flattens everything, while we now create maps for everything. I wonder if correct behaviour would have been possible with your approach, because I think it would not have the same performance implications, though I doubt it.

Issue #162, opened 2016-02-13
defaults for nested keys don't work with AllSettings or Unmarshal


jbeda:
```go
package main

import (
	"bytes"
	"fmt"
	"github.com/kr/pretty"
	"github.com/spf13/viper"
)

type Config struct {
	Explicit int
	Nested   struct {
		Explicit int
		Default  int
	}
}

func main() {
	var input = []byte(`
  explicit: 1
  nested:
    explicit: 2
`)
	viper.SetDefault("nested.default", 3)
	viper.SetConfigType("yaml")
	viper.ReadConfig(bytes.NewBuffer(input))

	fmt.Println("viper.AllSettings:")
	pretty.Println(viper.AllSettings())

	var c Config
	viper.Unmarshal(&c)
	fmt.Println("\nviper.Unmarshal:")
	pretty.Println(c)
}
```

Produces:
```
viper.AllSettings:
map[string]interface {}{
    "explicit": int(1),
    "nested":   map[interface {}]interface {}{
        "explicit": int(2),
    },
    "nested.default": nil,
}

viper.Unmarshal:
main.Config{
    Explicit: 1,
    Nested:   struct { Explicit int; Default int }{Explicit:2, Default:0},
}
```

I would expect the defaults would be applied in both of these cases.

awishformore:
Fixed by #195

PR #165, opened 2016-03-05
Prevent shadowning of keys when a nested key's value is nil.

abice:
This is taking the change request #156  and adding the test case.  I ran into the same problem so added a test case that fails without the fix.

Copied from #156 for easy access:

The change enables setting defaults for nested keys w/o those keys being shadowed by non-existent configuration values (nil) when read from a file.

For example:

```go
viper.SetDefault("nested.one","foo")
...
viper.ReadInConfig()     // { "nested" : { "two" : "bar" } }
...
```
Following the configuration read the "nested.one" will become inaccessible w/o this change.

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -788,8 +788,10 @@ func (v *Viper) find(key string) interface{} {
 		if source != nil {
 			if reflect.TypeOf(source).Kind() == reflect.Map {
 				val := v.searchMap(cast.ToStringMap(source), path[1:])
-				jww.TRACE.Println(key, "found in nested config:", val)
-				return val
+				if val != nil {
+					jww.TRACE.Println(key, "found in nested config:", val)
+					return val
+				}
 			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


xanadros:
This should fix Issue #162

kpacha:
any status on this?

ash2k:
Just ran into this bug. Can someone merge this please?

lebauce:
@spf13 This PR is the same than #156 but with the added test.

It seems like a few users are hitting this issue, could you please take an other look ?

wrouesnel:
This was literally the first issue I ran into using viper - I'd regard this as kind of critical.

lebauce:
@wrouesnel Indeed, it is critical. It makes the setting of default values completely useless in our case :-(

awishformore:
I had to rebase in another branch to solve the conflict, but this is merged now.

PR #193, opened 2016-06-02
Add support for Afero filesystems

matthieugrieger:
Hi there,

I was in the process of implementing some unit tests for a config write feature (as described in https://github.com/spf13/viper/pull/153), and noticed that Viper does not currently use the Afero filesystem framework.

It would be nice if Viper used the Afero backend as it would allow for mock filesystems in unit tests.

Let me know if I have missed anything or if something should be changed!

### DIFF
--- a/util.go
+++ b/util.go
@@ -77,7 +77,7 @@ func absPathify(inPath string) string {
 
 // Check if File / Directory Exists
 func exists(path string) (bool, error) {
-	_, err := os.Stat(path)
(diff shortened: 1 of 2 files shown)


bep:
Dang ... I just created this issue https://github.com/spf13/viper/issues/218 and then I saw this. I need this in Hugo.

matthieugrieger:
Glad I could help. :)

Issue #190, opened 2016-05-25
Binding flags and unmarshalling to nested structures


benoitmasson:
Hi,

I've run into a small issue when overriding a configuration value with a command-line flag: after unmarshalling to a struct, fields in nested sub-structures are not updated, while level-1 fields are.

For example, this piece of code:
```go
type Config struct {
	Test1  int
	Nested struct {
		Test2 int
	}
}

func main() {
	var cfg Config

	// init flags and bind to viper
	flag.Int("test1", 0, "Test1")
	flag.Int("nested.test2", 0, "Nested test2")
	flag.Parse()
	viper.BindPFlag("test1", flag.Lookup("test1"))
	viper.BindPFlag("nested.test2", flag.Lookup("nested.test2"))

	// read config
	viper.SetConfigName("config")
	viper.AddConfigPath(".")
	err := viper.ReadInConfig()
	if err != nil {
		panic(err)
	}
	err = viper.Unmarshal(&cfg)
	if err != nil {
		panic(err)
	}

	// run
	fmt.Println(viper.GetInt("test1"), viper.GetInt("nested.test2"))
	fmt.Println(cfg.Test1, cfg.Nested.Test2)
}
```
associated with this configuration file:
```toml
test1 = 11

[nested]
test2 = 12
```
produces the following output:
```sh
> ./main
11 12
11 12
> ./main --test1=21 --nested.test2=22
21 22
21 12
```
The last `nested.test2` value is correct when using `viper.GetInt("nested.test2")`, but it is not in the struct, while both values are correct for `test1`.

This may be related to Issue #168, where it also appears that the nested struct is not updated when the internal map changes.

benoitmasson:
PR #195 should have fixed this issue.

sm3142:
I'm not sure what is actually happening is what you hope for. If you dump out the resulting config with `AllSettings()`, you see that the binding has actually created an extra top level item:

    map[nested:map[test2:12] test1:21 nested.test2:22]

`viper.GetInt("nested.test2")` will actually fetch the top level item, not the nested one. The last example in [Accessing Nested Keys](https://github.com/spf13/viper#accessing-nested-keys) of the [Viper documentation](https://github.com/spf13/viper) explains why this is so:

> Lastly, if there exists a key that matches the delimited key path, its value will be returned instead.

PS: Yep, I just spent some ungodly amount of time trying to bind nested objects too. Apparently it is not supported.

benoitmasson:
[...]

PR #195, opened 2016-06-03
Nested maps

### LINKED ISSUE #168: Overriding nested keys
Hey viper team, and thanks for the amazing work on viper !
I just encountered an issue while playing around with viper, it seems to me that overriding values in nested keys doesn't update the parent value: 

With a simple nested config:
 ``` yaml
level1:
    level2:
        oldValue
 ```

 ``` go
func SampleTest(t *testing.T) {
	config := viper.New()
	config.SetConfigFile("sample-nested-conf.yaml")
	err := config.ReadInConfig()
	if err == nil {
		config.Set("level1.level2", "newValue")
		// Returns map[level2:oldValue]
		fmt.Println(config.Get("level1"))
		// Returns newValue
		fmt.Println(config.Get("level1.level2"))
	}
}
```

Outputs:
```
map[level2:oldValue]
newValue
```

"oldValue" should be overriden at all levels by Set(), and a Get() should return consistent values.

### PR DESCRIPTION
**Pull request by benoitmasson**
This PR should fix #168 and #190, among others, by updating the `Set*()` functions so that they change “deep” values in the configuration maps.

**Not ready for merge**: no code yet, only unit tests in file `overrides_test.go` (corresponding to the cases discussed in #168), to make sure we agree on the final behavior.

Please comment if you wish to change anything.

### DIFF
--- a/README.md
+++ b/README.md
@@ -458,16 +458,17 @@ Viper can access a nested field by passing a `.` delimited path of keys:
 GetString("datastore.metric.host") // (returns "127.0.0.1")
 ```
 
-This obeys the precedence rules established above; the search for the root key
-(in this example, `datastore`) will cascade through the remaining configuration
-registries until found. The search for the sub-keys (`metric` and `host`),
-however, will not.
+This obeys the precedence rules established above; the search for the path
+will cascade through the remaining configuration registries until found.
 
-For example, if the `metric` key was not defined in the configuration loaded
-from file, but was defined in the defaults, Viper would return the zero value.
+For example, given this configuration file, both `datastore.metric.host` and
+`datastore.metric.port` are already defined (and may be overridden). If in addition
+`datastore.metric.protocol` was defined in the defaults, Viper would also find it.
 
-On the other hand, if the primary key was not defined, Viper would go through
-the remaining registries looking for it.
+However, if `datastore.metric` was overridden (by a flag, an environment variable,
+the `Set()` method, …) with an immediate value, then all sub-keys of
+`datastore.metric` become undefined, they are “shadowed” by the higher-priority
+configuration level.
 
 Lastly, if there exists a key that matches the delimited key path, its value
 will be returned instead. E.g.
@@ -491,7 +492,7 @@ will be returned instead. E.g.
     }
 }
 
-GetString("datastore.metric.host") //returns "0.0.0.0"
+GetString("datastore.metric.host") // returns "0.0.0.0"
 ```
 
 ### Extract sub-tree
--- a/overrides_test.go
+++ b/overrides_test.go
@@ -1,0 +1,173 @@
+package viper
+
+import (
+	"fmt"
+	"strings"
+	"testing"
+
+	"github.com/spf13/cast"
+	"github.com/stretchr/testify/assert"
+)
+
+type layer int
+
+const (
+	defaultLayer layer = iota + 1
+	overrideLayer
+)
+
+func TestNestedOverrides(t *testing.T) {
+	assert := assert.New(t)
+	var v *Viper
+
+	// Case 0: value overridden by a value
+	overrideDefault(assert, "tom", 10, "tom", 20) // "tom" is first given 10 as default value, then overridden by 20
+	override(assert, "tom", 10, "tom", 20)        // "tom" is first given value 10, then overridden by 20
+	overrideDefault(assert, "tom.age", 10, "tom.age", 20)
+	override(assert, "tom.age", 10, "tom.age", 20)
+	overrideDefault(assert, "sawyer.tom.age", 10, "sawyer.tom.age", 20)
+	override(assert, "sawyer.tom.age", 10, "sawyer.tom.age", 20)
+
+	// Case 1: key:value overridden by a value
+	v = overrideDefault(assert, "tom.age", 10, "tom", "boy") // "tom.age" is first given 10 as default value, then "tom" is overridden by "boy"
+	assert.Nil(v.Get("tom.age"))                             // "tom.age" should not exist anymore
+	v = override(assert, "tom.age", 10, "tom", "boy")
+	assert.Nil(v.Get("tom.age"))
+
+	// Case 2: value overridden by a key:value
+	overrideDefault(assert, "tom", "boy", "tom.age", 10) // "tom" is first given "boy" as default value, then "tom" is overridden by map{"age":10}
+	override(assert, "tom.age", 10, "tom", "boy")
+
+	// Case 3: key:value overridden by a key:value
+	v = overrideDefault(assert, "tom.size", 4, "tom.age", 10)
+	assert.Equal(4, v.Get("tom.size")) // value should still be reachable
+	v = override(assert, "tom.size", 4, "tom.age", 10)
+	assert.Equal(4, v.Get("tom.size"))
+	deepCheckValue(assert, v, overrideLayer, []string{"tom", "size"}, 4)
+
+	// Case 4: key:value overridden by a map
+	v = overrideDefault(assert, "tom.size", 4, "tom", map[string]interface{}{"age": 10}) // "tom.size" is first given "4" as default value, then "tom" is overridden by map{"age":10}
+	assert.Equal(4, v.Get("tom.size"))                                                   // "tom.size" should still be reachable
+	assert.Equal(10, v.Get("tom.age"))                                                   // new value should be there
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

**Comment by benoitmasson** [comment]
PR rebased onto `master` and completed:

* commit `adeffda` fixes Issue #168 (`Set()` and `SetDefault()` now handle nested keys correctly, and according to the tests from `overrides_test.go`)

* commit `d8a62ff` fixes Issue #190 (`AllSettings()` takes nested flags into account, so that `Unmarshal` handles correctly overriding with flags)

* commit `639fa92` fixes `Get()`, which used to return a value for `foo.bar.baz` when `foo.bar` was defined but `foo.bar.baz` wasn't

* commit `bf1de71` fixes case insensivity for nested config elements (`insensitiviseMaps()` used to stop at depth-1 keys, now recurses through the whole structure)

It should now be ready for merge, but feel free to comment...

**Comment by rybit** [comment]
Am I understanding this PR right. It will address issues when you have code like this: 

``` go 
type Config struct {
  Port      int
  LogConfig struct {
    Level string
  } `mapstructure:"log_config"`
}

func LoadConfig()(*Config, error) {
  viper.SetEnvPrefix("NF")
  viper.SetEnvKeyReplacer(strings.NewReplacer(".", "_"))
  viper.AutomaticEnv()
 
  config := new(Config)
  if err := viper.Marshal(config); err != nil {
    return nil, err
  }
  return config, nil
}
```

and let it be overriden by env vars like `NF_LOG_CONFIG_LEVEL`? Just like you can override port with `NF_PORT`?

benoitmasson:
Hi @rybit,

You got it right, that's the idea for different contexts (overrides and flags). I didn't deal with env vars since I don't have a use for them, but that's interesting too.

I've tried to execute your code sample, and unfortunately the answer is no, it doesn't work... I'm investigating this and I'll let you know if it can be fixed easily.

benoitmasson:
I've added a new commit `9141a0a`, which replaces `6457afe` in that it provides a more general solution to the unmarshalling of overridden nested values.

It now works with both flags and environment variables, and probably everything else, since `viper.AllSettings()` now exclusively relies on the result of the `viper.Get(key)` function for all nested keys found in the configuration.

@rybit your code sample now works, just make sure to use `viper.Unmarshal()` function, and to set a default value to `port` and `log_config.level`, with `viper.SetDefault()` for example (otherwise viper doesn't know these config values exist and doesn't look for corresponding values in the environment).

rybit:
This is great! I have been poking at it!

Before I have been using [a (terrible) method](https://github.com/rybit/config_example/blob/master/conf/reflect.go#L13) to walk the structure and populate the fields by calling the 'GetXXX' methods. But it did mean that you loaded each value, including zero values. 

I was hoping that you wouldn't have to set defaults, that it would just return the zero value or the overridden value. It means you can add a value to the struct and it will get the zero or the right override. For instance, if you're setting defaults in the pflag declaration. Or if you're going to default to a zero value. 

Thanks for the work though! I am excited for when this goes in :)

awishformore:
I'll try to take a look at this soon, could you rebase the branch in the meantime?

benoitmasson:
Great, thanks!

The rebase is complete, however the test you recently added fails: `func TestShadowedNestedValue(t *testing.T)` (commit `2f6a414`).

Indeed, I followed the convention written in the README file, which stipulates that:

> This obeys the precedence rules established above; the search for the root key
(in this example, `datastore`) will cascade through the remaining configuration
registries until found. The search for the sub-keys (`metric` and `host`),
however, will not.

> For example, if the `metric` key was not defined in the configuration loaded
from file, but was defined in the defaults, Viper would return the zero value.

i.e., since `clothing` is first found in the `config` map, all sub-elements will be looked for in this map. Since `clothing.shirt` exists only in the `defaults` map, it should remain undefined.

This test worked previously (before my commits), because the whole “nested maps” stuff was buggy. My changes make `Viper.find()` work as described and `TestShadowedNestedValue()` fails.

I have two options to choose from before patching and pushing:

* either change the test in `TestShadowedNestedValue()` to `assert.Nil(t, Get("clothing.shirt"))`;

* or update the behavior of `find()` (and the message in the README) to make it look for `clothing.shirt` in all maps.

What should I do? I'd go for solution 1, which is has a much lower impact on the package (and on its possible usages).

awishformore:
You bring up a good point, I'll try to get some feedback about this, since I can't decide this on my own.

awishformore:
It's hard to argue with the README, but after getting some feedback, it seems my initial understanding is correct and the documentation should probably be updated. Basically, the precedence should overwrite at times and replace at times.

Quoting the response from @spf13 :

```
Default = {
    x: foo,
    y: {
        z: bar
   }
}
```

If I specify in the config:
```
y: yellow
```
then the final result would be
```
{
    x: foo,
    y: yellow
}
```

If I specify in the config:
```
y.n:yellow
```
then the final result would be:
```
{
    x: foo,
    y: {
        n: yellow,
        z: bar
   }
}
```

So this means that solution 2 would be the correct approach. Apologies about the confusion.

benoitmasson:
No problem, that was my first understanding too, until I saw the README :-)

However, that means changing the `find()` function accordingly, which may be quite a bit of work from what I remember. I'll fix that as  soon as possible, but that will probably take a few days. I'll keep you informed in this thread.

Thanks for the clarification anyway, in the end it will probably more logical that way.

benoitmasson:
Now, this is finally looking good!

To make it short, I've made a bunch of fixes, with 2 major improvements:

* `Set*()` functions correctly insert nested values in “deep” maps, and `Get()` retrieves them using the correct order of priority (issue #168). This was achieved by a partial rewrite of the underlying `find()` function.
    It takes into account shadowing, as you requested earlier, and precedence of keys containing dots, as specified in the README file.

* `AllKeys()` and `AllSettings()` have been rewritten, they now perform correctly the “configuration merge” (the “final result” you mentioned), taking into account all configuration levels with their priority, shadowing, etc.
    The main consequence is that `Unmarshall()` works well for nested structures (issue #190)

README has been updated, too, according to our discussion, and should now consistent with this implementation.

--
More detailed overview of the most significant changes:

* commit `da01ceb` fixes `Get()`, which used to return a value for `foo.bar.baz` when `foo.bar` was defined but `foo.bar.baz` wasn't

* commit `9c83c7e` fixes case insensivity for nested config elements (`insensitiviseMaps()` used to stop at depth-1 keys, now recurses through the whole structure)

* commits `5d7fd96`, `73ba7a3` and `6a8a4fb` add more tests, to reflect what is claimed in he README file

* commits `32c8b5f` and `a841f65` fix Issue #168 (`Set()` and `SetDefault()` now handle nested keys correctly, and according to the tests from `overrides_test.go`)

* commit `f513f5f` ensures that properties containing dots generate nested maps when parsed.

* commit `ae72cd4` performs all the heavy work about `find()` refactoring: nested keys reachable in any map, shadowing, … to comply with [your comment](#issuecomment-248708337)

* commit `3b8c1c9` fixes Issue #190 (`AllSettings()` takes nested flags into account, so that `Unmarshal()` handles correctly the different configuration levels)

I hope this can be merged shortly, it's becoming harder and harder to rebase… So please let me know if I can be of any help.

Thanks a lot!

Niondir:
Please merge this Pull Request, I could also need this. I guess it will also fix missing Environment variable overrides in `viper.AllSettings()`.

awishformore:
I will review this tomorrow and see if I can merge right away.

awishformore:
I checked the tests and most of the code. I also used this version in a private project and the improved behaviour is a bliss. I will go ahead and merge now.

One comment I have is whether setting a key to two different maps should maybe merge them instead of replacing the first one with the second one. This is of course a question of semantics and a design decision; I just feel it would possibly make more sense, as there is no `viper.Add` option to do this.

Thanks a lot for the great work.

arekkas:
@benoitmasson @awishformore this pull requests breaks `viper.BindEnv`, because it isn't being set when no default value is set:


Does not work:
```
	viper.BindEnv("HOST")

		fmt.Printf(`Config file not found because "%s"`, err)
		fmt.Println("")
	}

	if err := viper.Unmarshal(c); err != nil {
		fatal(fmt.Sprintf("Could not read config because %s.", err))
	}
```

Does work:
```
	viper.BindEnv("HOST")
	viper.SetDefault("HOST", "")

		fmt.Printf(`Config file not found because "%s"`, err)
		fmt.Println("")
	}

	if err := viper.Unmarshal(c); err != nil {
		fatal(fmt.Sprintf("Could not read config because %s.", err))
	}
```

benoitmasson:
OK @arekkas, I'll have a look.

PR #199, opened 2016-06-08
Changing config name should also reset file name cache

rndstr:
I stumbled over this when trying to merge multiple configs.

```
viper.SetConfigName("default")
err := viper.MergeInConfig()
```
which caches file path resolvemenet in `v.configFile`

and then
```
viper.SetConfigName("prod")
err := viper.MergeInConfig()
```

which reuses `v.configFile` without updating it accordingly to the new name.

See https://github.com/spf13/viper/blob/c1ccc378a054ea8d4e38d8c67f6938d4760b53dd/viper.go#L1240


mehcode:
This would be awesome! Let's get this merged in.

Issue #105, opened 2015-07-31
Durations don't work in viper.Marshal


kelvl:
Hi, I am unable to Marshal (or Unmarshal) Duration fields whereas, `GetDuration("fieldName")` works

I've attached an simple test case

```go
package main

import (
	"bytes"
	"testing"
	"time"

	"github.com/spf13/viper"
)

func TestDurationConfig(t *testing.T) {
	type DurationConfig struct {
		testduration time.Duration
	}

	var testYaml = []byte(`
               TestDuration: 5ms
        `)

	viper.SetConfigType("yaml")

	viper.ReadConfig(bytes.NewBuffer(testYaml))

	t.Logf("All keys: %#v", viper.AllSettings())

        // this works
	if viper.GetDuration("testduration") != 5*time.Millisecond {
		t.Errorf("Duration is not 5ms - %v", viper.GetDuration("testduration"))
	}

	var config = &DurationConfig{}

	viper.Marshal(config)

        // this doesn't
	if config == nil || config.testduration != 5*time.Millisecond {
		t.Errorf("Duration config is not 5ms - %#v", config)
	}
}

```

emilevauge:
Any chance to get this bug corrected?

TrueFurby:
+1

yauhen-l:
https://github.com/spf13/viper/pull/205

PR #205, opened 2016-07-09
Issue-105 Support time.Duration in viper.Unmarshal

yauhen-l:
Just add DecodeHook to DecoderConfig.
Allows to unmarshal into struct with time.Duration fields:
```
package main

import (
	"bytes"
	"fmt"
	"github.com/spf13/viper"
	"time"
)

type Config struct {
	Timeout time.Duration
	String  string
}

func main() {
	var cfg Config

	viper.SetConfigType("yaml")

	var yamlExample = []byte(`
Timeout: 200ms
String: it works!
`)

	viper.ReadConfig(bytes.NewBuffer(yamlExample))
	viper.Unmarshal(&cfg)

	// {Timeout:200ms String:it works!}
	fmt.Printf("%+v\n", cfg)
}
```

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -611,7 +611,19 @@ func (v *Viper) UnmarshalKey(key string, rawVal interface{}) error {
 // on the fields of the structure are properly set.
 func Unmarshal(rawVal interface{}) error { return v.Unmarshal(rawVal) }
 func (v *Viper) Unmarshal(rawVal interface{}) error {
-	err := mapstructure.WeakDecode(v.AllSettings(), rawVal)
+	config := &mapstructure.DecoderConfig{
+		Metadata:         nil,
+		Result:           rawVal,
+		WeaklyTypedInput: true,
+		DecodeHook:       mapstructure.StringToTimeDurationHookFunc(),
+	}
+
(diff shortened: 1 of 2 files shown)


jyellick:
This changeset looks very much like the code used in [`UnmashalExact`](https://github.com/yauhen-l/viper/blob/aebf0fcdc54f2101d72d3704b9cea0423fe66450/viper.go#L656).

Fixing Issue #105 is something I'd certainly like to see done, but could this patch be refactored so that `Unmarshal` and `UnmarshalExact` are combined to use the same codepath?  This patch, as is, I believe will still produce errors in `UnmarshalExact`.

awishformore:
I will try to take a look at this PR soon. In the meantime, if the suggestion by @jyellick could be implemented, thaht would be great.

awishformore:
@yauhen-l it seems straight forward enough to add the hook to the `UnmarshalExact` as well, since it already uses a custom configuration. Could you add that?

yauhen-l:
@awishformore, I will take a look in the evening

yauhen-l:
@awishformore, @jyellick please review

PR #208, opened 2016-07-18
Optional TOML Support (disabled by default)

akutz:
This patch makes TOML support optional, disabling it by default. To enable support include the Golang build tag `toml`. For example, the following command builds Viper without TOML support:

```sh
$ go build
```

Whereas this command will build Viper, along with support for parsing TOML:

```sh
$ go build -tags toml
```

The reason for making TOML optional is due to the dependency upon the BurntSushi package that is used to unmarshal TOML content. The package is licensed under the [WTFPL license](http://www.wtfpl.net/) and incompatible with the Kubernetes project. Pull requests submitted to K8 that transitively include a dependency upon packages licensed under the WTFPL license [are denied](https://github.com/kubernetes/kubernetes/pull/28599#discussion_r71100997). The reason given links to [this discussion](https://news.ycombinator.com/item?id=5733050).

To this end, until such time an alternative package is deemed acceptable for parsing TOML content, this patch would make TOML an optional component of Viper, enabled only when explicitly requested via the build tag `toml`.

To prove no TOML is present, from the root of the branch execute:

```bash
$ go list -f '{{join .Deps "\n"}}' | grep toml
```

No matching packages are found. Using the build tag `toml` will reintroduce the now optional component:

```bash
$ go list -tags toml -f '{{join .Deps "\n"}}' | grep toml
github.com/BurntSushi/toml
```

### DIFF
--- a/util.go
+++ b/util.go
@@ -21,7 +21,6 @@ import (
 	"strings"
 	"unicode"
 
-	"github.com/BurntSushi/toml"
 	"github.com/hashicorp/hcl"
 	"github.com/magiconair/properties"
 	"github.com/spf13/cast"
@@ -155,7 +154,7 @@ func unmarshallConfigReader(in io.Reader, c map[string]interface{}, configType s
 		}
 
 	case "toml":
-		if _, err := toml.Decode(buf.String(), &c); err != nil {
+		if err := unmarshalTOML(buf.String(), &c); err != nil {
 			return ConfigParseError{err}
 		}
 
--- a/util_notoml.go
+++ b/util_notoml.go
@@ -1,0 +1,17 @@
+// +build !toml
+
+// Copyright © 2014 Steve Francia <spf@spf13.com>.
+//
+// Use of this source code is governed by an MIT-style
+// license that can be found in the LICENSE file.
+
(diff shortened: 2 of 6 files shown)


akutz:
Hi @spf13,

Doh! It looks like I did not account for the tests that expect the TOML parser to have executed. I will update the PR.

akutz:
Hi @spf13,

I can provide far more information as to why this is necessary; one of my colleagues has a JD, and he typed up quite a summary relating to the issues of the transitive license in question. I bring this up because in my work to update the tests, more work than with which I felt comfortable performing was required, and I'm not sure it meets your standards for this sort of behavior.

akutz:
Hi @spf13,

Any thoughts on this? If nothing else, could we make this a branch in *this* repository so projects that use GoDeps can point to this import path and the Git ref of the branch's tip with this change? As it is, it doesn't appear that something like K8 can use this PR from my fork since K8 uses GoDeps, and that seems to only work with the proper import path (unlike Glide which enables you to point to a fork).

akutz:
Closed after discussion with @spf13 in favor of #217.

PR #214, opened 2016-07-27
replace SetEnvReplacer by SetEnvKeyReplacer

rougepied:
It may be a typo: There is no SetEnvReplacer in viper.go but a SetEnvKeyReplacer function that match the description in README.md


sagikazarmark:
This should already be fixed

PR #217, opened 2016-08-02
TOML Library Licensing

akutz:
This patch updates the package used for parsing TOML content from `github.com/BurntSushi/toml` to `github.com/pelletier/go-toml` as the latter uses a more accepted OSS license (MIT), enabling the inclusion of Viper or projects that depend on Viper in projects that have licensing requirements incongruent with the license of the previous TOML package.

This patch replaces the PR #208 after discussing the matter with @spf13 and deciding to update the TOML parser instead of making TOML build-optional.

### DIFF
--- a/util.go
+++ b/util.go
@@ -21,9 +21,9 @@ import (
 	"strings"
 	"unicode"
 
-	"github.com/BurntSushi/toml"
 	"github.com/hashicorp/hcl"
 	"github.com/magiconair/properties"
+	toml "github.com/pelletier/go-toml"
 	"github.com/spf13/cast"
 	jww "github.com/spf13/jwalterweatherman"
 	"gopkg.in/yaml.v2"
@@ -155,8 +155,13 @@ func unmarshallConfigReader(in io.Reader, c map[string]interface{}, configType s
 		}
 
 	case "toml":
-		if _, err := toml.Decode(buf.String(), &c); err != nil {
+		tree, err := toml.LoadReader(buf)
+		if err != nil {
 			return ConfigParseError{err}
+		}
+		tmap := tree.ToMap()
+		for k, v := range tmap {
+			c[k] = v
 		}
 
 	case "properties", "props", "prop":


akutz:
Hi @spf13, I [commented](https://github.com/spf13/viper/commit/abafbf243b9583ac1360ae4ef680d8acaa361089#commitcomment-18578325) on the reversion of this PR as well as introduced mysefl to @bep. Bjørn, what are your, as well as Steve's, thoughts on allowing a branch of this PR on the primary repo for the time-being in order to support Godeps-enabled projects such as K8. They cannot accept our contributions for now due to Viper and its dependency on BurntSushi (the TOML parser).

bep:
I suggest you keep your own fork, then we can sync with this one once this can be resolved in a more controlled manner.

akutz:
Hi @bep,

I have my own fork, but the issue is Godeps doesn't allow pointing at a forked repo, only a ref on the primary repo, and K8 uses Godeps. I apologize if that was not clear above.

akutz:
Hi @bep,

I realize the tone in [this comment](https://github.com/spf13/viper/commit/abafbf243b9583ac1360ae4ef680d8acaa361089#commitcomment-18578325) on the revert commit may not have been communicated well via text. I was attempting humor, and clearly failed. I was using colloquialisms that may not survive the language barrier well. I was not at all put off or upset by anything you did; it's just my way of saying "Gosh darnit, we needed that, you betcha." in a, hopefully, de-arming sort of way. 

Anyway, I don't wish to get off on the wrong foot with you and wanted to be very clear I appreciate you merging the PR and am very attentive to the reason you reverted it as well. I want to work with you and @spf13 to try and find a solution that helps everyone. To that end, I'll restate that I simply don't think testing a PR against all packages that import Viper is feasible. I do not mind, however, taking the burden of identifying the reason the inclusion of this new TOML parser broke Hugo. I'm the one pushing the PR so I should accept the burden of solving the problem it causes.

Thank you again for your help.

bep:
I'm not worried about Hugo, I probably could have patched up Hugo in a few minutes -- but the error indicated structural diff in the data types produced, so I got a little bit worried about others.

akutz:
Agreed. Please see my comment [here](https://github.com/spf13/viper/commit/abafbf243b9583ac1360ae4ef680d8acaa361089#commitcomment-18579887). Inlining it for our convenience:

> Hi @bep,
> 
> Please see my comment in the PR. I'm happy to handle the debugging of this issue. Regarding the data structures -- well, that's a gap in Viper's tests to be honest. Viper already does test that TOML returns parsed data types. Apparently though Viper doesn't test some form of map type that the old parser created differently than this new one. We should add a test that checks for that. I believe it is on Viper to validate the different expected types that the parsers produce. FWIW, I have an outstanding PR to Go-YAML for this very reason. It is in support of this branch on my fork I've had since October of 2015...for Viper. Since the standard Go-YAML project uses map[interface{}]interface{} by default, it means Viper cannot successfully marshal structures parsed from YAML to JSON using the Go StdLib JSON encoder since it enforces a rule that a map's key must be of type string.
> 
> So yes, I hear you. I just think it is on Viper to validate that parsed types stay the same. There was nothing to keep the BurntSushi library, for example, from changing its types and causing the same bug. The current set of tests in Viper should be augmented to guarantee consistent, underlying types arrive from the parsers.

vladimirvivien:
Hi @bep ,
I am looking into this issue along with @akutz .  Can you let me know which test in Hugo fails and how to reproduce ?  Thank you.

akutz:
Vlad, please see the revert commit for what caused Hugo to fail.

vladimirvivien:
@akutz
I saw the note in the revert commit.  Nevertheless, I wanted to clarify this is not a build/test error.
After some investigation, I was able to reproduce it against the site `docs` directory in the project  with `hugo` cmd.

bep:
Continue the discussion in https://github.com/spf13/viper/issues/225

PR #219, opened 2016-08-08
Fix README entry

w3st3ry:
Fix small fail in README.

Signed-off-by: Valentin Pichard <valentin.pichard@corp.ovh.com>


awishformore:
Done with #227

PR #220, opened 2016-08-08
Throw a ConfigFileNotFoundError in certain cases

r2d4:
When the config file isn't found, it would be useful to see the paths
that viper is searching.  As it is now, the UnsupportedConfigError is
thrown when the file isn't found, and doesn't return useful information.

I also added an additional test for the UnsupportedConfigError, as it wasn't covered after my change.


awishformore:
#161 is a cleaner implementation of this change in my opinion. If it is missing something you need, I'd kindly ask you to comment in that PR.

awishformore:
This has now been merged as part of #161

PR #231, opened 2016-08-25
Added default parameter on getters to return it in case it doesn't exist

rogersole:


rogersole:
Closing PR. Created another with tests.

PR #232, opened 2016-08-25
Added default parameter on getters to return it in case it doesn't exist

rogersole:
Added getters allowing default values:

- `func GetDefaultString(key string, defaultValue string) string`
- `func GetDefaultBoolean(key string, defaultValue bool) bool`
- `func GetDefaultInt(key string, defaultValue int) int`
- `func GetDefaultInt64(key string, defaultValue int64) int64`
- `func GetDefaultFloat64(key string, defaultValue float64) float64`
- `func GetDefaultTime(key string, defaultValue time.Time) time.Time`
- `func GetDefaultDuration(key string, defaultValue time.Duration) time.Duration`
- `func GetDefaultStringSlice(key string, defaultValue []string) []string`
- `func GetDefaultStringMap(key string, defaultValue map[string]interface{}) map[string]interface{}`
- `func GetDefaultStringMapString(key string, defaultValue map[string]string) map[string]string`
- `func GetDefaultStringMapStringSlice(key string, defaultValue map[string][]string) map[string][]string`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
+*.iml
+.idea/
 # Compiled Object files, Static and Dynamic libs (Shared Objects)
 *.o
 *.a
--- a/README.md
+++ b/README.md
@@ -406,14 +406,23 @@ The following functions and methods exist:
 
  * `Get(key string) : interface{}`
  * `GetBool(key string) : bool`
+ * `GetDefaultBool(key string, value bool) : bool`
  * `GetFloat64(key string) : float64`
+ * `GetDefaultFloat64(key string, value float64) : float64`
  * `GetInt(key string) : int`
+ * `GetDefaultInt(key string, value int) : int`
  * `GetString(key string) : string`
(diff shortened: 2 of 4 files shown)


bep:
You know you there is an ´viper.SetDefaultValue`?

rogersole:
Hi @bep, yes I know, but sometimes you don't want to define all the `SetDefaultValue` variables in a different code line. Just adding the option to have it in a oneliner.

bep:
Other may chime in, but this doesn't seem to be worth it to me. If Go had a good way of adding optional arguments (other than varags), maybe ...

awishformore:
I don't think it's necessary, seems like code bloat to me.

taybin:
I'd like to see Sub() have a default value of an empty map[].

awishformore:
I will close this unless someone chimes in with a real use case for this. Feel free to keep commenting.

PR #235, opened 2016-09-01
Add tenant feature

lucasvmiguel:
I needed to implement this feature in my company project, I thought interesting if Viper had this feature. (the feature is explained in README)

I'm beginner in golang , so sorry about any wrong code . =)

Any doubt talk to me...

### DIFF
--- a/README.md
+++ b/README.md
@@ -599,6 +599,56 @@ y.SetDefault("ContentDir", "foobar")
 When working with multiple vipers, it is up to the user to keep track of the
 different vipers.
 
+### Working with tenant
+
+Sometime you want your config to be override by a more specific one, use tenant for it.
+
+Example:
+
+config file:
+```json
+[shared]
+name = "name_shared"
+
+[tenant_a]
+name = "name_a"
+
+[tenant_b]
+
+```
+
+code:
+```go
+viper.SetTenantDefault("shared")
+viper.SetConfigName("teste")
+viper.AddConfigPath(".")
+viper.ReadInConfig()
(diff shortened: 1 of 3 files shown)


awishformore:
I'm not really sure what the use case for this is. Couldn't the example in the README just as easily be achieved by setting a default for `tenant_b`, which remains in place unless another config overwrites it?

Maybe this didn't work before because of the bug that was fixed in #156 ?

lucasvmiguel:
For example, at my job, we have more than one brand sharing the same code, we need share some variable between these brands, it's easier create one file with variable shared and variables specifics than create a lot of files with a lot in common.

something like that:

[shared]
database = "database_url"

[brand_a]
name = "brand_a"

[brand_b]
name = "brand_a"


these two brands have the same database url and differents names, it's good create just one file like this, because i don't want create two files with the same information.

bep:
@lucasvmiguel this sounds like a messy and error prone solution. I would rather go with the multiple files, aka. `common-config.toml`, `brand_a-config.toml` etc.

awishformore:
I understand what you are using it for, but I don't see the benefit of using this "tenant" feature over what @bep describes. You can have a base configuration loaded with the shared settings (`shared.yml`), and then you can use it as basis to import `brand_a.yml` and `brand_b.yml` configurations, which will only overwrite what has been set. This produces the exact same effect, using the already present functionality. As I see no added value in this PR, I will close it now. If I missed anything, don't hesitate to comment again.

Issue #200, opened 2016-06-12
String Slices from pflags are not handled correctly by viper.


aarondl:
The slice of strings just contains one element that is the stringified version of the string
ie: `[]string{"[firstarg,secondarg]"}`
or in the case of nil/empty: `[]string{"[]"}`

It appears that this is coming from the Pflag libraries `ValueString()` function. I've included a patch that "helps" the issue but I suspect it's a little bit more deep-seated as the cast package also doesn't seem to handle this case well at all hence I'm not making trying to make a PR for this.

For others hitting this issue the workaround is to simply use pflags directly for string slices until this is fixed.

``` diff
diff --git a/viper.go b/viper.go
index a2633b0..d1cce44 100644
--- a/viper.go
+++ b/viper.go
@@ -479,6 +479,8 @@ func (v *Viper) Get(key string) interface{} {
                                val = cast.ToInt(flag.ValueString())
                        case "bool":
                                val = cast.ToBool(flag.ValueString())
+                       case "stringSlice":
+                               val = []string{}
                        default:
                                val = flag.ValueString()
                        }
@@ -742,6 +744,8 @@ func (v *Viper) find(key string) interface{} {
                        return cast.ToInt(flag.ValueString())
                case "bool":
                        return cast.ToBool(flag.ValueString())
+               case "stringSlice":
+                       return strings.Split(strings.Trim(flag.ValueString(), "[]"), ",")
                default:
                        return flag.ValueString()
                }
```

Example program that shows the the failure:

``` go
package main

import (
    "fmt"
    "os"

    "github.com/davecgh/go-spew/spew"
    "github.com/spf13/cobra"
    "github.com/spf13/viper"
)

func main() {
    var rootCmd = &cobra.Command{
        Run: run,
    }

    rootCmd.PersistentFlags().StringSliceP("thing", "t", nil, "")
    viper.BindPFlags(rootCmd.PersistentFlags())

    if err := rootCmd.Execute(); err != nil {
        fmt.Printf("\n%+v\n", err)
        os.Exit(1)
    }
}

func run(cmd *cobra.Command, args []string) {
    str := viper.Get("thing")
    slice := viper.GetStringSlice("thing")
    slicePFlag, _ := cmd.PersistentFlags().GetStringSlice("thing")

    spew.Dump(str, slice, slicePFlag)
}
```

Output (annotated for clarity) of `go run main.go`

``` text
str   = (string) (len=2) "[]"
slice = ([]string) (len=1 cap=1) {
[...]

PR #240, opened 2016-09-22
Fix retrieval of pflag stringSlice

### LINKED ISSUE #112: BindPFlag don't work properly with pflag.StringSlice
An open and close bracket are added to the first and last item of a StringSlice when it's updated by a flag.

Here's some test code that demonstrates the issue:
```
package main

import (
	"fmt"

	jww "github.com/spf13/jwalterweatherman"
	"github.com/spf13/pflag"
	"github.com/spf13/viper"
)

func main() {
	jww.SetStdoutThreshold(jww.LevelDebug)

	viper.SetConfigName("config")
	viper.AddConfigPath(".")
	err := viper.ReadInConfig()
	if err != nil {
		panic(fmt.Errorf("fatal error: %s \n", err))
	}

	pflag.StringSlice("foo", []string{}, "Test flag for StringSlice")
	err = viper.BindPFlag("foo", pflag.Lookup("foo"))
	if err != nil {
		panic(fmt.Errorf("fatal error: %s \n", err))
	}
	pflag.Parse()

	for _, value := range viper.GetStringSlice("foo") {
		fmt.Println(value)
	}
}
```

And here's the output:
```
(04:14:41) brendan @ ~/workspace/golang/src/github.com/b0d0nne11/test 
→ go run test.go 
INFO: 2015/09/03 Trying to resolve absolute path to .
INFO: 2015/09/03 adding /home/brendan/workspace/golang/src/github.com/b0d0nne11/test to paths to search
INFO: 2015/09/03 Attempting to read in config file
INFO: 2015/09/03 Searching for config in  [/home/brendan/workspace/golang/src/github.com/b0d0nne11/test]
DEBUG: 2015/09/03 Searching for config in  /home/brendan/workspace/golang/src/github.com/b0d0nne11/test
DEBUG: 2015/09/03 Checking for /home/brendan/workspace/golang/src/github.com/b0d0nne11/test/config.json
DEBUG: 2015/09/03 Checking for /home/brendan/workspace/golang/src/github.com/b0d0nne11/test/config.toml
DEBUG: 2015/09/03 Checking for /home/brendan/workspace/golang/src/github.com/b0d0nne11/test/config.yaml
DEBUG: 2015/09/03 Found:  /home/brendan/workspace/golang/src/github.com/b0d0nne11/test/config.yaml
DEBUG: 2015/09/03 ToStringSliceE called on type: []interface {}
DEBUG: 2015/09/03 ToStringE called on type: string
DEBUG: 2015/09/03 ToStringE called on type: string
bar
baz

(04:14:52) brendan @ ~/workspace/golang/src/github.com/b0d0nne11/test 
→ go run test.go --foo='bar baz'
INFO: 2015/09/03 Trying to resolve absolute path to .
INFO: 2015/09/03 adding /home/brendan/workspace/golang/src/github.com/b0d0nne11/test to paths to search
INFO: 2015/09/03 Attempting to read in config file
INFO: 2015/09/03 Searching for config in  [/home/brendan/workspace/golang/src/github.com/b0d0nne11/test]
DEBUG: 2015/09/03 Searching for config in  /home/brendan/workspace/golang/src/github.com/b0d0nne11/test
DEBUG: 2015/09/03 Checking for /home/brendan/workspace/golang/src/github.com/b0d0nne11/test/config.json
DEBUG: 2015/09/03 Checking for /home/brendan/workspace/golang/src/github.com/b0d0nne11/test/config.toml
DEBUG: 2015/09/03 Checking for /home/brendan/workspace/golang/src/github.com/b0d0nne11/test/config.yaml
DEBUG: 2015/09/03 Found:  /home/brendan/workspace/golang/src/github.com/b0d0nne11/test/config.yaml
DEBUG: 2015/09/03 ToStringE called on type: string
DEBUG: 2015/09/03 ToStringSliceE called on type: string
[bar
baz]
```

moorereason:
With some extra trace log formatting changes thrown into the mix.

Fixes #112

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -749,12 +749,15 @@ func (v *Viper) find(key string) interface{} {
 	// PFlag Override first
 	flag, exists := v.pflags[key]
 	if exists && flag.HasChanged() {
-		jww.TRACE.Println(key, "found in override (via pflag):", flag.ValueString())
+		jww.TRACE.Printf("%q found in pflag override (%s): %s", key, flag.ValueType(), flag.ValueString())
 		switch flag.ValueType() {
 		case "int", "int8", "int16", "int32", "int64":
 			return cast.ToInt(flag.ValueString())
 		case "bool":
 			return cast.ToBool(flag.ValueString())
+		case "stringSlice":
+			s := strings.TrimPrefix(flag.ValueString(), "[")
+			return strings.TrimSuffix(s, "]")
 		default:
 			return flag.ValueString()
 		}
@@ -762,7 +765,7 @@ func (v *Viper) find(key string) interface{} {
 
 	val, exists = v.override[key]
 	if exists {
-		jww.TRACE.Println(key, "found in override:", val)
+		jww.TRACE.Printf("%q found in override: %s", key, val)
 		return val
 	}
 
@@ -770,24 +773,24 @@ func (v *Viper) find(key string) interface{} {
 		// even if it hasn't been registered, if automaticEnv is used,
 		// check any Get request
 		if val = v.getEnv(v.mergeWithEnvPrefix(key)); val != "" {
-			jww.TRACE.Println(key, "found in environment with val:", val)
+			jww.TRACE.Printf("%q found in environment: %s", key, val)
 			return val
 		}
 	}
 
 	envkey, exists := v.env[key]
 	if exists {
-		jww.TRACE.Println(key, "registered as env var", envkey)
+		jww.TRACE.Printf("%q registered as env var %q", key, envkey)
 		if val = v.getEnv(envkey); val != "" {
-			jww.TRACE.Println(envkey, "found in environment with val:", val)
+			jww.TRACE.Printf("%q found in environment: %s", envkey, val)
 			return val
 		}
-		jww.TRACE.Println(envkey, "env value unset:")
+		jww.TRACE.Printf("%q env value unset", envkey)
 	}
 
 	val, exists = v.config[key]
 	if exists {
-		jww.TRACE.Println(key, "found in config:", val)
+		jww.TRACE.Printf("%q found in config (%T): %s", key, val, val)
 		return val
 	}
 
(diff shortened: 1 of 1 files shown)


awishformore:
Don't see any issues with this, thank you for your contribution.

aarondl:
@awishformore @moorereason 

This PR seems to fix #200  (haven't tested)

But note that I changed two hunks in my workaround, one in `find()` as in this PR (pretty much the same code) and one in `Get()` . I wonder if the second one in `Get()` is necessary. Can anyone who knows the code base advise on this? The code I put in that case is probably wrong, but the fact that there's a similar type check going on there makes me suspicious. I just want to make sure we covered all the possibilities.

If it's truly done, then we should probably close #200 as well.

awishformore:
I'll try to take a look soon.

viper.go
```diff
@@ -749,45 +749,48 @@ func (v *Viper) find(key string) interface{} {
...
 	if exists && flag.HasChanged() {
-		jww.TRACE.Println(key, "found in override (via pflag):", flag.ValueString())
+		jww.TRACE.Printf("%q found in pflag override (%s): %s", key, flag.ValueType(), flag.ValueString())
 		switch flag.ValueType() {
 		case "int", "int8", "int16", "int32", "int64":
 			return cast.ToInt(flag.ValueString())
 		case "bool":
 			return cast.ToBool(flag.ValueString())
+		case "stringSlice":
+			s := strings.TrimPrefix(flag.ValueString(), "[")
```
shenwei356: This wrongly trims the "[" and "]" in the flags. 

Fix it by:

    s := flag.ValueString()
    return s[1:len(s)-1]

shenwei356: see https://github.com/spf13/pflag/pull/98

moorereason: @shenwei356, your solution is probably faster, but `strings.TrimPrefix` only removes a single "[" character. The problem with pflag was that it used `strings.Trim` which would remove multiples.

shenwei356: you are right!

PR #243, opened 2016-09-26
Remove expensive TRACE logging in find

### LINKED ISSUE #242: GetBool is surprisingly slow
See 438cc0d5c8791cdfc48adf605914f57c9bb0c29e

```
BenchmarkGetBool-4          	  100000	     12952 ns/op	     225 B/op	      10 allocs/op
BenchmarkGetBoolFromMap-4   	100000000	        10.8 ns/op	       0 B/op	       0 allocs/op
```

I know doing Viper lookups should be avoided in tight loops, but the above still surprises me -- not sure why 10 allocs is needed to do a simple bool lookup.

bep:
Also avoid doing a strings.Split in the Get common case.

Any TRACE statements in these hot paths must be totally turned off when not testing.

```
benchmark                     old ns/op     new ns/op     delta
BenchmarkGetBool-4            4090          409           -90.00%
BenchmarkGetBoolFromMap-4     6.33          6.28          -0.79%

benchmark                     old allocs     new allocs     delta
BenchmarkGetBool-4            6              3              -50.00%
BenchmarkGetBoolFromMap-4     0              0              +0.00%

benchmark                     old bytes     new bytes     delta
BenchmarkGetBool-4            129           33            -74.42%
BenchmarkGetBoolFromMap-4     0             0             +0.00%
```

Fixes #242

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,4 @@ _testmain.go
 
 *.exe
 *.test
+*.bench
--- a/viper.go
+++ b/viper.go
@@ -463,12 +463,11 @@ func GetViper() *Viper {
 // Get returns an interface. For a specific value use one of the Get____ methods.
 func Get(key string) interface{} { return v.Get(key) }
 func (v *Viper) Get(key string) interface{} {
-	path := strings.Split(key, v.keyDelim)
-
 	lcaseKey := strings.ToLower(key)
 	val := v.find(lcaseKey)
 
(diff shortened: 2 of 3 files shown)


moorereason:
Is there a way to make the trace logging calls a noop instead of removing them all?

bep:
>Is there a way to make the trace logging calls a noop instead of removing them all?

Sure.

PR #247, opened 2016-10-04
add link to Logrus and Viper integration Helper

heralight:
A Helper arround Logrus to wrap with spf13/Viper to load configuration with fangs!

And to simplify Logrus configuration use some behavior of Logrus_mate

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,10 @@ Many Go projects are built using Viper including:
 * [BloomApi](https://www.bloomapi.com/)
 * [doctl](https://github.com/digitalocean/doctl)
 
+Some use cases:
+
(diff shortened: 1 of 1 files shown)


awishformore:
Do we merge stuff like this? I wouldn't to be honest, it's not even a stand-alone package.

heralight:
Hi @awishformore what do you call " stand-alone package" ?
And to be honest, yes, it's a simple helper arround Viper and logrus, I never hide it.
But this simple wrapper can  help to understand howto use Viper, configure log and have a quick bootstrap.

awishformore:
I'm looking for opinions of other contributors / maintainers?

sagikazarmark:
To be honest, I don't really see this fit into the readme right now. I could imagine some sort of awesome-viper kind of list in the repo if there is demand for that.

Closing till then, feel free to open an issue or submit another PR with an alternative idea.

PR #255, opened 2016-10-12
Document case-insensitivity for Get, Sub, Set and SetDefault methods

bogem:


bogem:
Any offers or questions?

Issue #249, opened 2016-10-08
Latest Viper doubles Hugo build time


bep:
Candidate commit:
https://github.com/spf13/viper/commit/ec4eb2fa8549869ae7a2accd4fcc83d1c0555c15

We are in the process of removing viper.Get* invocations in Hugo hot paths, but I consider this to be a regression bug that should be fixed. 

We should also add some benchmarks for the most common functionality so we can spot this earlier.

See https://github.com/spf13/hugo/issues/2536

bep:
I have skimmed the commit and issues it solves. 

I think this could be solved by adding a fast path for keys without ".", so:

```
viper.GetBool("foo")
```

should follow a code path similar to before.

```
viper.GetBool("server.port")
```
needs more complicated logic.

@benoitmasson @spf13

bep:
Here is the current benchmark effect of that commit:

```
benchmark                     old ns/op     new ns/op     delta
BenchmarkGetBool-4            1037          421           -59.40%

benchmark                     old allocs     new allocs     delta
BenchmarkGetBool-4            6              3              -50.00%

benchmark                     old bytes     new bytes     delta
BenchmarkGetBool-4            113           33            -70.80%
```

That benchmark is as simple as it gets:

https://github.com/spf13/viper/blob/master/viper_test.go#L957

1037 ns is too much for a map lookup.

bep:
Here is the current benchmark effect of that commit:

```
benchcmp v1.bench v2.bench
benchmark                     old ns/op     new ns/op     delta
BenchmarkGetBool-4            469           1001          +113.43%

benchmark                     old allocs     new allocs     delta
BenchmarkGetBool-4            3              6              +100.00%

benchmark                     old bytes     new bytes     delta
BenchmarkGetBool-4            33            113           +242.42%
```

That benchmark is as simple as it gets:

https://github.com/spf13/viper/blob/master/viper_test.go#L957

10001 ns is too much for a map lookup.

benoitmasson:
Sorry about that... Though I must admit I didn't try to measure it before, I'm not surprised that the new code for `find()` slows down the `Get*()` process, due to the intrinsic complexity of “merging” deeply nested configuration layers.

On my computer, the performance loss is not that huge, it is around +25%:
```sh
# old
> go test -bench . -benchtime 10s -cpu 1,4 github.com/spf13/viper
BenchmarkGetBool            	 3000000	      4243 ns/op
[...]

PR #256, opened 2016-10-12
Increase performance of nested keys search

benoitmasson:
This follows PR #253, which closed Issue #249 induced by PR #195.

It generalizes to nested keys the improvement from PR #253 which consists in assuming that all keys are lower-cased, thus allowing to search for values by writing
```golang
val = m[key]
```
instead of
```golang
for k, v := range m {
  if strings.ToLower(k) == strings.ToLower(key) {
    val = v
  }
}
```
This induces a huge performance gain.

Function `insensitiviseMap()` was also fixed (it did not recurse correctly through the inner maps).

@bep my performance tests show no performance loss (and even very little gain) from your version, can I let you benchmark on your hardware to make sure I didn't remove anything essential (it should be fine, but let's double-check this time...)?

If that's OK for you, I think this would be a more generic solution than your patch, less likely to be bypassed in the future.

### DIFF
--- a/util.go
+++ b/util.go
@@ -41,15 +41,23 @@ func (pe ConfigParseError) Error() string {
 
 func insensitiviseMap(m map[string]interface{}) {
 	for key, val := range m {
+		switch val.(type) {
+		case map[interface{}]interface{}:
+			// nested map: cast and recursively insensitivise
+			val = cast.ToStringMap(val)
+			insensitiviseMap(val.(map[string]interface{}))
+		case map[string]interface{}:
+			// nested map: recursively insensitivise
+			insensitiviseMap(val.(map[string]interface{}))
+		}
+
 		lower := strings.ToLower(key)
 		if key != lower {
+			// remove old key (not lower-cased)
 			delete(m, key)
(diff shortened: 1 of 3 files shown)


bep:
This looks very good in general, but it is slightly slower...

```
▶ benchcmp /tmp/1.bench /tmp/2.bench
benchmark                     old ns/op     new ns/op     delta
BenchmarkGetBool-4            553           562           +1.63%
BenchmarkGet-4                452           477           +5.53%
BenchmarkGetBoolFromMap-4     7.00          6.51          -7.00%

benchmark                     old allocs     new allocs     delta
BenchmarkGetBool-4            4              5              +25.00%
BenchmarkGet-4                4              5              +25.00%
BenchmarkGetBoolFromMap-4     0              0              +0.00%

benchmark                     old bytes     new bytes     delta
BenchmarkGetBool-4            49            65            +32.65%
BenchmarkGet-4                48            64            +33.33%
BenchmarkGetBoolFromMap-4     0             0             +0.00%
```

So, if you reintroduce my `searchMapForKey` method the above be fine.

PR #262, opened 2016-10-17
expose mapstructure.DecoderConfig

rybit:
I wanted to have be able to easily port all the config structs that I had, they all use the `json:"xxx"` tag. 

In doing so I just grabbed the repeated code and put it together.


sagikazarmark:
`Unmarshal` accepts a decoder config hook now that allows you setting mapstructure options.

PR #265, opened 2016-10-24
Copy and insensitivise maps in Set

### LINKED ISSUE #261: Set(map) does not lowercase nested keys
Maybe it should, but it would probably be bad to modify the original map.

bep:
Fixes #261

### DIFF
--- a/util.go
+++ b/util.go
@@ -39,6 +39,39 @@ func (pe ConfigParseError) Error() string {
 	return fmt.Sprintf("While parsing config: %s", pe.err.Error())
 }
 
+// toCaseInsensitiveValue checks if the value is a  map;
+// if so, create a copy and lower-case the keys recursively.
+func toCaseInsensitiveValue(value interface{}) interface{} {
+	switch v := value.(type) {
+	case map[interface{}]interface{}:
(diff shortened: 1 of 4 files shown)


benoitmasson:
Sounds great indeed, I totally forgot about this case!

Just 2 minor remarks:

* don't forget to call it in `SetDefault()`, too. An maybe also in the function which updates the last map, `kvstore` (not sure about this one...)? The last map, `config`, is already dealt with in `unmarshallConfigReader()`.

* `cast.ToStringMap(v)` already performs a copy, so I guess the case `map[interface{}]interface{}` could simply be written as:
    ```golang
	value = insensitiviseMap(cast.ToStringMap(v))
```

Bonus remark: while looking for `insensitiviseMaps()` usages, I found that it is called in `Unmarshal()` and `UnmarshalExact()` (viper.go:727 and :764). These can probably be removed.

bep:
>cast.ToStringMap(v) already performs a copy

I thought about that, but it only makes a "shallow copy", i.e. the top level.  Thanks for the other tips, a sign that I really don't know my Viper.

bep:
I created https://github.com/spf13/viper/issues/266 for the `kvstore` case.

PR #267, opened 2016-10-25
Fixing error assignment

bengadbois:
err gets shadow declared and won't be able to be returned correctly later on


bengadbois:
Closing because this isn't going to work

PR #271, opened 2016-11-06
Add more intricate TypeByDefaultValue logic

nelsam:
The current documentation for `SetTypeByDefaultValue` suggests that I can set a default value and I will always be returned data for this key that is of that type.  The following code is similar to something I wrote based on that documentation before finding out that it actually ends up causing a panic:

```go
type Project struct {
    Name string
    Path string
    Gopath string
}

func init() {
    projects = viper.New()
    projects.SetTypeByDefaultValue(true)
    projects.SetDefault("projects", []Project{})
}

func Projects() []Project {
    return projects.Get("projects").([]Project)
}
```

This PR supports the above example, in addition to adding some tests for the prior code.  The solution is a bit naive, but I wanted to get feedback early on whether or not this is a feature that would be accepted.  If so, I think I'd need to at least fix up some of the panics in the code (specifically, anywhere `someVal.Set(convert(someOtherVal, typ))` is called - the current `convert` code falls back to returning the raw original value if it can't successfully convert, which will cause `someVal.Set` to panic).

A question for maintainers: what would be the correct thing to do when the data cannot be converted to the type of the default value?  I'm inclined to fall back to prior behavior (returning the raw data in `map[string]interface{}`, `[]interface{}`, `string`, or whatever format - just let the calling code complain if the data is wrong), but I think it's worth a discussion.

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -592,7 +592,6 @@ func (v *Viper) Get(key string) interface{} {
 
 	valType := val
 	if v.typeByDefValue {
-		// TODO(bep) this branch isn't covered by a single test.
 		path := strings.Split(lcaseKey, v.keyDelim)
 		defVal := v.searchMap(v.defaults, path)
 		if defVal != nil {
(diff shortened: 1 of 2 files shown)


nelsam:
Closing as it's been sitting here for ages.

PR #272, opened 2016-11-07
Add top level key support

mspaulding06:
Would be really nice to be able to get the top-level keys for a configuration so it is easier to iterate through a group of settings.

Here's an example.

Config file:

    ---
    my:
        settings:
            group:
                a:
                    one: thingone
                    two: thingtwo
                b:
                    one: anotherthingone
                    two: anotherthingtwo

Code:

    v := viper.Sub("my.settings.group")
    for _, i := range v.TopLevelKeys() {
        vv := v.Sub(i)
        for _, j := range vv.AllKeys() {
            // do something with this set of keys
        }
    }

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -30,6 +30,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/deckarep/golang-set"
 	"github.com/fsnotify/fsnotify"
 	"github.com/mitchellh/mapstructure"
 	"github.com/spf13/afero"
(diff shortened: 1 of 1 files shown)


mspaulding06:
I'm still interested in getting this merged.  Any chance of someone taking a look?

sagikazarmark:
Viper has an `AllKeys` method that exposes the list of keys from Viper. You could process that list to achieve the same goal.

I don't really see where it fits in Viper to be honest.

rafalkrupinski:
I'd like to have this merged too, I have the same use case as the PR's author

PR #277, opened 2016-11-21
Support multiple config change listeners (also fixes #175)

### LINKED ISSUE #175: Calling OnConfigChange() is mandatory, contradicting the README
A program using `WatchConfig()` **crashes** on config change if the `viper.onConfigChange` field is nil
(not set) - that is, **if `viper.OnConfigChange()` was not called**.

This contradicts the README, which states it is **not** required to call this function: 

> Viper supports the ability to have your application live read a config file while running.
> 
> Gone are the days of needing to restart a server to have a config take effect, viper powered applications can read an update to a config file while running and not miss a beat.
> 
> Simply tell the viper instance to watchConfig. Optionally you can provide a function for Viper to run each time a change occurs.
> 
> **Optionally** you can provide a function for Viper to run each time a change occurs.

Test program:

``` go
package main

import "github.com/spf13/viper"
import "fmt"
import "time"

func main() {
    viper.SetConfigName("config")
    viper.AddConfigPath(".")
    viper.SetConfigType("toml")
    err := viper.ReadInConfig()
    if err != nil {
        panic(fmt.Errorf("Fatal error config file: %s \n", err))
    }

    viper.WatchConfig()

    // If config.toml is edited while main() is sleeping, the program
    // crashes.

    // Uncommenting these three lines fixes the problem

    //viper.OnConfigChange(func(e fsnotify.Event) {
    //  fmt.Println("Config file changed:", e.Name)
    //})

    time.Sleep(8 * time.Second)

    fmt.Printf("key = %s", viper.GetString("key"))
}
```

```
❯ ./main
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xb code=0x1 addr=0x0 pc=0x69701]

goroutine 4 [running]:
panic(0x2b0d00, 0xc82000a100)
    /usr/local/Cellar/go/1.6/libexec/src/runtime/panic.go:464 +0x3e6
github.com/spf13/viper.(*Viper).WatchConfig.func1.1(0xc820108120, 0xc820086af0, 0x45, 0xc8201000e0)
    /Users/yberreby/code/go/src/github.com/spf13/viper/viper.go:256 +0x351
created by github.com/spf13/viper.(*Viper).WatchConfig.func1
    /Users/yberreby/code/go/src/github.com/spf13/viper/viper.go:263 +0x203
```

To reproduce the bug, **save `config.toml`** while the test program is sleeping.

remohammadi:
So we can pass the viper config to our packages, and those package which needs action on configuration change can register themselves without cluttering the main function.

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -156,7 +156,7 @@ type Viper struct {
 	aliases        map[string]string
 	typeByDefValue bool
 
-	onConfigChange func(fsnotify.Event)
+	onConfigChangeList []func(fsnotify.Event)
 }
 
 // New returns an initialized Viper instance.
@@ -226,9 +226,15 @@ var SupportedExts = []string{"json", "toml", "yaml", "yml", "properties", "props
 // SupportedRemoteProviders are universally supported remote providers.
 var SupportedRemoteProviders = []string{"etcd", "consul"}
 
+// OnConfigChange adds a change listener to the list of current listeners
 func OnConfigChange(run func(in fsnotify.Event)) { v.OnConfigChange(run) }
+
+// OnConfigChange adds a change listener to the list of current listeners
 func (v *Viper) OnConfigChange(run func(in fsnotify.Event)) {
-	v.onConfigChange = run
+	if v.onConfigChangeList == nil {
+		v.onConfigChangeList = make([]func(fsnotify.Event), 0, 1)
+	}
+	v.onConfigChangeList = append(v.onConfigChangeList, run)
 }
 
 func WatchConfig() { v.WatchConfig() }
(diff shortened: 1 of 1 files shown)


illyax:
will be merged any time soon?

sagikazarmark:
I don't think this is going to land. We are slightly pushing back on changes that can easily be worked around. In this case, you could write a few lines to add support for this:

```go
type OnConfigChangeFuncs []func(fsnotify.Event)

func (fns OnConfigChangeFuncs) OnConfigChange(event fsnotify.Event) {
	for _, fn := range fns {
		fn(event)
	}
}
```

or a different variation:

```go
func OnConfigChangeFuncs(fns ...func(fsnotify.Event)) func(fsnotify.Event) {
	return func (event fsnotify.Event) {
		for _, fn := range fns {
			fn(event)
		}
	}
}
```

PR #283, opened 2016-12-02
Fixed environment variable handling on Windows (fixes #282)

### LINKED ISSUE #282: AddConfigPath() is not portable
I faced a `panic` today when using `AddConfigPath` with an environment variable.

Basically, I did:

    viper.AddConfigPath("$HOME/.myprogram")
    viper.AddConfigPath("$MYPROGRAM_CONFIG_PATH/.myconfig")

This works fine on UNIX but not on Windows because [the code that looks for the end of the environment variable](https://github.com/spf13/viper/blob/master/util.go#L105) searches for `os.PathSeparator`. Obviously, I can just change the `AddConfigPath` declaration to:

    viper.AddConfigPath("$MYPROGRAM_CONFIG_PATH\\.myconfig")

But it then wouldn't work on UNIX anymore.

I feel like the environment variable parsing code is incorrect: it should not rely on the presence of a `/` (or `\\`) but on the first non valid character for a an environment variable name or the end of the string, whichever comes first.

Before I make a PR for this, what do you think ?

ereOn:


moorereason:
You can remove `userHomeDir()` since I don't think it would be used anymore.  I'm assuming you've tested and confirmed that this PR works as expected in Windows, right?

PS - I'm not a viper Contributor--I'm just a user.

bep:
It looks fine, but it would be really nice with a test or two.

moorereason:
It would be really nice to have AppVeyor.

util.go
```diff
@@ -94,16 +94,33 @@ func insensitiviseMap(m map[string]interface{}) {
...
 
-	if strings.HasPrefix(inPath, "$HOME") {
-		inPath = userHomeDir() + inPath[5:]
-	}
+	inPath = filepath.FromSlash(inPath)
 
-	if strings.HasPrefix(inPath, "$") {
-		end := strings.Index(inPath, string(os.PathSeparator))
-		inPath = os.Getenv(inPath[1:end]) + inPath[end:]
+	if strings.ContainsRune(inPath, '$') {
```
theherk: This will fail builds for the same reason mentioned [here](https://github.com/spf13/viper/pull/287#issuecomment-267423883). The solution is simple, as mentioned [here](https://github.com/spf13/viper/pull/287#issuecomment-267463231).

ereOn: @theherk Done. Sorry for the delay and thanks for the suggestion.

ereOn:
I also fixed the tests on Windows that were incorrectly relying on `path` instead of `path/filepath` resulting in inconsistent failures.

I had the remove the test directory `c\c` on Windows because it is a forbidden name.

All the tests pass now natively on Windows ! You may want to add an AppVeyor integration.

PR #285, opened 2016-12-06
Fix MergeInConfig error return

d33d33:
* **Fixed**: UnsupportedConfigError is returned if config file is not found
Swap getConfigFile and getConfigType call
Add a unit test


d33d33:
Up

d33d33:
Thanks!

PR #286, opened 2016-12-09
[WIP] Add write methods

theherk:
This PR is a work in progress. It is based on earlier code submitted by @g3rk6 in spf13/viper#153. I am working through satisfying the requests in @spf13's comments.

It adds methods to write the configuration back to the file or to a new file. There are methods to do so only if the file does not exist.

The two things stopping this from being done are:
- I cannot determine how to encode the map to hcl. I'm sure there is a way, but I have not worked it out yet. I think it will have to be printed to a writer. It would be nice if there was a map to string method. If anybody has a good idea, let me know.
- I have not written any tests yet. That may take a few.

After this is done, I also intend to add methods to write specific values in the map to file. In that case, you don't have to worry about adding unwanted values like env variables to the file.

### DIFF
--- a/util.go
+++ b/util.go
@@ -21,12 +21,14 @@ import (
 	"strings"
 	"unicode"
 
+	yaml "gopkg.in/yaml.v2"
+
 	"github.com/hashicorp/hcl"
 	"github.com/magiconair/properties"
 	toml "github.com/pelletier/go-toml"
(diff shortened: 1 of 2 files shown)


theherk:
Pull requests opened with wrong committer. Will reopen.

PR #296, opened 2017-01-04
Properly handle string slice values

### LINKED ISSUE #200: String Slices from pflags are not handled correctly by viper.

orian:
Added a code handling a pflag.StringSlice values. Also tested most cases.

It fixes #200.

Similar to: https://github.com/spf13/viper/pull/244

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -21,6 +21,7 @@ package viper
 
 import (
 	"bytes"
+	"encoding/csv"
 	"fmt"
 	"io"
 	"log"
@@ -880,7 +881,9 @@ func (v *Viper) find(lcaseKey string) interface{} {
 			return cast.ToBool(flag.ValueString())
 		case "stringSlice":
 			s := strings.TrimPrefix(flag.ValueString(), "[")
-			return strings.Split(strings.TrimSuffix(s, "]"), ",")
+			s = strings.TrimSuffix(s, "]")
+			res, _ := readAsCSV(s)
+			return res
 		default:
 			return flag.ValueString()
 		}
@@ -947,7 +950,9 @@ func (v *Viper) find(lcaseKey string) interface{} {
 			return cast.ToBool(flag.ValueString())
 		case "stringSlice":
 			s := strings.TrimPrefix(flag.ValueString(), "[")
-			return strings.Split(strings.TrimSuffix(s, "]"), ",")
+			s = strings.TrimSuffix(s, "]")
+			res, _ := readAsCSV(s)
+			return res
 		default:
 			return flag.ValueString()
 		}
@@ -955,6 +960,15 @@ func (v *Viper) find(lcaseKey string) interface{} {
 	// last item, no need to check shadowing
 
 	return nil
+}
+
+func readAsCSV(val string) ([]string, error) {
+	if val == "" {
+		return []string{}, nil
+	}
+	stringReader := strings.NewReader(val)
+	csvReader := csv.NewReader(stringReader)
+	return csvReader.Read()
 }
 
 // IsSet checks to see if the key has been set in any of the data locations.
--- a/viper.go
+++ b/viper.go
@@ -880,7 +880,7 @@ func (v *Viper) find(lcaseKey string) interface{} {
 			return cast.ToBool(flag.ValueString())
 		case "stringSlice":
 			s := strings.TrimPrefix(flag.ValueString(), "[")
-			return strings.TrimSuffix(s, "]")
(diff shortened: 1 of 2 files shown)


nullbio:
Nice, I was just coming to see if any progress had been made on this yet. Hoping to see this merged soon :-)

viper.go
```diff
@@ -880,7 +881,9 @@ func (v *Viper) find(lcaseKey string) interface{} {
 			return cast.ToBool(flag.ValueString())
 		case "stringSlice":
 			s := strings.TrimPrefix(flag.ValueString(), "[")
-			return strings.TrimSuffix(s, "]")
+			s = strings.TrimSuffix(s, "]")
+			res, _ := readAsCSV(s)
```
orian: Thinking out loud: maybe the error should not be silently ignored but shouted loudly as `panic`? Any thoughts?

bep: We silently swallow errors in the other cases, so ...

orian: Ok, let's be consistent ;)

orian:
@bep what about the merge? ;-)

viper_test.go
```diff
@@ -538,6 +538,44 @@ func TestBindPFlags(t *testing.T) {
...
 
+func TestBindPFlagsStringSlice(t *testing.T) {
+	for _, testValue := range []struct {
+		Expected []string
+		Value    string
+	}{
+		{[]string{}, ""},
+		{[]string{"jeden"}, "jeden"},
+		{[]string{"dwa", "trzy"}, "dwa,trzy"},
+		{[]string{"cztery", "piec , szesc"}, "cztery,\"piec , szesc\""}} {
```
Willyham: Use backticks for readability here?

Willyham:
Would love to see this be merged soon.

theckman:
@bep Hey there, just wanted to ping you again to see if you'd be interested in merging this! I'm hitting this issue in a new piece of software I'm using, and would like to use `viper` to handle the PFlag properly.

PR #299, opened 2017-01-13
Add check to see if v.onConfigChange is nil before calling

defond0:
Noticed #175, and thought it would be a relatively quick fix. Just added a check to see if we had a nil pointer at v.onConfigChange inside of WatchConfig. This fixes up the behavior described in #175, lemme know if there is anything else you would like done/ if I missed something.


txomon:
Hey, I was just going to submit this patch. Would it be possible to merge it? It's been here for months...

PR #303, opened 2017-01-21
Fix UnmarshalKey handling of time.Duration

mgilbir:
The use case arose when using UnmarshalKey on structs that contain nested structs that contain a time.Duration it would throw a message like:

`* '[0].Nested.Delay' expected type 'time.Duration', got unconvertible type 'string'`

This PR fixes the handling of time.Duration objects and adds a test for it.


favadi:
Anything needs to be done to have this PR merged?

Issue #312, opened 2017-02-21
Memory Leak Remote-Watch Consul?


wofr:
I'm using Viper to watch some values on Consul (like it is described in your sample see blow)

```
// open a goroutine to watch remote changes forever`
go func(){
    for {
        time.Sleep(time.Second * 5) // delay after each request

        // currently, only tested with etcd support
        err := runtime_viper.WatchRemoteConfig()
        if err != nil {
            log.Errorf("unable to read remote config: %v", err)
            continue
        }

       // unmarshal new config into our runtime config struct. you can also use channel
        // to implement a signal to notify the system of the changes
        runtime_viper.Unmarshal(&runtime_conf)
    }
}()
```

While doing this I observed that amount of go-function increases very fast.  The problem seems to be in the _Watch_ function in the _RemoteProvider_ (see below).  cm.Watch(rp.Path(), nil) creates a go function but without providing (and calling) the cancel channel. Thats why the created go function never stops! 
Thats why with every `runtime_viper.WatchRemoteConfig()` call a new function is created which is not stopped

```
func (rc remoteConfigProvider) Watch(rp viper.RemoteProvider) (io.Reader, error) {
	cm, err := getConfigManager(rp)
	if err != nil {
	    return nil, err
	}`
	resp := <-cm.Watch(rp.Path(), nil)
	err = resp.Error
	if err != nil {
		return nil, err
	}

	return bytes.NewReader(resp.Value), nil
}
```

Regards

Wolfgang

wofr:
Moreover the used lib to access consul (https://github.com/xordataexchange/crypt) is used in a way that also with every call a new go-function is called which never quits, due to the quite-channel is not even observed there.

wofr:
I guess the issue could also be solved by calling runtime_viper.WatchRemoteConfig only once outside the for loop in the go function. 
This should do the trick, but I need to verify this first.

```
go func(){
   err := runtime_viper.WatchRemoteConfig()
   if err != nil {
         log.Errorf("unable to read remote config: %v", err)
         return
    }
    for {
        time.Sleep(time.Second * 5) // delay after each request
        // currently, only tested with etcd support
     
        // unmarshal new config into our runtime config struct. you can also use channel
        // to implement a signal to notify the system of the changes
        runtime_viper.Unmarshal(&runtime_conf)
    }
}()
```

wofr:
Locally I've fixed the issue now.
[...]

PR #313, opened 2017-02-22
Avoid the start of go-routines which are never get stopped (Description Issue #312)

wofr:
This fix would avoid that while using the RemoteWatch-feature go-routines are started which never get stopped what causing a mem-leak (described here)

### DIFF
--- a/remote/remote.go
+++ b/remote/remote.go
@@ -40,6 +40,29 @@ func (rc remoteConfigProvider) Watch(rp viper.RemoteProvider) (io.Reader, error)
 
 	return bytes.NewReader(resp), nil
 }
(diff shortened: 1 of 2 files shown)


Boxx:
Experiencing the same issue with consul and `crypt`. A new goroutine is created for every poll of consul and is never closed.

bep:
@wofr I had to revert this PR and I don't think it is an option to reopen it in GitHub.

It would be good if you could create a new PR with these changes and one important change:

Don't pull dependencies from /remote into the /viper package (crypt).

wofr:
Thats fine. I will remove the import and create another PR

PR #321, opened 2017-03-15
Avoid the start of go-routines which are never get stopped (Description Issue #312; ) Follow Up on Pull-Request #313

wofr:
Removed dependency to crypto from viper.go

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -36,10 +36,14 @@ import (
 	"github.com/spf13/cast"
 	jww "github.com/spf13/jwalterweatherman"
 	"github.com/spf13/pflag"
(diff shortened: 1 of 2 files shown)


viper.go
```diff
@@ -40,13 +40,19 @@ import (
 
 var v *Viper
 
+type Response struct {
```
bep: I think we should rename this to `RemoteResponse`; which is in line with `RemoteProvider`, and it prevents people to confuse it with something they should care about. Fix that and I will merge.

wofr: Good point! Update is on the way

wofr:
changed Response to RemoteResponse

PR #329, opened 2017-04-07
Improve documentation for flags

bradleypeabody:
The existing example code for using flags does not include a complete usable snippet that you can paste into an existing project and easily see how to adapt it to use Viper.   This is a relatively minor change, but merging it in would make it simpler for the next person who comes along (like I have, multiple times) looking for a simple snippet to drop in that clearly shows how to integrate existing flags in your project with Viper.

### DIFF
--- a/README.md
+++ b/README.md
@@ -236,13 +236,26 @@ Like `BindEnv`, the value is not set when the binding method is called, but when
 it is accessed. This means you can bind as early as you want, even in an
 `init()` function.
 
-The `BindPFlag()` method provides this functionality.
+For individual flags, the `BindPFlag()` method provides this functionality.
 
 Example:
 
 ```go
 serverCmd.Flags().Int("port", 1138, "Port to run Application server on")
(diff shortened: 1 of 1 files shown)


bradleypeabody:
Hm, I guess nvm on this.  I'm realizing now that by calling pflag.Parse() instead of flag.Parse() it changes all the command line args from "-whatever" to "--whatever": that's a pretty significant change.  There doesn't appear to be a simple way to continue to use the default stdlib flag package with viper, which is unfortunate.

anthonyfok:
Thank you for your contribution, @bradleypeabody!

> Hm, I guess nvm on this. I'm realizing now that by calling pflag.Parse() instead of flag.Parse() it changes all the command line args from `-whatever` to `--whatever`: that's a pretty significant change. There doesn't appear to be a simple way to continue to use the default stdlib flag package with viper, which is unfortunate.

I am sorry, I am not qualified to answer that question, but let's ask @spf13 and see what he thinks about it.  (Hope he can see this comment in a now closed pull request.)

/cc @spf13

PR #330, opened 2017-04-10
adds test case for UnmarshalKey time.Duration

favadi:
I created this PR to prove that UnmarshalKey doesn't work with time.Duration type. I'm investigating why.


favadi:
I found out that there is a PR already: 303, closing.

PR #331, opened 2017-04-11
IsSet() returns false when flags are not set

### LINKED ISSUE #276: Binding with cobra flag results on IsSet always true
When I bind a key to a cobra flag, when I ask IsSet I always get true, even if the flag was not set in the command line flags as well.

Do I miss something or this is a real issue?

### LINKED ISSUE #580: IsSet always returns true if the key is bound to a flag.
It is not possible to use spf13/pflag without specifying a default value, and it is not a pointer, so you cannot set it to nil.  pflag does have a method `Changed()` which can test whether something was set with the flag, as well as unexported functions to test if the value is currently the zero value.

However, if you use viper.BindPFlag, when you call IsSet, the unexported find() method will return the value of the pflag as the very last condition, and it does not check whether it is the zero value, so it will always return true.

benoitmasson:
Added flexibility to `find()` function with an extra boolean argument, so that it can be used both by `Get()` (to retrieve a value, including flag's default) and `IsSet()` (whether the key is set, ignoring flag's default).

Fixes #276 
Fixes #580

(+ added some tests to enforce this behavior)

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -670,7 +670,7 @@ func GetViper() *Viper {
 func Get(key string) interface{} { return v.Get(key) }
 func (v *Viper) Get(key string) interface{} {
 	lcaseKey := strings.ToLower(key)
-	val := v.find(lcaseKey)
+	val := v.find(lcaseKey, true)
 	if val == nil {
 		return nil
 	}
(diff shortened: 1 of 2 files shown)


benoitmasson:
Just rebased on `master`, maybe this could be merged sometime?

This solves a small, but real issue.

sean-:
@benoitmasson Do you have an updated version of this PR that applies cleanly to `master`?

benoitmasson:
@sean- Not sure I get your point… `IsSet` is used to tell whether a config key is defined anywhere (other than flag default, which was not the case before this PR), so comparing the value to `defaultZeroValue()` does not make sense…

`nil` is the default zero value of a `map[string]interface{}` value, which is why values are matched against it to check for existence.

However I may have misunderstood your remark, so feel free to send a test which fails and I will try to fix it.

berniedurfee-ge:
Is this waiting on something specific?

tsuna:
Can we get this merged please?

hacdias:
@bep @spf13 what's the state of this PR? This would actually be really useful.

hairyhenderson:
There's a merge conflict, but otherwise this seems ready to merge. Would be awesome to see some movement here!

benoitmasson:
Just rebased, it all looks good.

Things have changed since my last update, but this PR is quite light and merging should not be an issue.

sagikazarmark:
Thanks for the PR and the ping. I'll review this PR shortly.

sagikazarmark:
Also fixes #580

PR #332, opened 2017-04-13
Add GetBoolSlice, GetDurationSlice

rvolosatovs:
Requires spf13/cast#57


bep:
I have merged the cast PR which fits nicely into that library, but I'm not sure we should add a convenience func for every cast operation available. 

User can easily do:

```
 cast.ToDurationSlice(v.Get(key))
```

Without Viper having to care about it.

rvolosatovs:
agreed, closing this

PR #333, opened 2017-04-17
Tests for flags handling improved

benoitmasson:
Highlighted and fixed a small bug in `TestBindPFlagsStringSlice()` (introduced in recently merged PR #296).

~While I was at it, moved all the flag-related tests from `viper_test.go` to the existing `flags_test.go` for consistency.~

See individual commits for details, they should be clear enough.

### DIFF
--- a/viper_test.go
+++ b/viper_test.go
@@ -557,11 +557,9 @@ func TestBindPFlagsStringSlice(t *testing.T) {
 
 			flagSet := pflag.NewFlagSet("test", pflag.ContinueOnError)
 			flagSet.StringSlice("stringslice", testValue.Expected, "test")
-			flagSet.Visit(func(f *pflag.Flag) {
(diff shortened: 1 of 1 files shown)


bep:
Can you:

1) Clean up your commit messages. ("Fix: TestBindPFlagsStringSlice() works", i.e. remove the prefix and create a sentence that makes it possible to understand what the commit is about)
2) Remove the move"all flag-related ..." commit. I'm not a big fan of PRs that "does this and then something remotely related".

benoitmasson:
Done.

The description of the commits followed the one-line statement, but I understand you're going to squash all of them and keep only the first line. I updated everything accordingly and will submit a separate PR for the contents move (unless you tell me it's not desired).

sagikazarmark:
Would be nice to get this merged, so that the tests are not breaking.

PR #342, opened 2017-04-29
Feature/unmarshal with meta

### LINKED ISSUE #341: No access to mapstructure decode metadata when using the unmarshal methods
I believe that methods should be added, `UnmarshalWithMeta` and `UnmarshalKeyWithMeta`. Those, methods would take an additional parameter, `mapstructure.Metadata`, and utilize a modified version of the `DecoderConfig` object that sets the `Metadata` field to a pointer to this value.

This way a user could get the metadata when decoding. The reason I suggest this, is because currently there is no good way to determine if values are decoded as their zero value or just defaulting to the initialized zero.

Let me know if you have questions or what you think about the proposal. I will be opening a PR later tonight or this weekend.

theherk:
Added parallel functions/methods to unmarshal data and return `mapstructure.Metadata`. This was not added for `UnmarshalExact`, since all keys should be used and none left unused.

Add a tests and examples for each new method.

Fixes #341

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -53,7 +53,7 @@ func init() {
 type remoteConfigFactory interface {
 	Get(rp RemoteProvider) (io.Reader, error)
 	Watch(rp RemoteProvider) (io.Reader, error)
-	WatchChannel(rp RemoteProvider)(<-chan *RemoteResponse, chan bool)
+	WatchChannel(rp RemoteProvider) (<-chan *RemoteResponse, chan bool)
 }
 
 // RemoteConfig is optional, see the remote package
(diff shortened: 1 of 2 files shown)


theherk:
I ended up implementing it slightly different than I suggested. To add an additional parameter in which we could put the meta data not only requires more work, but the library user wanting to use this would need to import mapstructure which seemed to unnecessarily obscure the simplicity.

The much better implementation that arose was just to have an additional return value for the meta data.

I chose not to simply add it to the previously existing unmarshal functions because, although it would be nice to simply ignore a return value, it would be a breaking change.

I hope this isn't over the top; I certainly don't want to bloat up a really elegant tool like Viper. However, I was finding that I needed to know in some cases if a value was simply the initialized zero value or set to the zero value in the configuration. To take it a step further, let me give an example.

Say a program has a configurable option to enable or disable each of a list of features. By default in the code, you set most of them to true and some to false. You also allow the user of the program to set these values in the config file. If your decision is made by accepting what is set in the code if no value provided in the config file or the value in the config file if given, there aren't any simple approaches (save this implementation).

You decisions may look like:

code  | override  | result
------|-----------|-----------
false | None      | false (default)
true  | None      | true
false | true      | true
true  | true      | true
false | false     | false
true  | false     | false

I hope it is clear how the addition of the metadata can help Viper users. It seems to be perfectly suited for cases such as these.

theherk:
I have sync'd the PR to resolve conflicts. Additionally, if anybody needs to use this feature, I am maintaining a fork with it included. See the PR description for details.

PR #344, opened 2017-05-02
add method for GetStringMapBool

chaseadamsio:
`GetStringMapBool` to allow for the example:

```
[disableContentType]
org = true
```

This makes for a clean way for checking if a key is set to a bool, rather than having to iterate over a slice of values to determine if that value is there.

This PR is a require for Hugo spf13/hugo#3424.


bep:
We're not adding convenience getters for every variant.

Use

```
cast.ToStringMapBool(v.Get(key))
```

Directly.

PR #345, opened 2017-05-08
Fix indentation in README.md

bogem:


bogem:
ping

anthonyfok:
Hi @bogem!

I saw you committing changes to Cobra directly, but not here in Viper, and I just realized that you are a Collaborator in Cobra, but only a Contributor in Viper, which is a bit odd, as in my mind Cobra and Viper are very much related to one another, and it is so good to see you actively invovled in both.

/cc @spf13  :-)

Anyhow, thanks for fixing the indentation in README.md!

bogem:
@anthonyfok Yes, it would be greatfor me to be a Viper collaborator too 😌

PR #357, opened 2017-06-05
If the extension is empty, do not append the dot when building the path

dmacvicar:
This allows to append "" to the viper.SupportedExts, while still set the config type, and have viper not append anything instead of appending just a dot.

This is needed if you need to parse a file where you know the format but that does not have an extension. In this case I found the bug by trying to read Salt's `/etc/salt/minion`.

```golang
minion.config = viper.New()
minion.config.SetConfigType("yaml")
minion.config.SetConfigName("minion")
minion.config.AddConfigPath("/etc/salt/")
viper.SupportedExts = append(viper.SupportedExts, "")
err := minion.config.ReadInConfig()
```


JimtotheB:
I wish this was merged. I had my hopes up for a `.<whatever>rc` file, but I'm stuck with a `.<whatever>rc.<ext>` unlike most of the other cli apps I use that read in a config.

dmacvicar:
I tried resolving the conflict via the web interface. I hope I got it right.

viper.go
```diff
@@ -1811,10 +1811,16 @@ func (v *Viper) getConfigFile() (string, error) {
 func (v *Viper) searchInPath(in string) (filename string) {
 	jww.DEBUG.Println("Searching for config in ", in)
 	for _, ext := range SupportedExts {
-		jww.DEBUG.Println("Checking for", filepath.Join(in, v.configName+"."+ext))
-		if b, _ := exists(v.fs, filepath.Join(in, v.configName+"."+ext)); b {
-			jww.DEBUG.Println("Found: ", filepath.Join(in, v.configName+"."+ext))
-			return filepath.Join(in, v.configName+"."+ext)
+		configFilePlusExt := ""
```
jaymickey: This is a personal preference, but nil value variable declarations make more sense as `var name type`. E.g. `var configPlusExt string`

dmacvicar: Done. Thanks.

dmacvicar:
https://github.com/spf13/viper/pull/722 is also an alternative

PR #360, opened 2017-06-12
Add glide file for dependency management

mspaulding06:
There are currently no vendor packages for this repository and I would like to propose that a glide file be added for simplifying dependency management.  Glide is an extremely useful tool and will only require users to `glide update` to pull in all required dependencies.

https://glide.sh/

### DIFF
--- a/glide.yaml
+++ b/glide.yaml
@@ -1,0 +1,19 @@
+package: github.com/spf13/viper
+import:
+- package: github.com/fsnotify/fsnotify
+- package: github.com/hashicorp/hcl
+- package: github.com/magiconair/properties
+- package: github.com/mitchellh/mapstructure
+- package: github.com/pelletier/go-toml
+- package: github.com/spf13/afero
+- package: github.com/spf13/cast
+- package: github.com/spf13/jwalterweatherman
+- package: github.com/spf13/pflag
+- package: github.com/xordataexchange/crypt
+  subpackages:
+  - config
+- package: gopkg.in/yaml.v2
+testImport:
+- package: github.com/stretchr/testify
+  subpackages:
+  - assert


bep:
It is a bad practice for a library to vendor dependencies. And if we would do that, I suspect the answer would to either 1) Wait for `deps` in Go 1.10 or 2) Use `govendor` like the other @spf13 project that actually uses vendoring, Hugo.

akutz:
Hi @bep,

FWIW, I disagree with regards to vendoring and libraries. libStorage vendors dozens of dependencies via Glide and the tool works fine to handle it. The `deps` tool will be nascent, unproven, and I, in addition to I suspect many others, won't be using it for a while. To that end, the `govendor` tool's claim that it _works with the standard vendor file_ is dubious as there is no such thing as a standard vendor file.

I vote to add the `glide.yaml` and `glide.lock` files as it causes no harm to do so and improves this project's friendliness towards the community.

bep:
>I vote to add the glide.yaml and glide.lock files as it causes no harm to do so and improves this project's friendliness towards the community.

It would be a new tool for me to learn, so if work is harm, then your claim is not true.

akutz:
Hi @bep,

Okay. Well, I'm an advanced user re: Glide and and routinely work with the Glide community. However, you've long since taken a rather acerbic approach to managing the Viper project, and you clearly do not wish to receive input from others, even those with administrative rights on this project such as myself. You closed the issue by declaring any work for you is _harm_? Plainly sir, that is ridiculous, rude, and unprofessional. 

Please remove me from this project as an administrator as there seems to be little point to me having any type of access beyond that of a normal user as you treat everyone equally poorly.

Thank you.

mattfarina:
@bep There's an important distinction to make between vendoring, that is storing the dependencies in the repo, and sharing dependency compatibility.

I completely agree that a library should not store its dependencies in the `vendor/` directory checked into VCS.

But, when an application pulls in a library it is useful for a dependency manager pulling that in to know the version compatibilities. It would be nice if things just worked and this was not a problem. Unfortunately, dealing with compatibility issues is a regular pain point for some of us. It happened again to me just last week.

I respect the goal of shooting for `dep` usage. I'm a supporter.

But, what to do before it's ready? `dep` is in many ways the child of `glide`. The same philosophy Glide has is what came to be in dep. It can be used for libraries (without vendoring in the VCS) and applications. These are close enough to each other that `dep` can now read glide files.

I'm not advocating for putting the dependencies within the VCS. I would just suggest having a `glide.yaml` file that broadcasts to glide users consuming this application what the compatibility is. And, it'll help get in the flow and be working toward `dep` when it's ready.

If you aren't clear on the use case please feel free to reach out and ask. I'm personally fine either way. I'd just like clear communication on the use cases and to remove any ambiguity.

bep:
@akutz You would have to ask @spf13 to be removed from his project(s). I have no access rights to do so.

bep:
And please open an **issue** about dependency management (if not already present) and take the discussion there before creating a PR with the conclusion and tool already chosen.

akutz:
Hi @bep,

This PR wasn't opened by me, and it was your initial response that indicated hesitance to a dependency manager.

bep:
@akutz my previous comment was not directed at you.

PR #370, opened 2017-07-22
Viper `Unmarshal` example fix

premshree:
The example for "Unmarshaling" in the README calls `Unmarshal` without a receiver. It might be clearer to indicate the receiver, either `viper` or a `viper_runtime`


sagikazarmark:
This should already be fixed

Issue #176, opened 2016-04-05
Improve usage instructions


ShanthkumarS079:
Tired installing viper with the below command
" **go get github.com/spf13/viper**"
It's not able to download all it's dependencies and it's child dependencies.

on99:
@spf13 dependency management is needed for viper.

bep:
@on99 no, a better README is needed for Viper.

sagikazarmark:
There is both an extensive readme and go mod support now. Please open a new issue or discussion if you have ideas for improvements.

PR #381, opened 2017-08-19
Feature/add glide dep mgmt

byron-janrain:
Should address #176.


bep:
This has been discussed before; Viper is a library and we're not adding some random dependency management tool here.

Issue #375, opened 2017-08-01
BindPFlags functionality does not seem to match documentation


therealfakemoot:
I've followed the README's example of how to bind flags into viper, but when I try to use the BindFlags() method versus explicitly binding flags by name, there's some sort of disconnect, the values never make it into viper: https://play.golang.org/p/7IsoizUFO8

The documentation ( https://godoc.org/github.com/spf13/viper#BindPFlags ) seems to indicate that this should be all that's necessary to allow config value lookups to also reference set flags. I do not discount the possibility that I'm misunderstanding but because the README only covers explicitly binding individual flag values I can't be sure.

My application will eventually have dozens upon dozens of commands, subcommands, and flag sets with common and uncommon flags between each. I'd *like* to be able to bind-once and not have to update my bindings any time I move, rename, or add flags.

juztin:
I too just ran across something similar:

This seem to have been working at some point for me  
_(did some refactoring and now I can't figure out why it's no longer working, if it ever was)_:

```
    cmd.Flags().String("address", "", "some address")
    viper.BindPFlag("address", cmd.Flags().Lookup("address"))
```

When looking up:

```
    fmt.Println("flag:", viper.GetString("address"))  // Returns empty
    fmt.Println(c.Flags().GetString("address"))        // Returns value set in 'address'
```

gadelkareem:
Any updates?

filinvadim:
Any updates?

pojntfx:
Thanks @juztin for your solution. Just to clarify; if one wants to use the `BindPFlags` functionality for multiple flags, they have to use _one pointer_ for all flags to bind.

i.e.:

```go
// constants.go (used by both commands)
var (
	serverHostPortFlag string
	configFileFlag     string
)
```

```
// cmd1.go
	applyCmd.PersistentFlags().StringVarP(&serverHostPortFlag, serverHostPortKey, "s", constants.DHCPDDHostPortDefault, constants.HostPortDocs)
```

```go
// cmd2.go
	deleteCmd.PersistentFlags().StringVarP(&serverHostPortFlag, serverHostPortKey, "s", constants.DHCPDDHostPortDefault, constants.HostPortDocs)
```

nick4fake:
This is a problem not only with PersistentFlags() but also with Flags(). Looks like viper only supports global flags

kalzoo:
Ran into this as well. This did not work as I had expected:

```go
err := viper.BindPFlags(cmd.Flags())
[...]

Issue #292, opened 2016-12-17
BindPFlag doesn't work


hiendv:
Hi, I'm using Viper with Cobra.
Firstly, this is my configuration file at `/home/hiendv/.app.json`.
```json
{
	"host": "0.0.0.0"
}
```

Secondly, this is my command
```go
var host string

// serveCmd represents the serve command
var serveCmd = &cobra.Command{
        //...
	Run: func(cmd *cobra.Command, args []string) {
		fmt.Println("The host is", host)
	},
}

func init() {
        RootCmd.AddCommand(serveCmd)
	viper.BindPFlag("host", serveCmd.Flags().Lookup("host"))
	serveCmd.Flags().StringVarP(&host, "host", "H", "", "Host to listen")
}
```

Lastly, the result is
```
Using config file: /home/hiendv/.app.json
The host is 
```
Expected `The host is 0.0.0.0`

Any ideas? Thanks a lot.

hiendv:
Without **BindPFlag**, **viper.GetString** works just fine
```go
func init() {
        RootCmd.AddCommand(serveCmd)
	serveCmd.Flags().StringVarP(&host, "host", "H", "", "Host to listen")
}
```
```go
fmt.Println("The host is", viper.GetString("host"))
// The host is 0.0.0.0
```

It won't when I use **BindPFlag**
```go
func init() {
        RootCmd.AddCommand(serveCmd)
        viper.BindPFlag("host", serveCmd.Flags().Lookup("host"))
	serveCmd.Flags().StringVarP(&host, "host", "H", "", "Host to listen")
}
```
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x50 pc=0x48be75]

goroutine 1 [running]:
panic(0x7f16a0, 0xc420012170)
	/usr/local/go/src/runtime/panic.go:500 +0x1a1
github.com/spf13/viper.pflagValue.HasChanged(0x0, 0xc4200dfc20)
	/data/go/src/github.com/spf13/viper/flags.go:41 +0x5
github.com/spf13/viper.(*Viper).find(0xc4200da0f0, 0x8588ec, 0x4, 0x4, 0x61)
	/data/go/src/github.com/spf13/viper/viper.go:875 +0x1021
github.com/spf13/viper.(*Viper).Get(0xc4200da0f0, 0x8588ec, 0x4, 0x475ea7, 0xa31c20)
[...]

Issue #189, opened 2016-05-18
Provide meaningful information when a key / config is wrong


kkirsche:
Currently I'm trying to debug this situation and I'm really struggling to find the issue due to the lack of meaningful error message. Specifically:

``` go
import (
    "fmt"

    "github.com/spf13/cobra"
    "github.com/spf13/viper"
)

// sendCmd represents the send command
var sendCmd = &cobra.Command{
    Use:   "send",
    Short: "",
    Long: ``,
    Run: func(cmd *cobra.Command, args []string) {

        fmt.Println("Cycling through keys")
        fmt.Println(viper.AllKeys())
        for _, key := range viper.AllKeys() {
            fmt.Println(viper.Get(key))
        }
    },
}

func init() {
    RootCmd.AddCommand(sendCmd)

    sendCmd.PersistentFlags().StringP("identifier", "i", "", "The identifier")
    viper.BindPFlag("send.identifier", sendCmd.Flags().Lookup("identifier"))

    sendCmd.PersistentFlags().StringP("frequency", "f", "30s", "The frequency")
    viper.BindPFlag("send.frequency", sendCmd.Flags().Lookup("frequency"))
}
```

When I run this though:

```
~/g/g/s/n/s/s/n/project ❯❯❯ sudo project send
Using config file: /Users/example/.project.json
Cycling through keys
[send.identifier send.frequency general.pidfile general send logging.verbosity logging.stdout]
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xb code=0x1 addr=0x50 pc=0xa62cc]

goroutine 1 [running]:
panic(0x645da0, 0xc8200100d0)
    /usr/local/Cellar/go/1.6.2/libexec/src/runtime/panic.go:481 +0x3ff
github.com/spf13/viper.pflagValue.HasChanged(0x0, 0xc820075500)
    /Users/example/git/go/src/github.com/spf13/viper/flags.go:41 +0x3c
github.com/spf13/viper.(*Viper).find(0xc8201360e0, 0x6f2980, 0xe, 0x0, 0x0)
    /Users/example/git/go/src/github.com/spf13/viper/viper.go:738 +0x19d
github.com/spf13/viper.(*Viper).Get(0xc8201360e0, 0x6f2980, 0xe, 0x0, 0x0)
    /Users/example/git/go/src/github.com/spf13/viper/viper.go:461 +0x12c
github.com/spf13/viper.Get(0x6f2980, 0xe, 0x0, 0x0)
[...]

PR #396, opened 2017-10-08
Improve common mistake when using Cobra flagsets

tbutts:
Users get tripped up when accidentally mix & matching
Cobra's `Flags()` and `PersistentFlags()`, where they will
call `Lookup(<flag>)` on the wrong flagset.

Following the README, users would call `Lookup` to fill in
the flag source parameter of Viper's `BindPFlag`, but by
using the wrong flagset, the parameter would simply be nil.
This would then result in a runtime panic whenever trying
to retrieve the flag's value with `Get(<flag>)`.

This work alleviates the mistake by:
+ Performing a nil check in `BindPFlag` and causing
  `BindFlagValue` to return an error if the parameter
  flag is nil, as it would be during the expected mixup.
+ Updating the godoc of `BindPFlag` and the README's example,
  to drill the fact that `Flags` and `PersistentFlags` are
  different flagsets.
+ Adding a test for this pattern of binding nil on accident
  and then retrieving flags. Before these changes, this test
  would panic. Now, the test instead gracefully errors and
  applies no changes to the viper instance.

-----

These changes should clarify usage of `BindPFlag`, which helps to solve issues such as #375, #292, and #189.

### DIFF
--- a/README.md
+++ b/README.md
@@ -246,6 +246,11 @@ serverCmd.Flags().Int("port", 1138, "Port to run Application server on")
 viper.BindPFlag("port", serverCmd.Flags().Lookup("port"))
 ```
 
+The above example would also work with
+[Cobra's Persistent Flags](https://github.com/spf13/cobra#persistent-flags),
+but careful not to mix and match Cobra `Flags` and `PersistentFlags` when
+binding with Viper.
+
 You can also bind an existing set of pflags (pflag.FlagSet):
 
 Example:
--- a/viper.go
+++ b/viper.go
@@ -799,9 +799,15 @@ func (v *Viper) BindPFlags(flags *pflag.FlagSet) error {
 //	 serverCmd.Flags().Int("port", 1138, "Port to run Application server on")
 //	 Viper.BindPFlag("port", serverCmd.Flags().Lookup("port"))
 //
+// NOTE: Be consistent with either Cobra's Flags() or PersistentFlags(),
+//       do not mix them.
 func BindPFlag(key string, flag *pflag.Flag) error { return v.BindPFlag(key, flag) }
 func (v *Viper) BindPFlag(key string, flag *pflag.Flag) error {
-	return v.BindFlagValue(key, pflagValue{flag})
+	var flagVal FlagValue
+	if flag != nil {
+		flagVal = pflagValue{flag}
+	}
+	return v.BindFlagValue(key, flagVal)
 }
 
 // BindFlagValues binds a full FlagValue set to the configuration, using each flag's long
--- a/viper_test.go
+++ b/viper_test.go
@@ -597,6 +597,15 @@ func TestBindPFlag(t *testing.T) {
 
 }
 
+func TestBindPFlagToNil(t *testing.T) {
+	v := New()
+	// nil arg often comes as an accident from calling something like:
+	//     `myCobraCmd.Flags().Lookup("missingFlag")`
+	err := v.BindPFlag("host", nil)
+	assert.Equal(t, nil, v.Get("host")) // `Get` here used to panic
+	assert.EqualError(t, err, `flag for "host" is nil`)
+}
+
 func TestBoundCaseSensitivity(t *testing.T) {
 	assert.Equal(t, "brown", Get("eyes"))
 


therealfakemoot:
Thanks for providing some illumination on the issue ( #375 ) I was having. I suspected user-error may have been at fault. I'm still having some issues understanding the correct usage even with your ( as-of-yet unmerged ) update to the README.

In short, I'm seeing "Do not mix PFlags and Flags", which seems reasonable enough. The problem I'm having with `BindPFlags`, however, is that it seems to expect PFlags. Cobra `Command` values only expose a `Flags()` method and I'm not sure how to reconcile this. How then can I avoid mixing PFlags and Flags? The example in the README is explicitly mixing the two, or seems to be:

```
serverCmd.Flags().Int("port", 1138, "Port to run Application server on")
viper.BindPFlag("port", serverCmd.Flags().Lookup("port"))
```

`port` is being defined as a `Flag`, not a `PersistentFlag` but the updated README advises against mixing the two?

Per my original issue, I'm able to bind individual flags correctly via `Lookup()` but my application will eventually have multiple dozens or hundreds of configuration parameters. I want a user to be able to override any one of them via the command line at any time and eventually, maintaining that long list of bindings will end up breaking something.

I've got a tag on repo [here](https://github.com/therealfakemoot/genesis/tree/BindPFlags-problems) with a live example of the codebase expressing this behavior. I'm worried this may be an order-of-operations thing ( am I calling BindPFlags before/after the viper loadConfig when it should be the other way around? )

Example run:
```
$ go run genesis.go --config=./genesis.yaml --verbose
INFO[0000] Config filepath                               config=
INFO[0000] Verbosity enabled                             verbose=false
INFO[0000] All settings                                  Settings="map[logdata:map[logfile:butts.log] mapdir:maps extdirs:ext verbose:true]"
```

Config file ( YAML )
```.genesis.yaml
verbose: True
LogData:
  LogFile: "butts.log"
```

tbutts:
I see your problem, @therealfakemoot, which is that you've mixed up the idea of `PFlags` the library, and cobra `Command.PersistentFlags()`, in the context of viper. It's definitely confusing and I have to double-take every time I revisit Viper/Cobra code.

First, I'll suggest bootstrapping viper in your code with this simple change to the example in the readme:
```go
// HERE: Save and reuse a Cobra Command's `Flags()` or `PersistentFlags()`, sharing the returned
// pflag `Flagset` collection between the Cobra Command and a Viper instance
server_flags := serverCmd.Flags()

server_flags.Int("port", 1138, "Port to run Application server on")  // Add a flag to Cobra Command
viper.BindPFlag("port", server_flags.Lookup("port"))  // And then to Viper (you could also use viper.BindPFlags(server_flags)
```

Let me try to break it down:
* [pflag](https://github.com/spf13/pflag) provides POSIX style flags. 
  * This goes beyond the std lib `flag` package, which doesn't let you easily do CLI options that have shorthand (single-dash)  and long (double-dash) options.
  * Many `pflag` can be bundled in `flagset` collections, just like the std lib package
* [cobra](https://github.com/spf13/cobra) then greatly extends `pflag`, allowing fancy CLI apps with nested subcommands.
  * Dash options can be specific to any level of the command (the Command's `Flags()`), or they can be shared from a top level to all subcommands (the Command's `PersistentFlags()`).  [See the Cobra readme](https://github.com/spf13/cobra#persistent-flags)
  * So you can have a `--verbose` flag only defined at the top level if it is set on the Command's `Flags()`, which is _a `flagset` of `pflag` defs_.
* [viper](https://github.com/spf13/viper) is a framework for app configuration.
  * It lets you easily parse a config file for options, while letting users override any config options with things like CLI flags, 
  * CLI flags are bound to a pflag, parsed by Cobra (whether it be Persistent or Command-specific), and then merged with Config File options set with Viper.
  * This relationship is what I tried to clear up with this tiny PR in the Viper readme

I know it all might seem befuddling, but I hope this helps you figure out what you need to do. Be consistent. Cheers.

therealfakemoot:
Thank you for responding so promptly! I'm going to take another crack at this and see if I can correct my misunderstandings and get things working. I did notice your example still explicitly calls lookup on a single flag, though. Is the functionality I'm seeking, specifically "free form" linking of CLI flags to config values, not actually possible?

To provide a more concrete example, the project is a procedural map generation project, so there's gonna a huge swath of parameters. I want the process to be interactive, repeatable, and modifiable. A user will generate a terrain map using parameters stored in a config file ( default values, or values put together from previous experimentation ), examine it in a visual renderer, and possibly say "I want taller mountains". So they rerun the terrain generation but with the `--mountain-tallness=3.0` flag ( with the config file value being 2.5 ).

I'll end up setting 'reasonable' defaults for all these parameters in the config file, allowing users to specify their own custom configuration files and then override some/all/none values during the iterative process.

I understand that Viper is the source-of-truth for the application, and I can get Viper reading from the config file no issue. I'm just having trouble getting Viper to serve values provided by CLI flags in preference to config file values, *without* doing a `Lookup()` call on each flag. If that's not possible, I'll eat it and deal with the maintenance overhead.

tbutts:
Now, I'm not a maintainer of viper, so I may be off base, but I think what you say is possible. Going by [viper's README (working with flags)](https://github.com/spf13/viper#working-with-flags), the example goes:
```go
pflag.Int("flagname", 1234, "help message for flagname")

pflag.Parse()
viper.BindPFlags(pflag.CommandLine)

i := viper.GetInt("flagname") // retrieve values from viper instead of pflag
``` 
(Don't get confused on the `pflag.Parse()` and `pflag.CommandLine` variable, that's just [a default set of flags](https://github.com/spf13/pflag/blob/master/flag.go#L1192), just as `YourCmd.Flags()` gives you).

I noticed you were trying to use this with `viper.BindPFlags(RootCmd.Flags())`. But, the mix up is that you added flags to your RootCmd in the `RootCmd.PersistentFlags()` set, but then bound viper to the `RootCmd.Flags()` set. You bound an empty flagset to your viper. I've made the same mistake. 

Fix that, and you should be able to define all flags you want on your Command `RootCmd`, add them to a Viper, and read them with identical names.
**E.G.** `viper.GetString("LogData.LogFile")` to get both the CLI arg `--LogData.Logfile="butts.log"`, or as the YAML config file option.

You can see that the CLI flags end up being more cumbersome, but it should be much easier to maintain.

therealfakemoot:
Thanks again for the quick feedback. I realized another glaring issue with my code. I was using `BindPFlags` and invoking `viper.Get*` in `init()`. At the time of my `Get*` calls, Cobra hasn't actually parsed any flags and naturally, the values are `nil`. I moved my `viper.Get*` invocations into the `Run` hook on my `RootCmd` and changed the `PersistentFlag`s to `Flag`s, and now viper is able to fetch those values. I'm going to have to think very carefully about how I structure my commands/subcommands, flags, and config file to make sure that the propagation hierarchy behaves sensibly and predictably, but I believe I've busted the first major roadblock.

I'm now using PreRun and PostRun, and finally added some logic to make the debug flag actually *mean* something, and I think I've got something approximating what I need.

I greatly appreciate your persistence in helping, I know there's a lot of stuff a man can do other than helping randos on the internet with their busted code. You'll get a mention in my commits and the blog post I write about the embarrassing incident.

sagikazarmark:
Nil check is already added. The documentation should be added to Cobra instead IMHO.

PR #398, opened 2017-10-17
Fix retrieval of pflag stringArray

databus23:
`pflag.StringArray` suffers from the same problems as `StringSlice` did described in this issue https://github.com/spf13/viper/issues/112

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -896,7 +896,7 @@ func (v *Viper) find(lcaseKey string) interface{} {
 			return cast.ToInt(flag.ValueString())
 		case "bool":
 			return cast.ToBool(flag.ValueString())
-		case "stringSlice":
+		case "stringSlice", "stringArray":
 			s := strings.TrimPrefix(flag.ValueString(), "[")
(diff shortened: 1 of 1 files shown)


carolynvs:
I have run into this same bug in cobra+viper and just tried out your patch. I'm still getting an empty string in the array, instead of an empty array when the flag was not specified.

```
([]string) (len=1 cap=1) {
 (string) ""
}
```

When I change my declaration of the flag from `cmd.Flags().StringArrayVarP` to `cmd.Flags().StringSliceVarP`, then I'm finally getting what I expected, an zero length slice/array.

```
([]string) {
}
```

I think fix this is close, because it is stripping out the extra square brackets, but it may need a bit more to remove the empty entry too.

denouche:
Hello, could we merge this bugfix?
Thanks!

nickgerace:
This would be extremely helpful for https://github.com/rancher/k3d/issues/506 to be merged. I'd be happy to help, if needed.

bukowa:
I believe some tests may push it forward

PR #399, opened 2017-10-17
Add SetKeyDelimiter

### LINKED ISSUE #170: RFE: Allow changing the Key Delimiter for accessing nested values
I am talking about this field of Viper struct:

```
// Delimiter that separates a list of keys
// used to access a nested value in one go
keyDelim string
```

This can be useful in the following circumstances:
1. The keys in the config file contain periods.
2. Config key is being generated by something that uses a different delimiter - can save an unnecessary strings.Replace().

Looking at the code, it should be a simple matter of adding a SetKeyDelim() method.

robertgzr:
Allow changing the Key Delimiter for accessing nested values.

Fixes #170

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -1560,6 +1560,13 @@ func (v *Viper) findConfigFile() (string, error) {
 	return "", ConfigFileNotFoundError{v.configName, fmt.Sprintf("%s", v.configPaths)}
 }
 
+func SetKeyDelimiter(in string) { v.SetKeyDelimiter(in) }
+func (v *Viper) SetKeyDelimiter(in string) {
+	if in != "" {
+		v.keyDelim = in
+	}
+}
+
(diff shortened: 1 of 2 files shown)


m4agenbitter:
I would really appreciate it if someone would be so kind to merge this.

GrantSeltzer:
This would be very useful because if I want to use configuration keys that are file names (i.e. `file.txt`) it currently breaks on any accessor method.

GrantSeltzer:
@spf13 @bep Any chance we could bother you to take a look at this?

Thoro:
Could we finally merge this, since it's approved and hanging here for 1.5 years?

sagikazarmark:
Fixed in #794

grantseltzer:
cc @indecisivedragon @jarvision :)

Issue #404, opened 2017-10-24
SetFs useless when not using default Viper instance


progrium:
When creating an instance of Viper (such as writing tests), you can use SetFs to work with, say, an in-memory filesystem. However, Viper will not use it when looking up a configuration file. `searchInPath` uses a `util.go` function called `exists` that is hardcoded to the global `v` instance. 

This can be fixed by simply letting `exists` take an instance of a filesystem to use.

PR #405, opened 2017-10-24
Allow exists util function to take afero.Fs

progrium:
This allows it to work with non-default instances of Viper as you would expect. `searchInPath` then passes it's own instance of `fs`.

Signed-off-by: Jeff Lindsay <progrium@gmail.com>


progrium:
closes https://github.com/spf13/viper/issues/404

bketelsen:
LGTM

PR #414, opened 2017-11-11
Declare required dependencies with `go dep` #413

### LINKED ISSUE #413: Declare required dependencies with `go dep`
The goal is to have a pair of `Gopkg.toml` and `Gopkg.lock` files to declare the dependencies in the project (but not commit the content of the `vendor` folder in the repo, as already discussed in https://github.com/spf13/viper/issues/223)

This would hopefully solve 2 cases:
- as a new contributor, I would just need to execute `dep ensure` to install all the project dependencies and `go build` or `go test` would run successfully
- as a consumer of the library, I would not have to explicitly included the project dependencies in my own project setup (dep, glide, etc.)

xcoulon:
Include the `Gopkg.toml` and `Gopkg.lock` files that
were produced by `dep init`.
Exclude the `vendor` folder from the git repo

Fixes #413

Signed-off-by: Xavier Coulon <xcoulon@redhat.com>


xcoulon:
closing as out of date since viper now uses go modules

PR #421, opened 2017-11-29
travis: update go versions

dvrkps:


dvrkps:
@spf13 @bep ping :) happy holidays.

PR #425, opened 2017-12-16
Document reading a specific config file in README

casept:
I think that specifying a particular config file using `SetConfigFile()` is a rather common use case, therefore it deserves mention in the README, which is the first (and perhaps only) documentation a user who's new to the library reads.

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,6 +103,17 @@ if err != nil { // Handle errors reading the config file
 }
 ```
 
+Alternatively (or in addition), the path to a specific config file,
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -103,6 +103,17 @@ if err != nil { // Handle errors reading the config file
 }
 ```
 
+Alternatively (or in addition), the path to a specific config file,
```
adamdecaf: I'd reword this to: 

> Alternatively (or in addition), the path to a specific config file can be provided (instead of a directory).

Issue #427, opened 2017-12-20
WriteConfig should not fail on unknown file extensions when the config type is set


mildwonkey:
This fails with an Unsupported Config Type error:
```
package main

import (
  "fmt"
  "github.com/spf13/viper"
)

func main(){
  example := viper.New()
  example.SetConfigFile("vars.tfvars")
  example.SetConfigType("hcl")
  example.Set("my_new_variable", "hi_mom")
  if err := example.WriteConfig(); err != nil {
    fmt.Println(err.Error())
  }
}
```

This works:
```
package main

import (
  "fmt"
  "github.com/spf13/viper"
)

func main(){
  example = viper.New()
  example.SetConfigFile("vars.hcl")
  example.SetConfigType("hcl")

  example.Set("my_new_variable", "hi_mom")
  if err := example.WriteConfig(); err != nil {
    fmt.Println(err.Error())
  }  
}
```

WriteConfig should first check if the ConfigType has already been set before looking at file extensions.

It's here-ish, as far as I can tell: https://github.com/spf13/viper/blob/1a0c4a370c3e8286b835467d2dfcdaf636c3538b/viper.go#L1238

vermaaakansha:
@mildwonkey  Has this issue got fixed?

Fuhrmann:
@mildwonkey have you found a workaround for this problem?

mildwonkey:
I worked around it by forking the project, writing the linked PR, and using the fork in my code. I do not recommend using my fork, it's many years out of date, but you could apply that change to your own fork of this repository.

PR #428, opened 2017-12-20
fix issue 427: use v.getConfigType to determine config type in WriteConfig

mildwonkey:
use the existing getConfigType function instead of figuring it out anew
#427


mildwonkey:
@spf13 @bep sorry to do this, but - ping! review pretty please

sgettys:
Is this ever going to make it into master?

traintraveler:
I'd like to know when this gets merged as well

Fuhrmann:
Could this be merged?

greg-szabo:
I think using `getConfigType` is not enough. You can build a configuration without defining a filename and then you can try to run `SafeWriteConfigAs` which again doesn't require to have a filename in your configuration (you are giving the filename as an input parameter to the function).

The problem is that `getConfigType` will check the filename to get the extension from it. So, in the above-mentioned case, in `SafeWriteConfigAs` the result of `getConfigType` will be an empty string: no filename was set in the object, there's no way to figure out it's file extension.

I've tried to figure out why `writeConfig` is dealing with file extensions and I think the reason is that there's the point where all filenames are in place and the file extension can be figured out. I think the real issue is that the file extension becomes an ultimate priority, even in cases where the developer explicitly set a specific config type using `SetConfigType`.

In #934 I'm proposign a solution that's making the config type the priority during write, so files with weird extensions (but a properly set ConfigType) can also be saved.

PR #429, opened 2017-12-22
Lookup environment variables instead of checking if the value is empty

### LINKED ISSUE #317: Allow EmptyStrings to read from Environmental Variables
The code below inside the find(key string) function in viper.go, makes it so that if an env var is set to "", the default value is used instead. I have a use case where I need the default to be overridden with empty string from the environment. Any chance this will get implemented in this way?

 envkey, exists := v.env[key]
  if exists {
    jww.TRACE.Println(key, "registered as env var", envkey)
    if val = v.getEnv(envkey); val != "" {
      jww.TRACE.Println(envkey, "found in environment with val:", val)
      return val

sagikazarmark:
Empty environment variables are perfectly valid values.
Fortunately Go provides a way to check whether an env var
is empty or not set.

Fixes #317

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -337,11 +337,11 @@ func (v *Viper) mergeWithEnvPrefix(in string) string {
 // getEnv is a wrapper around os.Getenv which replaces characters in the original
 // key. This allows env vars which have different keys than the config object
 // keys.
-func (v *Viper) getEnv(key string) string {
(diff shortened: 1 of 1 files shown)


sagikazarmark:
I accidentally deleted the source repository, resubmitted as #440

PR #439, opened 2018-01-16
Added functionality to mock the environment variable store, for tests

royra:
Mocking the filesystem is already possible using afero and the SetFs function.
This PR adds the option to mock the environment variable store as well, with the default implementation of OS environment variables.
I find it helpful when testing code that uses viper; the previous option of mutating the shared OS environment state interferes with running tests in parallel.

### DIFF
--- a/util.go
+++ b/util.go
@@ -94,16 +94,16 @@ func insensitiviseMap(m map[string]interface{}) {
 	}
 }
 
-func absPathify(inPath string) string {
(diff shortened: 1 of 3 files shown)


jippi:
This PR is pretty old - but would there be any chance for a fresh contribution of this kind of PR to be accepted? I have exactly this need as well

sagikazarmark:
Go 1.17 adds a setenv function that helps with testing.

PR #440, opened 2018-01-16
Lookup environment variables instead of checking if the value is empty

### LINKED ISSUE #317: Allow EmptyStrings to read from Environmental Variables

sagikazarmark:
Empty environment variables are perfectly valid values.
Fortunately Go provides a way to check whether an env var
is empty or not set.

Fixes #317 

Resubmitted #429

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,13 +179,14 @@ viper.GetBool("verbose") // true
 ### Working with Environment Variables
 
 Viper has full support for environment variables. This enables 12 factor
-applications out of the box. There are four methods that exist to aid working
+applications out of the box. There are five methods that exist to aid working
 with ENV:
 
  * `AutomaticEnv()`
  * `BindEnv(string...) : error`
  * `SetEnvPrefix(string)`
  * `SetEnvKeyReplacer(string...) *strings.Replacer`
+  * `AllowEmptyEnvVar(bool)`
 
 _When working with ENV variables, it’s important to recognize that Viper
 treats ENV variables as case sensitive._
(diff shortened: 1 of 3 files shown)


sagikazarmark:
@tsl0922 test cases provided in this PR

dcormier:
Sure would be nice to see this working. I'd like to be able to use empty environment variables.

sagikazarmark:
Can someone restart the build please? Tests passed for me locally.

tsl0922:
@sagikazarmark you can do a force push to trigger it.

sagikazarmark:
Ping Looks like tests are passing.

sagikazarmark:
Any news?

sagikazarmark:
Ping

b1v1r:
Any ETA on this being merged (or rejected)? Looks like @tsl0922 reviewed this as #429, but then it stalled when it was resubmitted.

sagikazarmark:
@bep @andrewstuart can you help us out here?

bep:
>So my main concern here is that this has the potential to be considered a breaking change if are using empty environment variables in place of unsetting.

On one side, I think the map lookups follow the `v, ok :=` pattern, which would be similar, but I agree -- we cannot change this as the default behavior. That this is OS env vars, something "invisible", will create some very subtle bugs around. But an option (default off) would be good.

sagikazarmark:
@andrewstuart @bep Thanks for your comments.

I added an `AllowEmptyEnv` method with tests and some documentation in the readme. By default this feature is turned off.

sagikazarmark:
Breaking tests seem to be unrelated

PR #441, opened 2018-01-16
Creation of Sub now copies parent properties and sets envPrefix

royra:
This PR fixes two issues with Sub():

1. Options were not copied from parent (e.g, `automaticEnvApplied`).

2. Environment variable names were wrong: `Sub("x").Get("y")` should look up env var `x.y`, but was instead looking up `y`

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -204,6 +204,13 @@ func New() *Viper {
 	return v
 }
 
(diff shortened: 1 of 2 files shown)


sagikazarmark:
This should be fixed in 1.16. Thanks for the PR though.

PR #449, opened 2018-01-29
Use 'viper' as decoder tag name

tylerb:
This PR changes the decoder tag name from "mapstructure" to "viper".

**NOTE** - this PR depends on https://github.com/mitchellh/mapstructure/pull/107 to prevent breaking changes. If merged before the `mapstructure` PR, any previous implementations of `viper` that use `mapstructure` as the tag name will break.

### DIFF
--- a/README.md
+++ b/README.md
@@ -579,7 +579,7 @@ Example:
 type config struct {
 	Port int
 	Name string
-	PathMap string `mapstructure:"path_map"`
+	PathMap string `viper:"path_map"`
 }
 
 var C config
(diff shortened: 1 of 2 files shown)


sagikazarmark:
Thanks for the contribution!

The linked PR does not seem to be on the road to being accepted.

You can use a `DecoderConfigOption` to overwrite the tag if you want, but mapstructure is fine IMHO as well.

I'm going to close this PR for now, because I don't really see the added value compared to the cost of change.

PR #450, opened 2018-01-30
Fix SafeWriteConfig

rchiossi:
If the config file does not exist and the force flag is not set,
OpenFile would not use O_CREATE flag, causing viper to fail with
error "File not exist" and to not create the config file.
This patch changes the behavior of writeConfig() so that if force is set
to false, OpenFile will use O_EXCL flag, thus failing if the file
already exists or creating a new file otherwise.

Signed-off-by: Rodrigo Chiossi <rodrigo.chiossi@intel.com>

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -1246,15 +1246,9 @@ func (v *Viper) writeConfig(filename string, force bool) error {
 	if v.config == nil {
 		v.config = make(map[string]interface{})
 	}
-	var flags int
-	if force == true {
-		flags = os.O_CREATE | os.O_TRUNC | os.O_WRONLY
-	} else {
-		if _, err := os.Stat(filename); os.IsNotExist(err) {
-			flags = os.O_WRONLY
-		} else {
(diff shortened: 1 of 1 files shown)


Nhoya:
Can someone please review this pull request?

Vaelatern:
This patch works for me, and the default behavior is indeed broken.

rchiossi:
Given that this is a very small PR and that it is waiting for review for over one year, I believe this project has sadly been abandoned.

Vaelatern:
@bep I'm sorry for pinging you directly, would you be able to merge this PR so as to fix this function?

estroz:
@bep @spf13 can you check this out/merge?

Vaelatern:
@sagikazarmark Could you possibly merge this when you have a chance? Thanks!

rchiossi:
Finally! \o/
Too bad I already moved away from Viper. Still, I'm happy to see someone is picking up on its development.

Vaelatern:
Celebration!

Pokerkoffer:
So when can we expect a new release containing this PR?

PR #454, opened 2018-02-05
New func GetConfigFilePath() which returns the actual config path

knorrkator:
Hi,

currently I need this function in my own project to log which config file is used. Actually is a wrapper around v.getConfigFile(). But refactoring v.getConfigFile() seems a bit too lot of work.
However. I'm sure sysadmins really are going to love this.

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -1738,7 +1738,7 @@ func (v *Viper) getConfigFile() (string, error) {
 //
 // Useful when you want to write the configPath to logfiles, stdout, etc.
 // Comes in handy during debugging or when running in production environments.
-func GetConfigFilePath() {
+func GetConfigFilePath() (string, error) {
 	return v.getConfigFile()
 }
 
--- a/viper.go
+++ b/viper.go
@@ -1734,11 +1734,11 @@ func (v *Viper) getConfigFile() (string, error) {
 	return v.getConfigFile()
 }
 
+// Returns the path where Viper actually found the logfile it uses.
+//
+// Useful when you want to write the configPath to logfiles, stdout, etc.
+// Comes in handy during debugging or when running in production environments.
 func GetConfigFilePath() {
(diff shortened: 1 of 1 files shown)


viper.go
```diff
@@ -1734,6 +1734,15 @@ func (v *Viper) getConfigFile() (string, error) {
 	return v.getConfigFile()
 }
 
+func GetConfigFilePath() (string, error) {
+	// Returns the path where Viper actually found the logfile it uses.
+	//
+	// Useful when you want to write the configPath to logfiles, stdout, etc.
+	// Comes in handy during debugging or when running in production environments.
+	cf, err := v.getConfigFile()
+	return cf, err
```
moorereason: Move your comment to a godoc comment and just do

```go
return v.getConfigFile()
```

knorrkator:
Hi, I've updated the PR accordingly

viper.go
```diff
@@ -1734,6 +1734,14 @@ func (v *Viper) getConfigFile() (string, error) {
 	return v.getConfigFile()
 }
 
+// Returns the path where Viper actually found the logfile it uses.
```
moorereason: A few things:

1. Godoc comments should begin with the element name.  `// GetConfigFilePath returns the path ...`.  See https://blog.golang.org/godoc-documenting-go-code.

1. You say *"where Viper actually found the **logfile** ..."*  Shouldn't that be *"configuration file"*?

2. I'd suggest removing the last two sentences describing use cases.  Just describe what the function does.

knorrkator: I updated the PR. Hope everything is find, now

knorrkator:
Hi,
want to ask, if this PR is going to be merged?

moorereason:
I don't have permission to merge, but I notice that you haven't signed the CLA.

moorereason:
Make sure you're git commits use the same email address as your Github user account.

PR #463, opened 2018-02-21
Cater for case-sensitive dependencies

pmoust:
Switching jww from jwalterweatherman to jWalterWeatherman,
`vgo` breaks otherwise.

```
vgo: import "github.com/elastic/<snipped>" ->
	import "github.com/elastic/<snipped>/cmd" ->
	import "github.com/spf13/viper" ->
	import "github.com/spf13/jwalterweatherman": module path of repo is github.com/spf13/jWalterWeatherman, not github.com/spf13/jwalterweatherman (wrong case)
```

### DIFF
--- a/util.go
+++ b/util.go
@@ -20,7 +20,7 @@ import (
 
 	"github.com/spf13/afero"
 	"github.com/spf13/cast"
-	jww "github.com/spf13/jwalterweatherman"
+	jww "github.com/spf13/jWalterWeatherman"
 )
 
 // ConfigParseError denotes failing to parse configuration file.
--- a/viper.go
+++ b/viper.go
@@ -42,7 +42,7 @@ import (
 	toml "github.com/pelletier/go-toml"
 	"github.com/spf13/afero"
 	"github.com/spf13/cast"
-	jww "github.com/spf13/jwalterweatherman"
+	jww "github.com/spf13/jWalterWeatherman"
 	"github.com/spf13/pflag"
 )
 


nate-johnston:
+1

pmoust:
Closes: #466

airylinus:
Anyone home? please kindly click the merge button. lol

stewartm83:
Second that. Please click the merge button, you are holding us back.

moorereason:
Related: https://github.com/spf13/jWalterWeatherman/issues/22

losinggeneration:
@pmoust try with a recent pull of vgo `go get -u golang.org/x/vgo` The issue seems to be fixed in vgo.

shantuo:
I tried with recent vgo and got no luck. Can we get this merged?

losinggeneration:
The case I tested was importing viper as a dependency. When I import viper and `jwalterweatherman` everything also works as expected. However, it absolutely fails when one project imports the lower case & another uses the mixed case. This isn't specific to vgo because both go & dep do not support this. 

Since vgo does not fail when importing viper anymore I'd consider this issue resolved. I'd also  generally suggest not importing jww because of the case sensitivity issues with the go toolchain & supporting tools.

andrewstuart:
LGTM :-)

h8liu:
I don't think this merge is doing the right thing.

it is breaking builds that contains other repos which has been using `jwalterweatherman` all along, like `hugo`.

bep:
See https://github.com/spf13/jWalterWeatherman/issues/22#issuecomment-425709215

The only correct thing is to rename the module. I will get that done. The canonical module name (all lowercase) is well established.

bep:
https://github.com/spf13/jwalterweatherman/issues/22#issuecomment-425752958

Issue #469, opened 2018-03-16
Detecting environment variables


hmhealey:
Is it possible to detect when a value has been read from an environment variable as opposed to the config file? We have a setup where we allow users to make changes from the UI that are written back to the config file, and we'd like to detect when a setting is overwritten by an environment variable so that we can prevent the user from changing that setting from the UI.

PR #479, opened 2018-04-06
Added Viper.EnvSettings method

hmhealey:
This is a followup to https://github.com/spf13/viper/issues/469 that adds a `Viper.EnvSettings` method which returns a map of all settings set through environment variables similar to `Viper.AllSettings`. I'm hoping it might be useful for someone else.

Let me know if there's anything that I'm missing or anything that should be done differently.

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -1675,6 +1675,26 @@ func (v *Viper) AllSettings() map[string]interface{} {
 	return m
 }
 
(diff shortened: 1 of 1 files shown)


onlyjob:
Any chance this could be merged so Mattermost could stop using their fork of this library?
Thanks.

sagikazarmark:
@hmhealey I wonder what the use case was for this and why you closed it.

PR #484, opened 2018-04-19
fix dep wrong case

oiooj:


h8liu:
the package's README actually suggests using lower cases:

https://github.com/spf13/jWalterWeatherman/blob/master/README.md

```
jww "github.com/spf13/jwalterweatherman"
```

it is also used in `hugo` with all lower cases:

https://github.com/gohugoio/hugo/search?q=jWalterWeatherman&unscoped_q=jWalterWeatherman

Using the capital case version will cause build failures when trying to go get both hugo and this package:

https://gopkgs.io/builds/20180828-074824-e4e0ab4a

I hence suggest revert this change.

Issue #422, opened 2017-12-03
Segfault When binding to nil PFlag


Toasterson:
Hello 

The Latest version of viper segfaults when binding with pflags.

Edit:
Turns out I am a moron and this does not happen with the code all the time but when you try to bind to a flag that does not exist.

Say you have
```go
RootCmd.PersistentFlags().String("loglevel", "", "Set the Log Level")
viper.BindPFlag("loglevel", RootCmd.Flags().Lookup("loglevel"))
```
This Will cause the Segfault below.

Cerrectly you would need
```go
RootCmd.PersistentFlags().String("loglevel", "", "Set the Log Level")
viper.BindPFlag("loglevel", RootCmd.PersistentFlags().Lookup("loglevel"))
```
Notice the PersistenFlags in both lines.

This is not easy to debug when you are totaly blind like I am.
Can we have an error Message that tells that the pointer to PFlag is nil and one should revisit the code?

Tested with Go 1.9 and 1.8.3 on Linux and Solaris. Same result both times.

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x50 pc=0x822ed5]

goroutine 1 [running]:
github.com/spf13/viper.pflagValue.HasChanged(0x0, 0xc4200df5c0)
	/export/home/toast/workspace/src/github.com/spf13/viper/flags.go:41 +0x5
github.com/spf13/viper.(*Viper).find(0xc4200fc2d0, 0xb75128, 0x8, 0x8, 0xc420160039)
	/export/home/toast/workspace/src/github.com/spf13/viper/viper.go:893 +0xf17
github.com/spf13/viper.(*Viper).Get(0xc4200fc2d0, 0xb75128, 0x8, 0xb71996, 0x5)
	/export/home/toast/workspace/src/github.com/spf13/viper/viper.go:594 +0x7c
github.com/spf13/viper.(*Viper).GetString(0xc4200fc2d0, 0xb75128, 0x8, 0xc4201620a0, 0x0)
	/export/home/toast/workspace/src/github.com/spf13/viper/viper.go:649 +0x3f
github.com/spf13/viper.GetString(0xb75128, 0x8, 0xc4201600f0, 0xc42016e0f0)
	/export/home/toast/workspace/src/github.com/spf13/viper/viper.go:647 +0x41
[...]

PR #487, opened 2018-04-23
Allow Sub Viper to work with PFlags & nil check on binding PFlags

### LINKED ISSUE #368: Binding flags as nested keys does not work
```
cmd.Flags().IntP("port", "p", 80, "Listening port of the server")
viper.BindPFlag("server.port", cmd.Flags().Lookup("port"))
```
Then `viper.Get("server")` is nil.

I would expect it to be `map[port:80]`

A result is that `viper.UnmarshalKey("server", &cfg)` does not work.

Actually I did not found any way to bind a flag `server.port` in a ways that `viper.UnmarshalKey(...)` would work correctly. Even when naming the flag  `"server.port"`

losinggeneration:
`Sub` with pflags didn't work as I expected because the key prefix wasn't checked for flags. This should allow binding pflags to behave very similar to structured configurations loaded from files.

The nil check is needed since it's put into the pflagValue* structure that never checks if the PFlag pointer is valid.

This has code partially based on #331 for b7a4909 that uses the same `flag` name & updated comment for `flag`. Though the comment isn't totally correct for this PR, but #331 IMO should be merged.

This fixes #368 & #422

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -811,6 +811,9 @@ func (v *Viper) UnmarshalExact(rawVal interface{}) error {
 // name as the config key.
 func BindPFlags(flags *pflag.FlagSet) error { return v.BindPFlags(flags) }
 func (v *Viper) BindPFlags(flags *pflag.FlagSet) error {
+	if flags == nil {
+		return fmt.Errorf("FlagSet cannot be nil")
+	}
 	return v.BindFlagValues(pflagValueSet{flags})
 }
 
@@ -822,6 +825,9 @@ func (v *Viper) BindPFlags(flags *pflag.FlagSet) error {
 //
 func BindPFlag(key string, flag *pflag.Flag) error { return v.BindPFlag(key, flag) }
 func (v *Viper) BindPFlag(key string, flag *pflag.Flag) error {
+	if flag == nil {
+		return fmt.Errorf("flag for %q is nil", key)
+	}
 	return v.BindFlagValue(key, pflagValue{flag})
 }
 
--- a/viper.go
+++ b/viper.go
@@ -612,7 +612,7 @@ func GetViper() *Viper {
 func Get(key string) interface{} { return v.Get(key) }
 func (v *Viper) Get(key string) interface{} {
 	lcaseKey := strings.ToLower(key)
-	val := v.find(lcaseKey)
+	val := v.find(lcaseKey, true)
 	if val == nil {
 		return nil
 	}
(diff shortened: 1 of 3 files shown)


losinggeneration:
It appears this may have an issue with having overrides work as expected. For instance, if both `v.config` & `v.pflags` are set, Sub doesn't respect the search order for values. This is partially because Sub flattens the entire config into `v.conf` after the call to `v.Get`. It might be less surprising if Sub worked more like a sub-tree clone of the parent Viper copying `override` `pflags` `kvstore` etc

robdefeo:
Are there any blockers on merging this or work around for the issues raised?

viper.go
```diff
@@ -879,9 +885,12 @@ func (v *Viper) BindEnv(input ...string) error {
...
 // flag, env, config file, key/value store, default.
-// Viper will check to see if an alias exists first.
-// Note: this assumes a lower-cased key given.
-func (v *Viper) find(lcaseKey string) interface{} {
+// Viper will then check in the following order:
+// flag, env, config file, key/value store.
+// Lastly, if no value was found and flagDefault is true, and if the key
+// corresponds to a flag, the flag's default value is returned.
+//
+func (v *Viper) find(lcaseKey string, flagDefault bool) interface{} {
```
estroz: @losinggeneration `flagDefault` is not being used anywhere.

losinggeneration: > This has code partially based on #331 for b7a4909 that uses the same flag name & updated comment for flag. Though the comment isn't totally correct for this PR, but #331 IMO should be merged.

I probably shouldn't have pulled that along from #331 for that commit. I can revert that part since it's not used here.

stevefan1999-personal:
@spf13 any chance if you would like to re-evaluate the PR? In the meanwhile I will have to manually copy the structure string value to each struct field, and I'll possibly do it by reflection...

bouzafr:
+1 this PR. Please.

xy3:
+1 this PR

dheeraj-coding:
+1 this PR

codexetreme:
+1 on this PR

mazenharake:
+1

Just a side note regarding 1750e0a, wouldn't that be redundant since there is a check further down in `BindFlagValue`?
See: https://github.com/spf13/viper/pull/487/commits/1750e0a9d30f43164393d9699de5302e6393f445#diff-786d4c5d3482047f0ac7e62876791b9f4ba727a70acd750cf0a1bdd19c51c3f1R989

PR #495, opened 2018-05-06
Fix environment variable expansion in absPathify

biochimia:
- Don't expand user home directory for variable names that simply have a
  HOME prefix;
- Support expansion of variables not followed by the path separator.

### DIFF
--- a/util.go
+++ b/util.go
@@ -91,13 +91,23 @@ func insensitiviseMap(m map[string]interface{}) {
 func absPathify(inPath string) string {
 	jww.INFO.Println("Trying to resolve absolute path to", inPath)
 
-	if strings.HasPrefix(inPath, "$HOME") {
+	if strings.HasPrefix(inPath, "$HOME") &&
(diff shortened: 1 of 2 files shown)


qjcg:
@sagikazarmark Please review this PR when you have a moment, it fixes an real issue (see my `lefthook` reference above for an example).

util.go
```diff
@@ -91,13 +91,23 @@ func insensitiviseMap(m map[string]interface{}) {
 func absPathify(inPath string) string {
 	jww.INFO.Println("Trying to resolve absolute path to", inPath)
 
-	if strings.HasPrefix(inPath, "$HOME") {
+	if strings.HasPrefix(inPath, "$HOME") &&
```
sagikazarmark, line 94: Could this be rewritten as something like `inPath == "$HOME" || strings.HasPrefix(inPath, "$HOME"+os.PathSeparator)` or equivalent (or even using some sort of splitting)? The current condition is not really expressive.

Side note: looking at the `path/filepath` package there isn't much that can help here, but ideally it would be nice.

Issue #507, opened 2018-06-02
Sub() ignores variables set with BindEnv/AutomaticEnv


lucasvo:
The following example fails because Sub() because variables overwritten in env are ignored.

```yaml,
grandParent:
    parent:
        child: bar
```

```go,
package main

import (
	"fmt"
	"github.com/spf13/viper"
	"os"
	"strings"
)

var v *viper.Viper

func main() {
	os.Setenv("GRANDPARENT_PARENT_CHILD", "foo")
	v = viper.New()
	v.SetConfigType("yaml")
	v.AutomaticEnv()
	v.SetEnvKeyReplacer(strings.NewReplacer(".", "_"))
	f, _ := os.Open("test.yaml")
	v.ReadConfig(f)
	sub := v.Sub("grandParent.parent")
	fmt.Println("Direct", v.GetString("grandParent.parent.child"))
	fmt.Println("Sub", sub.GetString("child"))
	if sub.GetString("child") != v.GetString("grandParent.parent.child") {
		panic("sub ignores os.environ")
	}
}
```

The output of the program is:

```
Direct foo
Sub bar
panic: sub ignores os.environ
[...]
```

lucasvo:
Related to: #307

vail130:
Ran into this issue, which took a couple hours of debugging to discover. I'd love to see some explicit documentation about this limitation, if not a fix.

lucasvo:
@spf13 would be awesome to get #508 merged! I am sure @vail130 and I are not the only ones here...

guyisra:
any updates on this?

89z:
https://github.com/spf13/viper/issues/140

https://github.com/spf13/viper/issues/507
https://github.com/spf13/viper/issues/522
https://github.com/spf13/viper/issues/576
https://github.com/spf13/viper/issues/911

Issue #307, opened 2017-01-28
Sub vipers are not fully functional


jbeda:
Lots of things in Viper are stored using the dotted path and don't make their way into Viper instances created via `Sub()`.  This includes things like `Unmarshal`

Example: 
```
package main

import (
	"fmt"

	flag "github.com/spf13/pflag"
	"github.com/spf13/viper"
)

type Config struct {
	Test   int
	Nested NestedConfig
}

type NestedConfig struct {
	Test int
}

func main() {
	var ncfg NestedConfig

	flag.Int("test", 0, "Test")
	flag.Int("nested.test", 1, "Nested test")
	viper.Set("nested.test", 2)

	// Without this line we crash as Sub returns nil
	viper.Set("nested", map[string]interface{}{})

	sub := viper.Sub("nested")
	err := sub.Unmarshal(&ncfg)
	if err != nil {
		panic(err)
	}

	fmt.Printf("%#v\n", ncfg)
}
```

When run the output:
```
main.NestedConfig{Test:0}
```

I'm guessing the "real" fix for this is to make Viper hierarchical at its core.  Instead of storing dotted things in the root viper, make sub vipers manage flags and such themselves.

PR #508, opened 2018-06-02
Update "viper.Sub()" doc string

lucasvo:
This documents the shortcomings of `Sub` described in #507 and #307. I know reworking viper.Sub() would be better than just adding a comment. Unfortunately time doesn't allow for more right now.


lucasvo:
@spf13 would be great to get some feedback from you on this!

sagikazarmark:
@lucasvo the comment is not 100% correct. Unmarshal doesn't work with flags, but works with other sources of config.

PR #509, opened 2018-06-06
config need sub Viper instance List

yang1992:
for example:
{"config": ["{"sub_config":xxx}"]}

viper.Get("config") returns [map[sub_config:xxx]]，but we hope all return data is viper, so we can use viper function to get, not map or any other struct.

Each viper instance need returns sub viper intance list which can deep traverse data.

### DIFF
--- a/.idea/vcs.xml
+++ b/.idea/vcs.xml
@@ -1,6 +1,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<project version="4">
-  <component name="VcsDirectoryMappings">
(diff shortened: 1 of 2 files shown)


andig:
@yang1992 your PR makes perfect sense and is exactly what I need in order to pre-populate objects and use `viper.Unmarshal` to apply additional settings.

Would you mind signing the CLA so this could get merged?

ps23:
Can this be signed?

andig:
@ps23 #863 is a signed replacement.

swengineer404:
Can we get an update on this please?

PR #512, opened 2018-06-08
Add global UnmarshalExact method

greghaynes:
There is no helper method for UnmarshalExact which calls the
corresponding method on the global viper instance.

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -798,6 +798,7 @@ func decode(input interface{}, config *mapstructure.DecoderConfig) error {
 
 // UnmarshalExact unmarshals the config into a Struct, erroring if a field is nonexistent
 // in the destination struct.
+func UnmarshalExact(rawVal interface{}) error { return v.UnmarshalExact(rawVal) }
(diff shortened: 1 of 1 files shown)


andig:
Ran into the same thing. Problem is that this method does not make so much sense. You may have more or different flags or pflags than your config file's struct supports. In that case `UnmarshalExact` would fail on those.

andig:
I think the underlying problem is that there is no option to:

- obtain the raw config file from the viper instance except for reading it again
- obtain a copy of the viper instance after reading config file and *before* adding (p)flags and other config sources

andig:
ping @sagikazarmark would it be possible to merge this? With the workaround mentioned above the method could be really helpful.

sagikazarmark:
I guess it could, but `viper.GetViper().UnmarshalExact()` works just as well. I will take a look at this later.

andig:
Oh yeah- I didn‘t notice GetViper... thank you! For sake of symmetry this would still make sense.

PR #517, opened 2018-06-21
add GetUint/GetUint32/GetUint64

smoreg:
Hi! There is no Gets for Uints


deterok:
Please add these changes :crying_cat_face:

nathanhack:
Nice work @smoreg  could you also add the uint bits under:
func Get(key string) interface{} {...} (in viper.go) in this PR?

sagikazarmark:
This is already done.

PR #521, opened 2018-06-28
Support customising mapstructure.DecoderConfig for Unmarshal

krak3n:
Hi

I recently ran into an issue when using `viper.Unmarshal` in-conjunction with Environment variables. I wished to pass a `json` object as a value to an Environment variable for user names and passwords that I didn't want to write to a config file, for example:

```
export MY_CONFIG_VAL={\"foo\":\"bar\",\"fizz\":\"buzz\"}
```

 Whilst `viper.GetStringMapString` would work perfectly, `Unmarshal` would fail with: 

```
* 'my_config_val' expected a map, got 'string'
```

I did some digging and saw this was failing in `mapstructure`. I saw it supported custom `DecodeHook` functions so I could write my own function to parse the value correctly, however `viper` does not support the ability to customise the `mapstructure.DecoderConfig`.

So this PR adds the following (it's worth noting this is 100% backwards compatible thanks to variadic arguments):

* Added a new `DecoderConfigOption` type allowing the user to write custom
  functions that can override the default mapstructure.DecoderConfig
  settings

* Added a new `DecodeHook` function which returns
  a `DecoderConfigOption`. This allows the user to easily set their own
  Decode hooks when Unmarshaling

* Updated Unmarshal, UnmarshalKey and defaultDecoderConfig to support variadic
  trailing `DecoderConfigOption` functions to allow for customisation of
  the default  mapstructure.DecoderConfig

* Added a test case with example usage

I look forward to your review and feedback :smile: 

p.s and thank you for this wonderful library.

**Edits: spelling**

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -113,6 +113,23 @@ func (fnfe ConfigFileNotFoundError) Error() string {
 	return fmt.Sprintf("Config File %q Not Found in %q", fnfe.name, fnfe.locations)
 }
 
+// A DecoderConfigOption can be passed to viper.Unmarshal to configure
+// mapstructure.DecoderConfig options
+type DecoderConfigOption func(*mapstructure.DecoderConfig)
+
+// DecodeHook returns a DecoderConfigOption which overrides the default
+// DecoderConfig.DecodeHook value, the default is:
+//
+//  mapstructure.ComposeDecodeHookFunc(
+//		mapstructure.StringToTimeDurationHookFunc(),
+//		mapstructure.StringToSliceHookFunc(","),
+//	)
+func DecodeHook(hook mapstructure.DecodeHookFunc) DecoderConfigOption {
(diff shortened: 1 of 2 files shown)


krak3n:
Also I believe this may also be a solution to #519. Since the user could instead customise the default existing configuration instead, for example:

``` go
func ViperStructTag() viper.DecoderConfigOption {
	return func(c *mapstructure.DecoderConfig) {
		c.TagName = "viper"
	}
}
```

And then calling `Unmarshal` with the function above as an Option:

``` go
type config struct {
	Foo string `viper:"foo"`
}
var C config
viper.Unmarshal(&C, ViperStructTag())
```

krak3n:
@spf13 @bep any feedback on this one?

krak3n:
:sob:

krak3n:
yey! thank you @bep

Issue #529, opened 2018-07-09
Enable Marshalling to string


bricef:
I'd like to marshall the config to a string (or to a file-like object) so that it can be written to stdout or used in some other way.

This isn't currently allowed as the [Marshal functions ](https://github.com/spf13/viper/blob/master/viper.go#L1333) are private.

bricef:
This can be solved by using the marshalling the entire config manually. for example:
```golang
c := viper.AllSettings()
bs, err := yaml.Marshal(c)
if err != nil {
    // handle marshalling error
}
fmt.Print(string(bs))
```

PR #530, opened 2018-07-09
WIP: Expose MarshalWriter publicly

bricef:
Resolves https://github.com/spf13/viper/issues/529 by exposing Marshalling functions


bricef:
Exposing `afero.File`  can get you somewhere, but it's not an ideal interface. Marking back to WIP for now.

bricef:
Not needed. We can use 
```
c := viper.AllSettings()
bs, err := yaml.Marshal(c)
fmt.Print(string(bs))
```
instead.

PR #531, opened 2018-07-09
Add example of marshalling to string

bricef:
This is a useful usecase, which is not documented.

Maybe this could be brought into viper itself with the same format string that are currently supported by `WriteConfigAs()` and friends. This could also simplify the `v.marshalWriter()` function by removing error handling duplication.


spf13:
LGTM

PR #535, opened 2018-07-19
if config type is already set, ignore file extension

cmohrb:
…and use the save…d config type

e.g. if viper.SetConfigType("yaml") was previously called, yaml will be used as
config format, even if the file is called XXX.conf, instead of XXXX.yaml


sagikazarmark:
I'm failing to see the connection between the description and the contents of this PR.

fengjian1993:
did [if config type is already set, ignore file extension and use the save…] merged?

Issue #542, opened 2018-07-27
WriteConfig() does not close the File resource


bpizzi:
Function writeConfig() (which is very useful!) currently does not call Close() after writing. I encountered this one because I wanted to rename the config file just after having written to it (root cause is that there would be a need for aliases in SupportedExts...).

I'll attach a PR.

PR #543, opened 2018-07-27
Fixed missing f.Close() in writeConfig()

bpizzi:
Defering can cause trouble because we're writing to the file outside the function
where the defering is registered, calling f.Sync() ensures that bytes are flushed
to disk even is Close() is called too soon.

Ref #542


sagikazarmark:
Thanks @bpizzi

PR #552, opened 2018-08-12
Fix example code

okazu-dm:
hello
I found example code doesn't work.

I checked with following code.

```
package main

import (
	"fmt"

	"github.com/spf13/viper"
)

// Config sample
type Config struct {
	Id int
}

func main() {
	var c *Config
	viper.Unmarshal(&c) // this work
	// Unmarshal(&c) // this doesn't work
	fmt.Println(c.Id)
}

```

Thank you.

### DIFF
--- a/README.md
+++ b/README.md
@@ -585,7 +585,7 @@ type config struct {
 
 var C config
 
-err := Unmarshal(&C)
+err := viper.Unmarshal(&C)
 if err != nil {
 	t.Fatalf("unable to decode into struct, %v", err)
 }


aarti:
What would be great is if you could add an /examples folder and place your example in an unmarshal_example.go file. We could start to build out these examples instead of having them in  the readme. What do you think @spf13

daixiang0:
Any chance to merge it?

sagikazarmark:
Thanks!

PR #555, opened 2018-08-16
Adding a method to merge config in the override section

hush-hush:
**What is does**

This allows users to set configuration that would have precedence on env
variable. This is very useful when some post-processing is done on the
configuration.

Currently there is no way to override part of the configuration with the highest priority.

**Use Case**

Here is our use case: we load a yaml configuration file where every default might be overridden by the environment (to be easily used inside a container).

After loading the configuration, we have some post-processing on it. Some secrets might be present in the conf (from the yaml file or the env) and we hit a secret management system to resolve them (KMS, Vault, ...). We now need to update the main configuration with the new setting being retrieved before the one in the env.

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -1231,6 +1231,22 @@ func (v *Viper) MergeConfig(in io.Reader) error {
 	return nil
 }
 
+// MergeConfigOverride merges a new configuration within the config at the
+// highest lever of priority (similar to the 'Set' method). Key set here will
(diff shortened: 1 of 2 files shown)


hush-hush:
@spf13 or @aarti: would love some feedback on this is possible.

hush-hush:
@spf13 @aarti last ping in the hope to get some feedback on this PR !

PR #564, opened 2018-09-15
String Array support for pflag

pjdufour:
This PR allows `viper` to pull an array of string values directly from `pflag` instead of having to encode/decode as CSV.  Depends on https://github.com/spf13/pflag/pull/181

### DIFF
--- a/flags.go
+++ b/flags.go
@@ -14,6 +14,7 @@ type FlagValue interface {
 	HasChanged() bool
 	Name() string
 	ValueString() string
+	ValueStringArray() []string
 	ValueType() string
 }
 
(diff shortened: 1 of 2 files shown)


flags.go
```diff
@@ -14,6 +14,7 @@ type FlagValue interface {
 	HasChanged() bool
 	Name() string
 	ValueString() string
+	ValueStringArray() []string
```
sagikazarmark: This is a breaking change in the interface.

pjdufour: Yep.  This is dependent on https://github.com/spf13/pflag/pull/181 being merged first.

sagikazarmark: I mean it's a breaking change for anyone using custom flag types and for the 1.x version there is a backward compatibility promise.

sagikazarmark:
Viper already supports stringArray

PR #566, opened 2018-09-17
config need sub Viper instance List

profbiss:
Close #509


sagikazarmark:
Why not just use `viper.Get("config.sub_config")`?

Issue #570, opened 2018-09-30
Gocritic 3.3 warnings: typeSwitchVar


KosToZyB:
Find using: https://go-critic.github.io/overview#typeSwitchVar-ref
```
$GOPATH/src/github.com/spf13/viper/util.go:71:3: typeSwitchVar: case 1 can benefit from type switch with assignment
$GOPATH/src/github.com/spf13/viper/viper.go:494:3: typeSwitchVar: case 1 can benefit from type switch with assignment
$GOPATH/src/github.com/spf13/viper/viper.go:538:4: typeSwitchVar: case 1 can benefit from type switch with assignment
$GOPATH/src/github.com/spf13/viper/viper.go:1678:3: typeSwitchVar: case 0 can benefit from type switch with assignment
$GOPATH/src/github.com/spf13/viper/overrides_test.go:158:3: typeSwitchVar: case 1 can benefit from type switch with assignmen
```

sagikazarmark:
See the linked PR.

PR #571, opened 2018-09-30
fixed typeSwitchVar warnings

KosToZyB:
fixed https://github.com/spf13/viper/issues/570


sagikazarmark:
Some of the changes have already been made. The rest is good as it is IMHO.

Issue #572, opened 2018-10-04
Require configs are present


caalberts:
In my projects at work, we have been wrapping viper to add checks to ensure required config values have been set, either through environment variables or config file.

The reason for this is to avoid runtime panics when a critical config is missing.

Something like:
```go
func Require(key string) (interface{}, error)
```

Has viper considered this functionality? If it's worth adding to viper, I can work on a PR.

gabeduke:
@caalberts what is remaining to start a PR? This feature is quite useful since cobra `RequiredFlags` is not usable after binding the flag to viper.

caalberts:
I have actually created a PR #573 for this.

sagikazarmark:
Please see my comment: https://github.com/spf13/viper/pull/573#issuecomment-759630829

PR #573, opened 2018-10-04
Implements #572 - `viper.Require`

caalberts:
`viper.Require` and the related functions gets the value associated
to the given key, returns an error if the key is not set.

This can be used to handle missing keys within applications and
prevent runtime panics

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -789,6 +789,165 @@ func (v *Viper) GetSizeInBytes(key string) uint {
 	return parseSizeInBytes(sizeStr)
 }
 
+// Require returns the value associated with the given key or
+// an error if the key is not set.
+func Require(key string) (interface{}, error) { return v.Require(key) }
+func (v *Viper) Require(key string) (interface{}, error) {
+	value := v.Get(key)
+	if value == nil {
+		return nil, fmt.Errorf("key %s is not set", key)
+	}
+	return value, nil
+}
+
+// RequireString returns the value associated with the key as a string or
+// an error if the key is not set.
+func RequireString(key string) (string, error) { return v.RequireString(key) }
+func (v *Viper) RequireString(key string) (string, error) {
+	value, err := v.Require(key)
+	if err != nil {
+		return "", err
+	}
+	return cast.ToString(value), nil
+}
+
+// RequireBool returns the value associated with the key as a boolean or
+// an error if the key is not set.
+func RequireBool(key string) (bool, error) { return v.RequireBool(key) }
+func (v *Viper) RequireBool(key string) (bool, error) {
+	value, err := v.Require(key)
+	if err != nil {
+		return false, err
+	}
+	return cast.ToBool(value), nil
+}
+
+// RequireInt returns the value associated with the key as an integer or
+// an error if the key is not set.
+func RequireInt(key string) (int, error) { return v.RequireInt(key) }
+func (v *Viper) RequireInt(key string) (int, error) {
+	value, err := v.Require(key)
+	if err != nil {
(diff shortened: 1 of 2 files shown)


gabeduke:
I'm unable to recreate the test failure locally. Do you know how to repro? Here are the steps I took:

```bash
git fetch origin pull/573/head:require-tests
git checkout require-tests

go get -t -v ./...
go install ./...
go test -v ./...
```

These are the same cmds from the Travis job, am I missing anything? One thing I have noticed is that the line number on the Travis job indicates the line number is 629 (`viper_test.go:629:`) but in the file locally the function is on line 596. It's possible my environment is not set up correctly...

caalberts:
@gabeduke I've fixed the failing test following #596, but I'm getting into another failure on CI which I can't reproduce locally.

viper.go
```diff
@@ -789,6 +789,165 @@ func (v *Viper) GetSizeInBytes(key string) uint {
 	return parseSizeInBytes(sizeStr)
 }
 
+// Require returns the value associated with the given key or
+// an error if the key is not set.
+func Require(key string) (interface{}, error) { return v.Require(key) }
+func (v *Viper) Require(key string) (interface{}, error) {
+	value := v.Get(key)
+	if value == nil {
```
sagikazarmark: The problem is that `nil` can be a perfectly valid value. For example after manually calling `viper.Set`.

brendanstennett:
@caalberts Are you still interested in getting this feature merged? I think it's a great idea!

I would recommend using `v.IsSet` instead of checking if the value is `nil` since as @sagikazarmark points out, `nil` is indeed a valid value (sadly).

If you don't have the time, I would be happy to pick it back up and get it across the finish line.

viper.go
```diff
@@ -789,6 +789,165 @@ func (v *Viper) GetSizeInBytes(key string) uint {
...
+		return nil, fmt.Errorf("key %s is not set", key)
```
brendanstennett: Since we are calling this "required" would it make more sense to just to panic in this circumstance?

sagikazarmark: Panic is for errors from which the program cannot recover from. While in some cases this might be true for this Required function, when calling it somewhere in the middle of the program it might not be.

But TBH I don't think this function is a good idea. Viper is for consolidating and unmarshaling configuration. Validation shouldn't be a concern of Viper IMHO.

brendanstennett: While I do agree there are likely certain circumstances that would read a config in the middle of program execution, I think that's probably not the most common case to do so and then you would probably just not want use this functionality and use `GetX` instead.

I think for the vast majority of cases where this is useful, a program would have no way of recovering from a missing param being passed in (such as a missing DB_URL environment variable) and having this functionality can cause very useful error message during startup.

ofpiyush:
Is this feature getting merged?

hypergig:
I would love to see a whole variant set of `Get` functions with `OrFail` suffixes. eg

```golang
viper.GetOrFail("foo.bar")
viper.GetStringOrFail("foo.bar")
```
That would be pretty darn userful

sagikazarmark:
Hey guys,

Sorry for the late reply.

As I explained earlier, `Required` as a feature would complicate the library with very little benefit and there are more elegant solutions to this problem.

The `OrFail` variant for `Get` functions actually sounds interesting, but I'm not completely sold on it yet. Maybe open a discussion about it (or just reply to this one: https://github.com/spf13/viper/discussions/1060).

Closing as this feature won't land. Thanks for your work @caalberts anyway.

PR #581, opened 2018-10-18
Add method `GetStringEnv` which will interpolate environmental variables

useafterfree:
This will interpolate values that look like `${HOME}` and assign with $HOME from environment.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,10 +6,12 @@ require (
 	github.com/magiconair/properties v1.8.0
 	github.com/mitchellh/mapstructure v1.0.0
 	github.com/pelletier/go-toml v1.2.0
+	github.com/pmezard/go-difflib v1.0.0 // indirect
 	github.com/spf13/afero v1.1.2
 	github.com/spf13/cast v1.2.0
 	github.com/spf13/jwalterweatherman v1.0.0
 	github.com/spf13/pflag v1.0.2
+	github.com/stretchr/testify v1.2.2
 	golang.org/x/sys v0.0.0-20180906133057-8cf3aee42992 // indirect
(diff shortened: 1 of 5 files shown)


andrewstuart:
Hey, thanks for the contribution!

I'm hesitant to bring in a feature like this, for a couple reasons.

- It's quite specific to environment variables, compared to the rest of viper, which is concerned with aggregating and overlaying multiple configuration sources.
- This could be easily achieved by simply wrapping with `os.ExpandEnv`, e.g. `os.ExpandEnv(viper.GetString("foo"))`.

useafterfree:
The README.md states Viper states that Viper supports:
`reading from environment variables`

Maybe that can read:
`reading from specifically formatted unique environment variables`

Since, 
```
SECTION:
   CONFIG: '${AMI}'
OTHER_SECTION:
   CONFIG: '${AMI}'
```

will never work for just `$AMI` but `$SECTION_CONFIG_AMI` `$OTHER_SECTION_CONFIG_AMI`

If I use environment variables in several places in a file, I don't want to make new ones as 
my expectation is Environmental variables are to be used as one -> many.

Why would that change if I use Viper?

sagikazarmark:
How about some configuration that optionally turns on variable expansion (eg. calls `os.ExpandEnv) for environment variables?

Alternatively variable expansion could be an option for not just env vars, but all configurations.

astockwell:
I would like to second the request and add my usecase: configuring multiple environments with the same yaml/toml field structure, but needing different values:

```toml
[qa]
host = '${HOSTQA}'

[prod]
host = '${MAINLB}'
```

PR #582, opened 2018-10-19
Fixed unmarshalReader not returning an error

f0rmiga:
Quick fix on unmarshalReader not returning an error when the input `io.Reader` returns an error. I caught that while mocking the input `io.Reader` and making the `Read` method to return an error on a unit test.


f0rmiga:
@sagikazarmark will add ASAP!

PR #590, opened 2018-11-02
Added SetDefaults (multiple) func accepting map[string]interface{}.

WillDrug:
Quality of life function, making it easier to populate defaults for a config.


WillDrug:
Travis failed on an existing test in master somehow.
If noone merges this I'll go and check how to fix this, then...

PR #591, opened 2018-11-06
Added opts ...DecoderConfigOption to UnmarshalExact

nmiculinic:
This makes the signature the same as ordinary Unmarshal with optional options added.


nmiculinic:
Could you merge override since I don't have permissions unless tests pass (and the failing test is flaky and has nothing to do with this PR)?

sagikazarmark:
Yeah, I agree.

/cc @bep

nmiculinic:
What's the status big this PR?

Issue #373, opened 2017-07-28
UnmarshalKey normalizes all map keys to lower case


lnashier:
```
{
   "someKey": 1,
   "myMap": {
       "A": "v1",
       "B": "v2"
   }
}
```

```
config := viper.New()
config.AddConfigPath("myconfigpath")
config.SetConfigName("myconfig")
```

```
var myMap map[string]string
config.UnmarshalKey("myMap", &myMap)
```

`myMap` has all lower-cased keys. 

Can somebody comment if it is expected or is it a bug ?

knadh:
This doesn't seem to be documented anywhere but internally, Viper lowercases all keys which is weird: https://github.com/spf13/viper/blob/d9cca5ef33035202efb1586825bdbb15ff9ec3ba/viper.go#L1303

This goes against the YAML, TOML, and JSON specs where keys are case sensitive.

lnashier:
I will request viper authors to comment.

lnashier:
@spf13 any comment?

lnashier:
#482

wjdp:
Just came across this today. Is there an option or workaround to disable this behaviour?

lnashier:
@wjdp #482 https://github.com/lnashier/viper

lnashier:
#666 fixes it

codemodify:
This issues is still there

knadh:
@codemodify https://github.com/spf13/viper/pull/635 has been pending for almost a year and should give you better context on the state of this issue.

ackratos:
want a fix for this issue :(

sagikazarmark:
@codemodify @ackratos this is not an easy fix as it goes directly against core Viper principles and goals. I'm working on it, but please keep in mind this is an open source project with a lot of users. A major change like this is not easy to pull in.

codemodify:
Thank you, completely understand, will be ok to have this highlighted on the front page that can be used for making decisions when picking up tools ? The reason for that is half way through it is hard to drift away when you understand the limits

tomqwpl:
I've just hit this too, and it's most unexpected. I'm demarshalling a struct that contains a map. The struct fields demarshall fine but the map keys all come out lower case, which is unexpected.

fefe982:
I had to make the map into a slice to work around this, making the key a field value of the slice element. Lucky I have control of the data structure.

But still want to see viper handling case-sensitive map keys locally.

farukarakaya:
[...]

Issue #293, opened 2016-12-22
key auto lower?


elvizlai:
```yaml
ENV:
    REPO: git@github.com:xyz/abc.git
    WORKDIR: /gopath/src/github.com/xyz/abc
```

```go
type conf struct {
    ENV    map[string]string
}

```

Updating viper to newest, the key are all lower.(Previous version works well)

bep:
Yes, that is a change. The Get* methods are also case insensitive.

elvizlai:
Any way to make it sensitive? It is not reasonable.

sagikazarmark:
See #1014

Issue #260, opened 2016-10-14
Loading a YAML file makes all keys lowercase, but YAML keys are case sensitive.


barsanuphe:
Since the recent changes, all keys have become lowercase. 
If I'm not mistaken ([see this reply from YAML spec author](https://sourceforge.net/p/yaml/mailman/message/12307333/)), YAML keys are case sensitive. 
For instance, calling GetStringMapStringSlice() while loading a YAML file makes all of the keys of the resulting map lowercase, which is unexpected and was not the previous behavior.

bep:
>GetStringMapStringSlice() while loading a YAML file

When do you call that method while loading a YAML file?

barsanuphe:
If you have: 

```
yaml_key:
         Case Sensitive Key: 
               - value1
               - value2
```
GetStringMapStringSlice("yaml_key") now gives a map with a "case sensitive key" key.

bep:
To conclude: Viper have always *tried* to be case-insensitive by doing a lower-casing of the keys given. It was however not being doing that deeply for maps. Now that is fixed and we are seeing some breakage in the wild.

The main problem seems to be the expectation that you should be able to unmarshal the config map into a struct or similar.

Because for plain lookups, like `viper.Get("a.b.A.d.")`, this is clearly an improvement.

@spf13 should chime in here.

bep:
>GetStringMapStringSlice("yaml_key") now gives a map with a "case sensitive key" key.

No, it gives a lower-cased key.

barsanuphe:
> No, it gives a lower-cased key.

Yes, I meant it gives the value: `case sensitive key` in my example.

I understand your point of view that this is the intended behavior, but I find it surprising for YAML. And I'm not expecting to unmarshall the config map into anything, I'm happy dealing with a map[string][]string. I was just relying on the keys to have the same case I put in the YAML file.

My "top" keys are lowercase though, so I didn't realize viper was already returning lowercased versions for anything but maps.

bep:
I understand your situation -- I have a similar breakage in one of my apps. We should probably have handled this in a better way, but I believe it is the intended behavior of Viper from the start. @spf13 should decide.

spf13:
The way I think it should work is a bit of a hybrid of case sensitivity.

Keys should be treated as case insensitive when getting, setting or
merging. I can't see another approach that would make sense here.

Returning keys would ideally be in the case originally defined.

The challenge here is that while yaml is case sensitive, not all of the
[...]

PR #592, opened 2018-11-07
added a setting to make viper case sensitive

e-nikolov:
In some cases it is useful to preserve the case sensitivity of keys. This change adds the option to call viper.SetCaseSensitive(true) to enable case sensitivity.

This feature is needed for example when unmarshalling a config into a map[string]interface{}, changing some of the config values and then saving it again to a file. If the keys are always lowercased, the structure of the regenerated config will be changed, which is undesirable.

Addressing #260, #293, #373

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -203,6 +203,7 @@ type Viper struct {
 	properties *properties.Properties
 
 	onConfigChange func(fsnotify.Event)
+	caseSensitive  bool
 }
 
 // New returns an initialized Viper instance.
@@ -219,6 +220,7 @@ func New() *Viper {
 	v.env = make(map[string]string)
 	v.aliases = make(map[string]string)
 	v.typeByDefValue = false
+	v.caseSensitive = false
 
 	return v
 }
@@ -275,6 +277,13 @@ var SupportedRemoteProviders = []string{"etcd", "consul"}
 func OnConfigChange(run func(in fsnotify.Event)) { v.OnConfigChange(run) }
 func (v *Viper) OnConfigChange(run func(in fsnotify.Event)) {
 	v.onConfigChange = run
+}
+
+func SetCaseSensitive(flag bool) {
+	v.SetCaseSensitive(flag)
+}
+func (v *Viper) SetCaseSensitive(flag bool) {
+	v.caseSensitive = flag
 }
 
 func WatchConfig() { v.WatchConfig() }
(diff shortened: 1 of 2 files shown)


viper.go
```diff
@@ -277,6 +279,13 @@ func (v *Viper) OnConfigChange(run func(in fsnotify.Event)) {
 	v.onConfigChange = run
 }
 
+func SetCaseSensitive(flag bool) {
```
sagikazarmark: Maybe add some comments here that will appear in godoc.

viper.go
```diff
@@ -277,6 +279,13 @@ func (v *Viper) OnConfigChange(run func(in fsnotify.Event)) {
...
+	v.SetCaseSensitive(flag)
+}
+func (v *Viper) SetCaseSensitive(flag bool) {
```
sagikazarmark: Maybe add some comments here that will appear in godoc.

viper.go
```diff
@@ -1827,6 +1850,14 @@ func (v *Viper) findConfigFile() (string, error) {
 	return "", ConfigFileNotFoundError{v.configName, fmt.Sprintf("%s", v.configPaths)}
 }
 
+func (v *Viper) properCase(s string) string {
```
sagikazarmark: Maybe call this `prepareKey` or something less tied to the casing.

viper.go
```diff
@@ -1572,6 +1591,10 @@ func (v *Viper) WatchRemoteConfigOnChannel() error {
 }
 
 func (v *Viper) insensitiviseMaps() {
+	if v.caseSensitive {
```
sagikazarmark: This looks unnecessary, at least I think you will never reach this check, since you do it before calling this method:

https://github.com/spf13/viper/pull/592/files#diff-38f1012180f40493dcf780c312a40ec3R1407

I think it would be better to do these checks outside of the `insensitive` methods.

viper.go
```diff
@@ -219,6 +220,7 @@ func New() *Viper {
 	v.env = make(map[string]string)
 	v.aliases = make(map[string]string)
 	v.typeByDefValue = false
+	v.caseSensitive = false
```
sagikazarmark: It will be false by default, no need to set it here IMHO.

kansz:
What time the feature can merged back to master branch? I need it

Issue #579, opened 2018-10-14
TestBindPFlagsStringSlice is broken on master


caalberts:
Recent PRs #573, #574, #575, #577 are failing on `TestBindPFlagsStringSlice`.

Does anyone have a clue how to fix this?

```
--- FAIL: TestBindPFlagsStringSlice (0.00s)
	viper_test.go:618: 
			Error Trace:	viper_test.go:618
			Error:      	Not equal: 
			            	expected: []string{}
			            	actual  : []string(nil)
			            	
			            	Diff:
			            	--- Expected
			            	+++ Actual
			            	@@ -1,3 +1,2 @@
			            	-([]string) {
			            	-}
			            	+([]string) <nil>
			            	 
			Test:       	TestBindPFlagsStringSlice
	viper_test.go:618: 
			Error Trace:	viper_test.go:618
			Error:      	Not equal: 
			            	expected: []string{}
			            	actual  : []string(nil)
			            	
			            	Diff:
			            	--- Expected
			            	+++ Actual
			            	@@ -1,3 +1,2 @@
			            	-([]string) {
			            	-}
			            	+([]string) <nil>
			            	 
			Test:       	TestBindPFlagsStringSlice
```

zhsj:
So I come across to this issue when I'm going to fix regression in mitchelh/mapstructure https://github.com/mitchellh/mapstructure/issues/146

I call this a regression in mitchelh/mapstructure since I thought the old behaviour was correct.

empty slice != nil slice.

I don't know viper too much, what is the expected result actually in TestBindPFlagsStringSlice? Whether empty or nil slice do you want when the input is `""`?

PR #596, opened 2018-11-15
Fix failing TestBindPFlagsStringSlice

toddy15:
After mitchelh/mapstructure has been updated, the expected test result needs to be updated to the actual returned result.

Closes: spf13/viper#579.


anthonyfok:
Thank you for your patch to viper!  The same fix was used as a part of commit  41cd1c3aa32b2685fae6c296e6f044bc68922c0e "Restrict Travis to >= Go 1.11, use Go Modules, and get the test to pass" (2018-12-05) and was released as viper 1.3.1.

PR #598, opened 2018-11-19
Fix file link tests on Travis

sagikazarmark:


sagikazarmark:
This PR is ready. Together with #596 the tests will finally stop breaking. Would be nice if in the future this would be a blocker for merging PRs 🙂

sagikazarmark:
Looks like this is not necessary after all.

PR #603, opened 2018-11-27
Adds support for merging slices

tlmiller:
- Work allows slices to be merged by appending of src onto target

It seems odd to not have this feature when the function is called merge. I could understand the reasoning if ordering of slices was important. If this is required could implement a merge strategy interface.

Thanks for checking out the PR

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -1510,6 +1510,12 @@ func mergeMaps(
 		case map[string]interface{}:
 			jww.TRACE.Printf("merging maps")
 			mergeMaps(sv.(map[string]interface{}), ttv, nil)
+		case []interface{}:
+			jww.TRACE.Printf("merging slices")
+			tgt[tk] = append(tgt[tk].([]interface{}), sv.([]interface{})...)
+			if itgt != nil {
+				itgt[tk] = append(itgt[tk].([]interface{}), sv.([]interface{})...)
+			}
 		default:
 			jww.TRACE.Printf("setting value")
 			tgt[tk] = sv
--- a/viper_test.go
+++ b/viper_test.go
@@ -1079,6 +1079,11 @@ hello:
     - uk
     - fr
     - de
+list:
+- one
+list2:
+- name: name1
+  value: value1
 `)
 
 var yamlMergeExampleSrc = []byte(`
@@ -1088,6 +1093,14 @@ hello:
     universe:
     - mw
     - ad
+list:
+- two
+- three
+list2:
+- name: name2
+  value: value2
+- name: name3
+  value: value3
 fu: bar
 `)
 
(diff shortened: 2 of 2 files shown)


tlmiller:
Thanks for the feedback @bep,

It's odd for me in that without any documentation on the merge strategy being use I personally expected slices to be appended together based on the order configuration files are being processed. I wasn't expecting to have non duplicating slice data just lost.

For our use case we have 1 or more config files that need to be read in where a root key is a list of objects describing endpoints that need to be scraped. Because of the way provisioning tools like Ansible are being used we need to define these in seperate files. [consul](https://www.consul.io/) is a good example of what I was expecting with regards to merging.

The sticking point I can see to this PR is lists are an ordered thing and may not be considered mergeable by some as different people may have different ideas on how this ordering should work. My personal opinion is that merging of lists should happened in append fashion based on the order the data is received (i.e files read).

As mentioned above the use of merge strategy types could be used for the user to explicitly decide the decision making being used and or write new decision makers.

Whats the additional information you think is needed when merging slices?

bep:
>Whats the additional information you think is needed when merging slices?

This is depends on the use case and cannot (in my head) be easily automated.

I maintain a project called Hugo. We use Viper and have a configuration that is a mix of maps of slices. 

* `languages` is an a slice. It has an identifier, so if Viper had known about this, it could maybe do a decent job of merging, but it doesn't -- and I don't think I would want that behaviour even if it did.
* `mediaTypes` is a map. Merging in new media type definitions from themes etc. makes sense, and this works great with Viper.

So, I think of slices as regular string/int values. I would be surprised if `myStringOption` was somehow merged, to take one absurd example.

So, this PR would tear down the house of cards I build, and will not be merged by me. You could probably do this by "some hooks", but as I said, this isn't something I need, so you need to take this with someone else. I was just a little bit curous about your "odd" reaction.

tlmiller:
Thanks again @bep,

I had a look at the Hugo config docs and I couldn't see where languages or mediaTypes where being configured as lists in the config. To me they both look like maps based on [here](https://gohugo.io/content-management/multilingual/#configure-languages) & [here](https://gohugo.io/templates/output-formats/#media-types).

In any event I think you are correct in that taking the PR as is would possibly break code for some people based on what they expect to happen. To this I am trying to feel the waters that if some sort of strategy type was introduced and could be set by the user where the default is to just do what happens now. Would that be acceptable?

The configuration example we have is we have a list of "providers" that need to be configured on the application, where the providers are generic and not known by the config system. So we need to be able to merge multiple file lists of providers then make decisions to find the correct provider mapper that can build that provider based on the config provided.

Hope that make sense.
Cheers

saplla:
I also need this for a similar reason - multiple config files containing different parts of a list I want appending together. I agree that doing this by default would probably be a surprise (so bad), but it'd be good if there was a way to have list values for e.g. named keys all appended when merging configs (i.e. so this behaviour was explicitly opt-in).

bep:
>I had a look at the Hugo config docs and I couldn't see where languages or mediaTypes where being configured as lists in the config. 

There are plenty of slices in Hugo. And if that wasn't the case, there are plenty of other apps in the wild ...

tlmiller:
Closing this PR in favor of #634

Issue #622, opened 2019-01-03
Init WaitGroup is never done in WatchConfig()


cocorambo:
No call to `initWG.Done()` is made in case getConfigFile return an error so WatchConfig() never return

See
https://github.com/spf13/viper/blob/6d33b5a963d922d182c91e8a1c88d81fd150cfd4/viper.go#L282-L296

sagikazarmark:
Fixed in #623

PR #623, opened 2019-01-03
Add missing call to initWG.Done() in WatchConfig()

cocorambo:
Fix issue https://github.com/spf13/viper/issues/622


sagikazarmark:
Thanks @cocorambo

PR #625, opened 2019-01-08
Fix indentation for yamlStringSettings

pawanrawal:
The indentation seemed messed up and didn't read well. This change fixes it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -630,11 +630,11 @@ import (
 
 func yamlStringSettings() string {
     c := viper.AllSettings()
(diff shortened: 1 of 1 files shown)


sagikazarmark:
It looks like the rest of the examples are wrong (they use spaces), but this way at least it will be consistent.

sagikazarmark:
@pawanrawal can you please rebase from the latest master?

pawanrawal:
Created #651 instead which should have the changes from latest master.

Issue #371, opened 2017-07-25
Does viper support uppercase in yaml map?


Kvaz1r:
MCVE:
Yaml:
```
Test :
 Var1 : Hello
 Var2 : world
```
Code:
```
package main

import (
	"fmt"
	"log"

	"github.com/spf13/viper"
)

func main() {

	viper.SetConfigType("yaml")
	viper.SetConfigName("input")
	viper.AddConfigPath(".")
	if err := viper.ReadInConfig(); err != nil {
		log.Fatal(err)
	}
	fmt.Println(viper.GetStringMap("Test"))
}
```

Output:

> map[var1:Hello var2:world]

Epected output:

> map[Var1:Hello Var2:world]

That's bug or I do something wrong?

benoitmasson:
Hi, this is the expected behavior, see #260 

In summary, viper configuration keys are case insensitive, and internally lowercased to enforce that.

What I do as a workaround when I really need case sensitivity of keys is the following:
```yaml
test:
    var1:
        contents: "Hello"
        key: "Var1"
    var2:
        contents: "world"
        key: "Var2"
```
but I agree this is not very convenient…

Kvaz1r:
Thanks, also will use your method.

Issue #131, opened 2015-11-09
Nested keys are case sensitive ?


meirwah:
Seems that when I use for the next Json file:
```json
"Bot":{
    "Name":"watson"
  }
```

the path "Bot.name" it does not work ...
but path "Bot.Name" works...

Should it be case insensitive on all levels??

ishidawataru:
Hi, I'm also hitting the same problem.

I think this can fix the issue. 
https://github.com/ishidawataru/viper/commit/79a5e3bd5bab855ce86e241556af9af9b445a34b

AkihiroSuda:
I'm also hitting the problem.

+1 @ishidawataru

ishidawataru:
This fixed the issue.

https://github.com/spf13/viper/commit/110492b3001269a0777997fcbe68a60828fac4cb

meirwah:
:+1:

sagikazarmark:
See #1014

PR #635, opened 2019-01-25
Add viper.SetKeysCaseSensitive() to disable automatic key lowercasing.

knadh:
YAML, TOML, and JSON dictate keys to be case-sensitive. Viper's default
behaviour of lowercasing the keys for key insensitivity is incompatible
with these standards and has the side effect of making it difficult for
use cases such as case sensitive API credentials in configuration.
For eg: MyApiKey=MySecret (in TOML).

See #131, #260, #293, #371, #373

This commit adds a global function `viper.SetKeyCaseSensitivity()` that
enables this behaviour to be turned off, after which, all keys, irrespective
of nesting, retain their cases. This respects all configuration operations
including getting, setting, and merging.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,7 +62,8 @@ item below it:
  * key/value store
  * default
 
-Viper configuration keys are case insensitive.
+Viper configuration keys are case insensitive by default. They can be made case
+sensitive with `viper.SetKeyCaseSensitivity(true)`.
 
 ## Putting Values into Viper
 
--- a/viper.go
+++ b/viper.go
@@ -188,6 +188,7 @@ type Viper struct {
 	automaticEnvApplied bool
 	envKeyReplacer      *strings.Replacer
 	allowEmptyEnv       bool
+	caseSensitiveKeys   bool
 
 	config         map[string]interface{}
 	override       map[string]interface{}
@@ -489,7 +490,6 @@ func (v *Viper) providerPathExists(p *defaultRemoteProvider) bool {
 
 // searchMap recursively searches for a value for path in source map.
 // Returns nil if not found.
-// Note: This assumes that the path entries and map keys are lower cased.
 func (v *Viper) searchMap(source map[string]interface{}, path []string) interface{} {
 	if len(path) == 0 {
 		return source
@@ -536,7 +536,7 @@ func (v *Viper) searchMapWithPathPrefixes(source map[string]interface{}, path []
 
 	// search for path prefixes, starting from the longest one
 	for i := len(path); i > 0; i-- {
-		prefixKey := strings.ToLower(strings.Join(path[0:i], v.keyDelim))
+		prefixKey := v.caseKey(strings.Join(path[0:i], v.keyDelim))
 
 		next, ok := source[prefixKey]
 		if ok {
@@ -657,7 +657,7 @@ func GetViper() *Viper {
 }
 
 // Get can retrieve any value given the key to use.
-// Get is case-insensitive for a key.
+// Get's case-sensitivity for key is determined by viper.keyCaseSensitivity.
 // Get has the behavior of returning the value associated with the first
 // place from where it is set. Viper will check in the following order:
 // override, flag, env, config file, key/value store, default
@@ -665,8 +665,8 @@ func GetViper() *Viper {
 // Get returns an interface. For a specific value use one of the Get____ methods.
 func Get(key string) interface{} { return v.Get(key) }
 func (v *Viper) Get(key string) interface{} {
-	lcaseKey := strings.ToLower(key)
-	val := v.find(lcaseKey)
+	casedKey := v.caseKey(key)
+	val := v.find(casedKey)
 	if val == nil {
 		return nil
 	}
@@ -674,7 +674,7 @@ func (v *Viper) Get(key string) interface{} {
 	if v.typeByDefValue {
 		// TODO(bep) this branch isn't covered by a single test.
 		valType := val
-		path := strings.Split(lcaseKey, v.keyDelim)
+		path := strings.Split(casedKey, v.keyDelim)
 		defVal := v.searchMap(v.defaults, path)
 		if defVal != nil {
 			valType = defVal
@@ -704,7 +704,7 @@ func (v *Viper) Get(key string) interface{} {
 }
 
 // Sub returns new Viper instance representing a sub tree of this instance.
-// Sub is case-insensitive for a key.
+// Sub's case-sensitivity for key is determined by viper.keyCaseSensitivity.
 func Sub(key string) *Viper { return v.Sub(key) }
 func (v *Viper) Sub(key string) *Viper {
 	subv := New()
@@ -811,7 +811,9 @@ func (v *Viper) UnmarshalKey(key string, rawVal interface{}, opts ...DecoderConf
 		return err
 	}
 
-	v.insensitiviseMaps()
+	if !v.caseSensitiveKeys {
+		v.insensitiviseMaps()
+	}
 
 	return nil
 }
@@ -828,7 +830,9 @@ func (v *Viper) Unmarshal(rawVal interface{}, opts ...DecoderConfigOption) error
 		return err
 	}
 
-	v.insensitiviseMaps()
+	if !v.caseSensitiveKeys {
+		v.insensitiviseMaps()
+	}
 
 	return nil
 }
@@ -872,8 +876,9 @@ func (v *Viper) UnmarshalExact(rawVal interface{}) error {
 		return err
 	}
 
-	v.insensitiviseMaps()
-
+	if !v.caseSensitiveKeys {
+		v.insensitiviseMaps()
+	}
 	return nil
 }
 
@@ -918,7 +923,7 @@ func (v *Viper) BindFlagValue(key string, flag FlagValue) error {
 	if flag == nil {
 		return fmt.Errorf("flag for %q is nil", key)
 	}
-	v.pflags[strings.ToLower(key)] = flag
+	v.pflags[v.caseKey(key)] = flag
 	return nil
 }
 
@@ -933,7 +938,7 @@ func (v *Viper) BindEnv(input ...string) error {
 		return fmt.Errorf("BindEnv missing key to bind to")
 	}
 
-	key = strings.ToLower(input[0])
+	key = v.caseKey(input[0])
 
 	if len(input) == 1 {
 		envkey = v.mergeWithEnvPrefix(key)
@@ -950,13 +955,14 @@ func (v *Viper) BindEnv(input ...string) error {
 // Viper will check in the following order:
 // flag, env, config file, key/value store, default.
 // Viper will check to see if an alias exists first.
-// Note: this assumes a lower-cased key given.
-func (v *Viper) find(lcaseKey string) interface{} {
+// Note: By default, this assumes that a lowercase key is given.
+// This behavior can be modified with viper.SetKeysCaseSensitive.
+func (v *Viper) find(key string) interface{} {
 
 	var (
 		val    interface{}
 		exists bool
-		path   = strings.Split(lcaseKey, v.keyDelim)
+		path   = strings.Split(key, v.keyDelim)
 		nested = len(path) > 1
 	)
 
@@ -966,8 +972,8 @@ func (v *Viper) find(lcaseKey string) interface{} {
 	}
 
 	// if the requested key is an alias, then return the proper key
-	lcaseKey = v.realKey(lcaseKey)
-	path = strings.Split(lcaseKey, v.keyDelim)
+	key = v.realKey(key)
+	path = strings.Split(key, v.keyDelim)
 	nested = len(path) > 1
 
 	// Set() override first
@@ -980,7 +986,7 @@ func (v *Viper) find(lcaseKey string) interface{} {
 	}
 
 	// PFlag override next
-	flag, exists := v.pflags[lcaseKey]
+	flag, exists := v.pflags[key]
 	if exists && flag.HasChanged() {
 		switch flag.ValueType() {
 		case "int", "int8", "int16", "int32", "int64":
@@ -1004,14 +1010,14 @@ func (v *Viper) find(lcaseKey string) interface{} {
 	if v.automaticEnvApplied {
 		// even if it hasn't been registered, if automaticEnv is used,
 		// check any Get request
-		if val, ok := v.getEnv(v.mergeWithEnvPrefix(lcaseKey)); ok {
+		if val, ok := v.getEnv(v.mergeWithEnvPrefix(key)); ok {
 			return val
 		}
 		if nested && v.isPathShadowedInAutoEnv(path) != "" {
 			return nil
 		}
 	}
-	envkey, exists := v.env[lcaseKey]
+	envkey, exists := v.env[key]
 	if exists {
 		if val, ok := v.getEnv(envkey); ok {
 			return val
@@ -1050,7 +1056,7 @@ func (v *Viper) find(lcaseKey string) interface{} {
 
 	// last chance: if no other value is returned and a flag does exist for the value,
 	// get the flag's value even if the flag's value has not changed
-	if flag, exists := v.pflags[lcaseKey]; exists {
+	if flag, exists := v.pflags[key]; exists {
 		switch flag.ValueType() {
 		case "int", "int8", "int16", "int32", "int64":
 			return cast.ToInt(flag.ValueString())
@@ -1080,11 +1086,12 @@ func readAsCSV(val string) ([]string, error) {
 }
 
 // IsSet checks to see if the key has been set in any of the data locations.
-// IsSet is case-insensitive for a key.
+// IsSet is case-insensitive for a key. This behavior can be modified
+// with viper.SetKeysCaseSensitive.
 func IsSet(key string) bool { return v.IsSet(key) }
 func (v *Viper) IsSet(key string) bool {
-	lcaseKey := strings.ToLower(key)
-	val := v.find(lcaseKey)
+	casedKey := v.caseKey(key)
+	val := v.find(casedKey)
 	return val != nil
 }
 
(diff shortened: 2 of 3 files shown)


knadh:
@sagikazarmark The existing tests and the new tests that were added with this commit all pass locally. Travis is failing at a certain `go get`. Not sure what's happening there.

sagikazarmark:
@knadh it seems to be a go modules issue. I sent a PR fixing it: #636

Once it's merged, you can just rebase your branch.

v1k0d3n:
we really need this for our use case (exactly what you mentioned about case sensitive API creds). it would be awesome to see this merged in. 🤞

viper.go
```diff
@@ -1162,34 +1168,38 @@ func (v *Viper) InConfig(key string) bool {
 }
 
 // SetDefault sets the default value for this key.
-// SetDefault is case-insensitive for a key.
+// SetDefault's case-sensitivity for key is determined by viper.keyCaseSensitivity.
```
sagikazarmark: I would retain the original comment and add something like this:
`This behavior can be modified by viper.SetKeyCaseSensitivity`

viper.go
```diff
@@ -1162,34 +1168,38 @@ func (v *Viper) InConfig(key string) bool {
...
+	lastKey := v.caseKey(path[len(path)-1])
 	deepestMap := deepSearch(v.defaults, path[0:len(path)-1])
 
 	// set innermost value
 	deepestMap[lastKey] = value
 }
 
 // Set sets the value for the key in the override register.
-// Set is case-insensitive for a key.
+// Set's case-sensitivity for a key is determined by viper.keyCaseSensitivity.
```
sagikazarmark: I would retain the original comment and add something like this:
`This behavior can be modified by viper.SetKeyCaseSensitivity`

viper.go
```diff
@@ -1778,6 +1792,14 @@ func (v *Viper) SetConfigType(in string) {
 	}
 }
 
+// SetKeyCaseSensitivity disables the default behaviour of
+// case-insensitivising keys and preserves the key casing
+// as they are found in the config files.
+func SetKeyCaseSensitivity(on bool) { v.SetKeyCaseSensitivity(on) }
```
sagikazarmark: Again, case sensitivity is not boolean IMHO. I would rather call this something like `SetKeysCaseSensitive` or `CaseSensitiveKeys`.

viper.go
```diff
@@ -1778,6 +1792,14 @@ func (v *Viper) SetConfigType(in string) {
...
+func (v *Viper) SetKeyCaseSensitivity(on bool) {
```
sagikazarmark: Same as above

sagikazarmark:
@knadh the mentioned PR is merged.

Also, I found this: #592 it seems to be a duplicate, it's probably worth checking.

v1k0d3n:
I was thinking the same when I saw this PR @sagikazarmark, but it appears that #592 just seemed to fall flat on the floor for some reason (reviews weren't addressed). Hopefully, we can see this PR through to the end.

sagikazarmark:
Yeah, I just mentioned it because I think it still makes sense to compare the two solutions, but ultimately I think this one should be merged eventually.

knadh:
Thanks for the merge, the tests now pass. I've incorporated the changes you posted.

#592 follows the same approach, a separate that function that handles key casing. Just that it's incomplete--no tests and no semantic changes to comments or local variables.

v1k0d3n:
Awesome! Would love to see this get in.

knadh:
@bep Have you had a chance to review the PR? Thanks.

knadh:
@bep It'd be great if you could take a look at the PR. Thanks.

v1k0d3n:
would a cute animal gif help?

[donkey](https://media.tenor.com/images/41d5b948ed7a6c795359117031ee177e/tenor.gif)

mr-karan:
@bep Thanks for your work. Would be really awesome if this PR is reviewed. This feature could really help us, as well :+1:

terev:
@spf13 Can we get a review on this? This a pretty crucial option.

bep:
I don't have time to review this at the moment. Maybe @spf13 can chime in.

prymitive:
What is status of this project? Is viper still actively maintained? Can we assume this will get merged in the near future?

alexandre-deceneux:
It seems that a new commit has been merged that removes a bunch of `insensitiviseMaps` method calls.
It should unlock most of the cases.

https://github.com/spf13/viper/commit/9e56dacc08fbbf8c9ee2dbc717553c758ce42bc9

knadh:
I've updated the PR to incorporate https://github.com/spf13/viper/commit/9e56dacc08fbbf8c9ee2dbc717553c758ce42bc9.

v1k0d3n:
ok, how many animal pics do we need? i can do this all day long folks...

babygoats)

v1k0d3n:
Ok, so we’re at a crossroads with this PR. I see this as the submitter has done everything to keep this PR fresh, several other folks have expressed interest/need for this PR, and reasons for not pulling it in seem a little trivial or can be addressed rather easily.

Based on this, we need to either find another solution (if the viper maintainers have an extreme objection to this PR), or fork (which we’d rather not do). Can the maintainers give us some feedback, please? Right now, this seems like it’s in limbo, and I’m kind of at a loss. I’ve even reached out to a few folks on gopher slack, in good faith, but I haven’t gotten any response.

Since this has been out there for a couple of months now; what are we recommending here? Anyone who’s following this, and requesting it as well, what are your thoughts or how are you thinking of addressing this? Can we discuss openly what our options are here?

Thanks for any feedback and consideration. I’m getting concerned with all of the outstanding PR’s. We can help, but not if there’s much feedback or any additional requests from the maintainers.

demond2:
As of yesterday we are using a forked Viper with a **replace** in go.mod for our base yaml handling package under the assumption this PR will eventually get merged. If this makes it through our pipeline it will apply to all of our products. Not ideal, and it could easily push our development into a corner if this does not get resolved.
It's feasible we could end up in a situation where the fork becomes permanent and start work on a V2, but I imagine if that happens ours will be one of a dozen active hard forks.

sagikazarmark:
I will try to review the PR during the weekend, maybe that helps the maintainers.

sagikazarmark:
After going through the code, I have one main concern (which I have expressed earlier):

> My concern regarding this change is that it introduces a strange side effect: what if I set viper to case sensitive mode, add a few values and set it back to case insensitive mode? I think with the current implementation it won't recognize the added keys (I might be wrong, I quickly went through the code).

I still think this is a problem, but I also have a solution: instead of using a setter, let's create a new constructor which accepts functional options. Having one constructor would be better, but adding a variadic argument to the current New function is technically a backward incompatible change. Maybe in a v2 version we can merge the two constructors.

Maybe the added comments and documentations could be worded a bit better (and most of them would need modification because of the setter->constructor option change)

Other than these issues, I think this PR is great and implements a hard problem nicely. Good job!

knadh:
Thanks @sagikazarmark. I'd thought of the constructor approach initially, but looking at some of our own projects where initialisations are done entirely using the package level functions, I realised they'd require some level of refactoring to accommodate the constructor based approach. I imagine this would be the case with most projects using Viper.

Since the setter function has to be invoked explicitly, and only implementations seeking this specific behaviour would use it, I feel explicit documentation should suffice. It does feel a little unsemantic, but probably not as much as losing case information by default.

sagikazarmark:
Well, we will have to agree to disagree. 🙂 But that's fine.

Honestly, I'm not a huge fan of the global Viper instance in the first place, but I think there are better alternatives for re initializing the global instance (using an alternative constructor) than setting options with side-effects on it.

For example a global `SetViper` function sounds better. Easier to explain and understand.

knadh:
Could you elaborate on the global `SetViper` approach?

sagikazarmark:
I meant something like this:

`viper.SetViper(viper.NewWithOptions(viper.CaseSensitiveKeys()))`

I can imagine it called: `Reinitialize` or something similar as well.

knadh:
Ah, got it. That still does not help with the side effects as re-initialization will have to throw away the global instance, right? Unclear how a full reinitialization with new options helps over setting one option on an already initialized instance.

sagikazarmark:
> Unclear how a full reinitialization with new options helps over setting one option on an already initialized instance.

Your problem is making the global instance case sensitive, but the current solution has side effects in both the global and custom instances. A global setter does not remove the side effect from the global instance, it does remove it from custom instances.

KISSMonX:
nice

knadh:
Looks like this is never going to see the light of the day :)

demond2:
Can we get a consensus on the changes needed at this point?

hieman:
I also wait to get this change merged, but I think the viper.Sub methode should init the case sensitivity as set for the based viper instance. Currently viper.New() doesn't set this value. Maybe this could be reasonable for new instances, but should not be the case for Sub. 
To avoid inconsitent settings for case sensivity, the setting could be rejected if there is already read config content.

demond2:
Not that my voice counts for much, but I agree with @hieman with one clarification.
viper.New() with functional option and Sub inheriting the setting would seem to get us what we need while being safe.

mdone-cloudian:
> what if I set viper to case sensitive mode, add a few values and set it back to case insensitive mode?

What if you just document that you should call `viper.SetKeyCaseSensitivity()` only once (per project?) before doing anything with keys?

ToJen:
@spf13 is this okay to merge?

README.md
```diff
@@ -62,7 +62,8 @@ item below it:
  * key/value store
  * default
 
-Viper configuration keys are case insensitive.
+Viper configuration keys are case insensitive by default. They can be made case
+sensitive with `viper.SetKeyCaseSensitivity(true)`.
```
gbunt: I think this should be: `viper.SetKeysCaseSensitive(true)`

gbunt:
fwiw, the changes in this PR seem to work nicely with the app we're developing which contains a relatively complex data structure with camelCase parameter names, which are now being retained. Thanks so far @knadh. Hope this gets merged in soon so we can avoid the additional steps of merging this PR at build time.

hieman:
As I mentioned before the viper.Sub should be changed like this

`func (v *Viper) Sub(key string) *Viper {
	subv := New()
	subv.SetKeysCaseSensitive()	
...
}`

to ensure that the contained part still uses the orginated case sensitivity.

knadh:
We've moved away from viper because of this and a [few other fundamental issues](https://github.com/knadh/koanf#alternative-to-viper). I ended up writing [github.com/knadh/koanf](https://github.com/knadh/koanf). Although it doesn't do everything viper does, most common usecases are covered, and we've successfully integrated it into several of our projects.

I would not be maintaining this PR anymore. Thank you everyone for participating.

demond2:
> We've moved away from viper because of this and a [few other fundamental issues](https://github.com/knadh/koanf#alternative-to-viper). I ended up writing [github.com/knadh/koanf](https://github.com/knadh/koanf). Although it doesn't do everything viper does, most common usecases are covered, and we've successfully integrated it into several of our projects.
> 
> I would not be maintaining this PR anymore. Thank you everyone for participating.

Myself and others in this thread are still keenly interested in seeing this merged.

@knadh Would you (or anyone) have any objections to either myself or someone else grabbing the reins on this one?

knadh:
@demond2 Please feel free.

v1k0d3n:
@knadh (and whom else this concerns): @demond2 passed away a week ago and is unable to continue this work. he forked the repo, which you and others are welcome to leverage in the time being.

it's been 6 months since this was originally submitted, with little feedback. it doesn't seem to garner the support from the maintainers. as our team has learned, a _lot_ can happen in 6 months.

is there any way we can please get this merged, or at least get final nay from the maintainers? it seems like a sensible addition to me and the growing list of others in this thread...

knadh:
Wow, terribly sorry to hear that ;(

If there's no answer still, I think it may just be better for the PR to be closed so that people don't end up waiting indefinitely.

hieman:
Sorry @knadh you are right it is terrible that there is still no commitment for this PR, but I don't agree that closing is an option. As case sensitity is a common use case for most of the supported file formats, it is without a doubt a resonable request, which has to be solved and I think your solution, with my minor requested add ons, fix the problem very well.   
So I totally agree with @v1k0d3n, please get it merged....

lnashier:
is #635 still required. I think #666 fixes this issue

gbunt:
> is #635 still required. I think #666 fixes this issue

#666 does _not_ result in keys retaining their case (luckily, for folks depending on auto-lowercasing). #635 is definitely still required (and wanted by many by the looks of it, myself included)

gbunt:
#743 is removing lowercasing as well, but without it being configurable in any way so would become the default then i guess.

knadh:
I think #666 breaks backwards compatibility by completely removing the automatic lowercasing behaviour users would've built on. It's bound to silently break in production.

sagikazarmark:
@knadh so I've hit a wall with Viper and key insensitivity, so I will kinda have to pick this issue up. 😄

knadh:
@sagikazarmark It was just a matter of time!

Maybe you'd want to give https://github.com/knadh/koanf a shot ^_^.

sagikazarmark:
@knadh although it looks pretty cool, we are kinda locked on Viper. It would be great if we could rework Viper from the basics, because it's already widely adopted.

Personally I hate to see that the number of config and logging libraries is still growing. 😢

sagikazarmark:
@knadh I would probably finish your PR. The only thing I would change is adding a factory function that allows setting the sensitivity at construction time, so that users cannot accidentally call a setter.

The rest should just work, right? 😄

knadh:
@sagikazarmark Yep, that should work. However, I'm not sure of https://github.com/spf13/viper/pull/666.

sagikazarmark:
@knadh what do you mean?

knadh:
@sagikazarmark #666 removes `insensitiviseMaps()`. I cannot recollect how that interacts with key cases. Not sure if it affects the PR.

hieman:
@sagikazarmark you should add my earlier mentioned change for viper.Sub, otherwise it could result in inconstant content.

sagikazarmark:
@hieman Sure, thanks for the reminder.

vitalyisaev2:
Hi guys, are there any updates? We need this feature as well.

prashantthakre:
Any idea as to when will the fix go in?

stephencheng:
+1 Any idea as to when will the fix go in??

knadh:
@stephencheng I opened this PR 1 year and 1 month ago ^_^ and waited for 6 months for a merge before ditching Viper for good and writing [koanf](https://github.com/knadh/koanf) :smile:. You can give it a shot. Addresses this, and several other issues.

[image](https://github.com/knadh/koanf)

sagikazarmark:
Unfortunately this change affects Viper at its core and we have to make sure it's aligned with the future of Viper and does not break existing behavior. Believe me, I've given this PR a considerable amount of time to find a way to integrate it and meet the above requirements.

prymitive:
If this PR is not going to get merged then could it be declined, so everyone gets clarity on the situation here?
It's perfectly fine if viper mainteners say they won't support case sensitive keys, especially that there are alternative solutions, it's just that both the issue and the PR being open for so long puts the whole problem in a limbo.

sagikazarmark:
We plan to add support for case sensitivity. Whether it will go in a Viper 2 or can be integrated into Viper 1 is a question though. If you need this feature now though, Viper is probably not the best choice at the moment and probably isn't worth waiting for it. That's the best answer I have right now.

vividvilla:
@sagikazarmark PR doesn't seems to be affecting exisiting behavior though as you mentioned above. Was there any edge case missed?

gbunt:
fwiw i've merged the changes from @knadh into a fork here: https://github.com/gbunt/viper as i had to also pin mapstructure to v1.0.0. Any mapstructure version above that and i ran into issues with validator.v9 not handling function `omitempty` correctly on lists and thus, we weren't able to use validation + empty lists. Still have to debug what the issue is exactly but for now, it's working quite nicely for us. Glad to see support for case sensitivity is seriously being looked at 👍

gauravv7:
This PR's case is very sensitive

wfernandes:
It seems that this is the discussion thread for case sensitivity. Could we add this issue to the README? 
> Important: Viper configuration keys are case insensitive. There are ongoing discussions about making that optional.

gauravv7:
> It seems that this is the discussion thread for case sensitivity. Could we add this issue to the README?
> 
> > Important: Viper configuration keys are case insensitive. There are ongoing discussions about making that optional.

I am trying to make a sequence engine which uses json config files, to pass query filters as user-defined configs e.g. mongo query filters. Now the keys are fieldName of my documents. Mongodb is just one supported db, there could be many likewise and other algorithmic supports.
Please see if this can count as a factor towards approving this PR? let me know, kinda will help me understand the global standards as well and though process :)
-with a good heart

hypergig:
If you merge this in today... I will literally drop 100 bucks in any donation / paypal / venmo account you tell me to

gauravv7:
covid vaccine is nearly out for production use

knadh:
> If you merge this in today..

Too old and stale to merge at this point @hypergig. The maintainers have [clarified this](https://github.com/spf13/viper/pull/635#issuecomment-580659676) on this thread that it's not happening.

hypergig:
> > If you merge this in today..
> 
> Too old and stale to merge at this point @hypergig. The maintainers have [clarified this](https://github.com/spf13/viper/pull/635#issuecomment-580659676) on this thread that it's not happening.

:cry: :panda_face: 

Maybe we can close the PR so it doesn't continue to provide false hope?

knadh:
> Maybe we can close the PR so it doesn't continue to provide false hope? 

Yeah, I'd left it open hoping it'll get merged at some point, but makes sense to close now.

cocktail828:
oops! It's 2023 now, and the issue still not fixed...

PR #636, opened 2019-01-25
Update go.sum with go1.11.4

sagikazarmark:
In Go 1.11.4 a bug was fixed related to checksum calculation.
As a result, some packages might end up with a different checksum
from this version. The solution is upgrading, cleaning the mod cache
and recalculating the go.sum file.

See https://github.com/golang/go/issues/27093


sagikazarmark:
Thanks @bep

PR #638, opened 2019-01-27
Add INI support

### LINKED ISSUE #194: INI support through gcfg
I'm interested in using https://godoc.org/gopkg.in/gcfg.v1 to support INI files in viper, eventually to resolve https://github.com/s3git/s3git/issues/15 and possibly others. It reads the key-value pairs into a two-level struct: https://godoc.org/gopkg.in/gcfg.v1#ex-ReadStringInto--Subsections

Could you guide me through the steps necessary to implement that new configuration file backend for Viper? I want to make sure I implement it in a way that's acceptable.

techknowlogick:
Fixes  #194

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -46,6 +46,7 @@ import (
 	jww "github.com/spf13/jwalterweatherman"
 	"github.com/spf13/pflag"
 	"github.com/subosito/gotenv"
+	"gopkg.in/ini.v1"
 )
 
 // ConfigMarshalError happens when failing to marshal the configuration.
@@ -231,7 +232,7 @@ func New() *Viper {
 // can use it in their testing as well.
 func Reset() {
 	v = New()
-	SupportedExts = []string{"json", "toml", "yaml", "yml", "properties", "props", "prop", "hcl", "dotenv", "env"}
+	SupportedExts = []string{"json", "toml", "yaml", "yml", "properties", "props", "prop", "hcl", "dotenv", "env", "ini"}
 	SupportedRemoteProviders = []string{"etcd", "consul"}
 }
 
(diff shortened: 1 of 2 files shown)


spf13:
I personally just really dislike the ini format. Is there a reason you need it?

techknowlogick:
@spf13 I too am not a fan of the format either, which is exactly the reason for this PR. I am a lead on https://github.com/go-gitea/gitea/ which for historical reasons uses ini, however if this PR is accepted, then the project has a roadmap in place to switch to using another configuration format, but in a backwards compatible way.

sagikazarmark:
Personally I'd rather see a way to extend the current set of parsers (something like drivers in the *database/sql* package)

earlzo:
#194

earlzo:
hey @spf13, are you gonna to merge this mr?

techknowlogick:
@sagikazarmark I'd like to see extensions in that way too, perhaps that should is an item for 2.0.

As for now, I've resolved merge conflicts and updated tests so that INI can be understood in 1.x version of viper.

sagikazarmark:
@techknowlogick TBH I'm not so keen on adding INI support. Actually I would be open to adding it temporarily, because right now there is no way to extend current encoders. But what then? Can we remove it in v2?

Let me think about the extendable encoder solution. If it's feasible to implement in a timely manner, would it be acceptable for you to use that new API to register your custom encoder in your own code?

techknowlogick:
I would be completely open to that solution, as that is preferable for me (I too dislike INI, but sadly it is a format that I have to work with, and having viper be able to support INI in some fashion would allow me to migrate away from INI).

Also, I think moving between major versions allows for backwards incompatible changes, so if this gets merged now, then INI doesn't have to be in v2 (although hopefully an extendable encoder solution would be available).

techknowlogick:
@sagikazarmark following up on this, have you discussed with other maintainers re: extending v1 with encoders?

sagikazarmark:
@techknowlogick no, unfortunately I didn't have the chance. I started playing with an encoding abstraction, but hit the wall with a few problems unfortunately.

I don't want to block this PR, so if you find some time to rebase the branch I'll review and merge it.

techknowlogick:
@sagikazarmark thanks for the update :) I've just merged the most recent commits to master into this PR. Please let me know if you'd like me to squash this branch.

sagikazarmark:
@techknowlogick Please do. Also, the branch is still conflicting with master. So I would suggest rebasing instead of merging.

techknowlogick:
Squashed & rebased. Looks like Travis is also passing too. 🎉 pleade let me know if there is anything else I can do :)

Issue #33, opened 2015-01-09
Config export


krzysztofantczak:
Hi,

How can i export loaded config (with ie. override some parts of it using ENV or CLI args) back to original or even different config file?

cytec:
+1 option for writing config back to a file would be great

cytec:
Nevermind, i've just wrote a simple function which saves everything to a file:

    type config struct {
        DataDir      string
        Debug        bool
        Devel        bool
        ApiKey       string
        SnatchFolder string
        Port         int
        TvdbUrl      string
        TvdbKey      string
        LogJSON      bool
        FirstRun     bool
    }
    
    var C config
    
    func saveConfig(cfgpath string) error {
    
        viper.Marshal(&C)
    
        b, err := json.MarshalIndent(C, "", "    ")
        if err != nil {
            return err
        }
    
        f, err := os.Create(cfgpath)
        if err != nil {
            return err
        }
    
        defer f.Close()
    
        f.WriteString(string(b))
    
        return nil
    }

theherk:
I feel like this would be a huge gain for viper. I would love to be able to modify the configuration and save it back. Viper already knows which file type was loaded and how to format that language.

theherk:
Needing this feature, too, I have opened PR #287. Once complete, writing will be available.

patrick-motard:
@theherk @spf13  It looks like this issue can be closed now that #287 is complete. Although it would be nice if the functionality was documented. I didn't see anything about it in the README.md.

theherk:
Excellent point. I'll make sure the documentation for this feature set is ship shape this week.

patrick-motard:
@theherk Thanks for the work you did on this feature. It saved me time writing my own!!

koliyo:
Great to see this functionality available. But it is really missing from the main README.md!

f1mishutka:
There is no viper.Marshal() method at https://github.com/spf13/viper/blob/master/viper.go

What am I doing wrong? :)

mattouille:
> There is no viper.Marshal() method at https://github.com/spf13/viper/blob/master/viper.go
> 
> What am I doing wrong? :)

I believe it's actually supposed to be `viper.Unmarshal()`. The idea being that you save everything to the struct (Unmarshal is a good way of getting the "final" values if you have multiple sources) and then writing it all to JSON.

cytec:
[...]

PR #649, opened 2019-02-15
Add WriteConfig documentation

EmilLuta:
After running through the readme, I needed the functionality. After digging, seems it's been there since 2k16, but no docs were provided. Decided it might help others as well.

This should also help to close https://github.com/spf13/viper/issues/33.

@spf13 Looking forward to feedback.


felicianotech:
This is useful.

Bump.

Issue #480, opened 2018-04-06
writeConfig Always Writes to 644 File


kristinaspring:
Currently, any function that uses the v.writeConfig function to write to a file does not have a way to specify the permissions for the file.  Here is the line of code in question:
https://github.com/spf13/viper/blob/8dc2790b029dc41e2b8ff772c63c26adbb1db70d/viper.go#L1259

Because of the properties being saved in the file, I want it to have 0600 permissions.  Currently, my solution is to use os.chmod after writing the configuration to a file, but it would be nice to be able to specify the file permissions.

PR #652, opened 2019-02-22
Add API for setting file permissions

krousey:
This allows users to set the file permissions that Viper uses for the config file.

fixes: #480


krousey:
@spf13 @bep Is there a process I need to follow to get this considered/reviewed?

PR #663, opened 2019-03-12
WIP: Remove superflous insensitiviseMaps in Unmarshal methods

### LINKED ISSUE #482: UnmarshalKey function is not safe for concurrent access.
Config json
```
{
   "someKey": 1,
   "options": {
       "x": "v1",
       "y": "v2"
   }
}
```

Some struct:
```
type Options struct {
  X string `json:"x"`
  Y string `json:"y"`
}
```

Sudo code:
```
config := viper.New()
config.AddConfigPath("myconfigpath")
config.SetConfigName("myconfig")

go func() {
 for {
    config.getString("someKey")
 }
}()

go func() {
 for {
   var opt Options
   config.UnmarshalKey("options", &opt)
 }
}()
```

Fatal error:
```
fatal error: concurrent map read and map write
```

This error happens due to https://github.com/spf13/viper/blob/master/util.go#L84
I would like to ask purpose of normalizing the keys here. 

Normalization of keys is also the reason for #373

bep:
Fixes #482

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -811,8 +811,6 @@ func (v *Viper) UnmarshalKey(key string, rawVal interface{}, opts ...DecoderConf
 		return err
 	}
 
-	v.insensitiviseMaps()
-
 	return nil
 }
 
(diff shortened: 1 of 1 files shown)


bep:
I need to have a closer look at this later. This code looks odd.

Issue #324, opened 2017-03-30
Error ``Unmarshal`` map when key contains dot


StephenPCG:
Test code:

```
package main

import (
	"bytes"
	"fmt"
	"github.com/spf13/viper"
)

type Config struct {
	Foo struct {
		Bar map[string]int `mapstructure:"bar"`
	} `mapstructure:"foo"`
}

var content = []byte(`
foo:
  bar:
    "x": 1
    "y": 2
    "z.z": 3
`)

func main() {
	var c1, c2 Config

	viper.SetConfigType("yaml")
	viper.ReadConfig(bytes.NewReader(content))

	viper.Unmarshal(&c1)
	fmt.Println("c1:", c1)

	viper.UnmarshalKey("foo", &c2.Foo)
	fmt.Println("c2:", c2)
}
```

Output:
```
c1: {{map[x:1 y:2]}}
c2: {{map[x:1 y:2 z.z:3]}}
```

As you can see, when the map key contains dot, ``UnmarshalKey`` works correctly but ``Unmarshal`` doesn't.

dlouwers:
This still seems to be an issue for me. This test fails depending on v1.3.1:
```go
package main

import (
	"strings"
	"testing"

	"github.com/spf13/viper"

	"github.com/stretchr/testify/assert"
	"github.com/stretchr/testify/require"
)

func TestConfigMarshalling(t *testing.T) {
	data := `
root:
  ele.one: value
  eletwo: value`
	reader := strings.NewReader(data)
	viper.SetConfigType("yaml")
	err := viper.ReadConfig(reader)
	require.NoError(t, err)

	// Test if the fields are even correct
	expected := map[string]interface{}{"ele.one": "value", "eletwo": "value"}
	assert.Equal(t, expected, viper.Get("root"))
	var s struct {
		Root map[string]string
	}
	err = viper.UnmarshalKey("root", &s)
	require.NoError(t, err)
}
```
Changing the dot to, say, an underscore fixes the issue.

alexeldeib:
Also hit this. Seems like if there's a key like "word.other" containing a dot as a map[string]interface{} key, the output turns it into a struct like: 

```
{
    "word": {
        "other": "value"      
    }
}
```

and not
```
{
    "word.other": "value"
}
```

Is that expected? Anyone have pointers to override this to get the former behavior?

inkychris:
Looks like `AllKeys()` returns a list of all the keys it can find as delimited strings:
```golang
// main.go
[...]

PR #667, opened 2019-03-13
Add setters for key delimiter

inkychris:
As discussed in #324, the use of the dot delimiter for keys can cause issues when keys themselves also contain dots. To get around this issue, I propose we add setters that can be used to change the delimiter for cases when it is required that dots are allowed in key values.


mschneider82:
Important feature if you have a domainname as a key.

inkychris:
Superseded by a native solution (#673) that doesn't change the public API or require explicit attention from user.

PR #669, opened 2019-03-16
Fix tests that use temp dirs on Windows

inkychris:
Fixes #668.

Tests added in #440 called `os.Clearenv()` which on windows causes `os.TempDir()` to default to `C:\Windows` and cause subsequent file permission failures. This has been repaced with `os.Unsetenv()` for the variables used in those tests.

The `TestDirsSearch` test was using `path.Base()` in its assertion which did not account for backslashes used on Windows. This has been replaced with `filepath.Base()` which correctly returns the expected value.

The `initDirs()` method now terminates a test if the temporary root test directory does not exist which should help debug tests that use it if `os.Clearenv()` gets used in future tests.

PR #673, opened 2019-03-23
Unmarshal keys containing dots

### LINKED ISSUE #324: Error ``Unmarshal`` map when key contains dot

inkychris:
Fixes #324 

Includes fix #669

- Replaced the use of `strings.Split` with a method that takes a fully qualified key and returns a list of valid map keys which may include dots in them.

- Reformatted test objects to improve readability and Git change diffing.

- Added key containing dot to yaml test.

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -1789,6 +1789,24 @@ outer:
 	return shadow
 }
 
+// Converts a fully qualified map key into a list of relative
+// map keys, allowing for keys to contain the delimiter themselves
+func keyComponents(v *Viper, key string) []string {
+	var result []string
+	components := strings.Split(key, v.keyDelim)
+	for index := 0; index < len(components); index++ {
+		potentialKey := strings.Join(components[0:index], v.keyDelim)
+		if v.Get(potentialKey) != nil {
+			result = append(result, potentialKey)
+		}
+	}
+	result = append(result, key)
+	for i := len(result) - 1; i > 0; i-- {
+		result[i] = strings.Replace(result[i], result[i-1]+v.keyDelim, "", 1)
+	}
+	return result
+}
+
 // AllSettings merges all settings and returns them as a map[string]interface{}.
 func AllSettings() map[string]interface{} { return v.AllSettings() }
 func (v *Viper) AllSettings() map[string]interface{} {
@@ -1801,7 +1819,7 @@ func (v *Viper) AllSettings() map[string]interface{} {
 			// check just in case anything changes
 			continue
 		}
-		path := strings.Split(k, v.keyDelim)
+		path := keyComponents(v, k)
 		lastKey := strings.ToLower(path[len(path)-1])
 		deepestMap := deepSearch(m, path[0:len(path)-1])
 		// set innermost value
--- a/viper_test.go
+++ b/viper_test.go
@@ -13,6 +13,7 @@ import (
 	"os"
 	"os/exec"
 	"path"
+	"path/filepath"
 	"reflect"
 	"runtime"
 	"sort"
@@ -45,6 +46,10 @@ clothing:
 age: 35
 eyes : brown
 beard: true
+emails:
+  steve@hacker.com:
+    created: 01/02/03
+    active: true
 `)
 
 var yamlExampleWithExtras = []byte(`Existing: true
@@ -207,11 +212,16 @@ func initHcl() {
 func initDirs(t *testing.T) (string, string, func()) {
 
 	var (
-		testDirs = []string{`a a`, `b`, `c\c`, `D_`}
+		testDirs = []string{`a a`, `b`, `C_`}
 		config   = `improbable`
 	)
 
+	if runtime.GOOS != "windows" {
+		testDirs = append(testDirs, `d\d`)
+	}
+
 	root, err := ioutil.TempDir("", "")
+	require.NoError(t, err, "Failed to create temporary directory")
 
 	cleanup := true
 	defer func() {
@@ -224,7 +234,7 @@ func initDirs(t *testing.T) (string, string, func()) {
 	assert.Nil(t, err)
 
 	err = os.Chdir(root)
-	assert.Nil(t, err)
+	require.Nil(t, err)
 
 	for _, dir := range testDirs {
 		err = os.Mkdir(dir, 0750)
@@ -415,7 +425,10 @@ func TestEmptyEnv(t *testing.T) {
 	BindEnv("type") // Empty environment variable
 	BindEnv("name") // Bound, but not set environment variable
 
-	os.Clearenv()
+	os.Unsetenv("type")
+	os.Unsetenv("TYPE")
+	os.Unsetenv("name")
+	os.Unsetenv("NAME")
 
 	os.Setenv("TYPE", "")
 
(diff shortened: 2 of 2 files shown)


mschneider82:
Hi can you please rebase this into once commit? So we stand a chance to get this reviewed fast and merged?

inkychris:
Sure thing, will do it when I get home this afternoon.

albertvaka:
We have very similar code in our fork: https://github.com/DataDog/viper/pull/2/files

inkychris:
@albertvaka yes looks very similar, although yours uses the `keyDelim` property which I should change in this.

xixuejia:
Hello, any update on this? I also encountered this issue

silvin-lubecki:
Hello @inkychris, any update on this PR? How can I help to make it merge?

inkychris:
@silvin-lubecki As far as I'm aware, this should be okay to merge; it's just waiting on a maintainer to review/approve.

spf13:
This looks good. Can you rebase against tip and I'll gladly merge.

inkychris:
@spf13, this should be good to go now, thanks!

silvin-lubecki:
ping @spf13 😺

mschneider82:
please merge @spf13

inkychris:
@spf13 @sagikazarmark any chance one of you could merge this please? I'd rather not have to keep rebasing... Thanks!

sagikazarmark:
I'll try to take a look at it soon, but no promises. :)

mschneider82:
@sagikazarmark ping ;-)

mschneider82:
Hopefully someone can review this and merge, its so annoying to be stuck on a old forked patched version :( I have settings for "domains", domains always contains dot's...

sagikazarmark:
I started reviewing this PR. I like the idea and the implementation, but I suspect there might be an edge case where this PR introduces a backwards incompatible change:

```
emails:
  steve@hacker:
    com:
      created: 01/02/04
      active: false
  steve@hacker.com:
    created: 01/02/03
    active: true
```

In this case I would expect that `emails.steve@hacker.com.active` returns false, because that's the original behavior.

@inkychris can you add a test case (probably with a different email address) that verifies the original behavior remains intact in such cases?

Otherwise 👍 and sorry you had to wait so much for moving forward with this. Thanks @inkychris for the PR.

inkychris:
For me, the original behaviour results in the value of that key being `true` rather than `false`. This is the test I'm running but I could be checking the wrong thing:
```go
package viper

import (
	"github.com/stretchr/testify/assert"
	"strings"
	"testing"
)

const yamlConfig = `
emails:
  steve@hacker:
    com:
      created: 01/02/04
      active: false
  steve@hacker.com:
    created: 01/02/03
    active: true
`

func TestConflict(t *testing.T) {
	SetConfigType("yaml")
	assert.NoError(t, ReadConfig(strings.NewReader(yamlConfig)))
	assert.True(t, GetBool("emails.steve@hacker.com.active"))
}
```
I should also note I'm testing on Windows with go 1.12 in case this is something that arises from un-ordered parsing or something?

I would expect the issue here is that the address `emails.steve@hacker.com.active` ambiguously refers to both properties but I guess this is a pre-existing issue anyway?

Apologies, it's been a little while since I thought about this so trying to get up to speed again; thanks for looking at it!

sagikazarmark:
As far as I can tell the issue is only related to unmarshaling. Keys containing dots is supported by Viper AFAIK, so `Get`-ing such keys should work.

IMHO the right test would be:

```
package viper

import (
	"strings"
	"testing"

	"github.com/stretchr/testify/assert"
)

const yamlConfig = `
emails:
  steve@hacker:
    com:
      created: 01/02/04
      active: false
  steve@hacker.com:
    created: 01/02/03
    active: true
`

func TestConflict(t *testing.T) {
	SetConfigType("yaml")
	assert.NoError(t, ReadConfig(strings.NewReader(yamlConfig)))

	var v string
	assert.NoError(t, UnmarshalKey("emails.steve@hacker.com.created", &v))
	assert.Equal(t, "01/02/04", v)
}
```

Currently this test fails with the following error:

```
=== RUN   TestConflict
--- FAIL: TestConflict (0.00s)
    conflict_test.go:27:
        	Error Trace:	conflict_test.go:27
        	Error:      	Not equal:
        	            	expected: "01/02/04"
        	            	actual  : "01/02/03"

        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1 +1 @@
        	            	-01/02/04
        	            	+01/02/03
        	Test:       	TestConflict
FAIL
FAIL	github.com/spf13/viper	0.014s
```

sagikazarmark:
Hm, I guess you are right. After thinking it through again, it makes sense. Otherwise the fix itself wouldn't work in the first place, since it builds on keys being `Get`able.

mschneider82:
thanks @sagikazarmark and @inkychris 
cannot wait to change my import paths from custom to upstream..

bogdandrutu:
Any update on this? Can this be merged?

sagikazarmark:
Yup, I think it's good to go.

sagikazarmark:
Well, it seems this PR did introduce a regression after all. Dot-notated keys so far were always parsed as structures. Sources, like manually set values, defaults and flags always use dot notated keys, compared to config files:

```
v.Set("foo.bar", "hello")
v.Default("foo.bar", "hello")
pflag.String("foo.bar", "hello", "")
```

In this case there is no way to know the original intention (is the key `foo.bar` or `bar` under `foo`?). In the unmarshaling scenario this can probably be solved by including both versions in the source map, but `AllSettings` should return the same as before.

At this point we either revert this PR, or change it's scope to unmarshal only and generate a map that's compatible with both approaches, although that would potentially be much slower than the current solution.

bogdandrutu:
In order to support both use-cases (I have an usecase where some keys contain `.`), if we revert this, can we allow setting the keyDelimiter? I want to set it to `..` or something that does allow me to have a key with one `.`.

mschneider82:
setting keyDelimiter is fine with me, no solution would be bad

sagikazarmark:
I submitted a PR reverting the effective changes of this PR (I left the reformatting in place).

Sorry folks, this cannot land in the current version. 🙁 

I will send another PR for setting the key delimiter, although we will have to check first if it has any unwanted effects.

inkychris:
The underlying issue was conceptual (not being able to determine the key path components from it's fully qualified path) which is only really solvable by either escaping or changing the delimiter used, so it's understandable.

Just a heads up for the following PR that the following line appears to have missed the `keyDelim` refactoring, in case tests don't hit this:

https://github.com/spf13/viper/blob/99520c81d86ee7d0ee2ef143c8576be0c7533827/viper.go#L1441

sagikazarmark:
Thanks for the heads up @inkychris !

PR #675, opened 2019-03-26
Warn when a yaml file has duplicated keys

albertvaka:
Try using `UnmarshalStrict`, if it fails, print the errors as a warning and try again with plain `Unmarshal`.

Warnings look like this:
```
2019/03/26 15:11:33 warning reading config file: yaml: unmarshal errors:
  line 6: key "api_key" already set in map
```


albertvaka:
Closing since this didn't get any attention 😢

PeterBocan:
@albertvaka I would like to see the response from maintainers, as this is, IMO crucial.

Issue #680, opened 2019-04-02
Exchangeable Logger


generikvault:
Hi,

I want to use viper with a custom logger. Viper contains currently logs to `log.Fatal`, `log.Printf`, `jww.INFO.Println`, `jww.ERROR.Println`.

Both `log` and `jww` (jwalterweatherman) allow custom output streams but using them gets a bit hacky with my system wide log format, log level and special log files for different components.

If you're interessested I would create a pull request for a more exchangeable logging approach.

That could probably look like this:
``` go
// Logger collects log messages.
var Logger interface {
    Infof(string, ...interface{})
    Errorf(string, ...interface{})
    Fatalf(string, ...interface{})

    Info(...interface{})
    Error(...interface{})
    Fatal(...interface{})
} = logrus.New()

func DoX(){
    logger.Info("Done X")
}
```

Kind Regards

GeneriK

PR #682, opened 2019-04-08
#680 hide logger behind an interface

generikvault:
Hi,
As described in #680 I want to use viper with a custom logger. Viper contains currently logs to `log.Fatal`, `log.Printf`, `jww.INFO.Println`, `jww.ERROR.Println`.

Both `log` and `jww` (jwalterweatherman) allow custom output streams but using them gets a bit hacky with my system wide log format, log level and special log files for different components.

So I've hidden the actual Logger behind an interface like this:
``` go
// Logger collects log messages.
var Logger interface {
    Infof(string, ...interface{})
    Errorf(string, ...interface{})
    Fatalf(string, ...interface{})
    ...

    Info(...interface{})
    Error(...interface{})
    Fatal(...interface{})
    ...
} = logrus.New()

func DoX(){
    logger.Info("Done X")
}
```

Kind Regards

GeneriK


sagikazarmark:
Can we use the builtin log package or a simple wrapper around it instead? I'm not a huge fan of depending on external libraries in projects like this one.

generikvault:
Good point. I wrote the local logger type in a way, that it ignores trace, debug and warning messages analog to the defaultsetting of jwalterweatherman.

generikvault:
@sagikazarmark do you like the local logger type as implemented?

sagikazarmark:
I think it would be better to have a separate interface defined (not inline), and instead of a single global logger, one should be able to override it per viper instance as well.

`viper.New` should probably initialize a default logger, so a global logger would essentially become unnecessary.

PR #686, opened 2019-04-16
switch to gopkg.in/fsnotify.v1

dims:
Can we please start using the v1 API for fsnotify explicitly:
https://gopkg.in/fsnotify.v1

v1 refers to v1.4.7 version (essentially
https://github.com/fsnotify/fsnotify/tree/v1.4.7)

Change-Id: I872bbbc8937c065da70e5ed7641c2939aab60a41

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,7 +5,6 @@ require (
 	github.com/coreos/etcd v3.3.10+incompatible // indirect
 	github.com/coreos/go-etcd v2.0.0+incompatible // indirect
 	github.com/coreos/go-semver v0.2.0 // indirect
-	github.com/fsnotify/fsnotify v1.4.7
 	github.com/hashicorp/hcl v1.0.0
 	github.com/magiconair/properties v1.8.0
 	github.com/mitchellh/mapstructure v1.1.2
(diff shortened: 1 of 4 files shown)


dims:
@spf13 this helps avoid having 2 copies of fsnotify in kubernetes tree as well, please see https://github.com/kubernetes/kubernetes/pull/76652#issuecomment-483724753

sagikazarmark:
Viper uses go modules and already uses the version: https://github.com/spf13/viper/blob/master/go.mod#L8

As far as I can tell, fsnotify doesn't use gopkg, at least it's not mentioned in their readme.

dims:
@sagikazarmark the problem is not the version number, the problem is with the url itself. kubernetes also uses go modules and we are ending up having 2 copies of fsnotify that we would like to avoid.

thanks!

dims:
please try `go get -d gopkg.in/fsnotify.v1` to confirm

sagikazarmark:
I understand the problem, but as I mentioned, using the gopkg url is an arbitrary decision of Kubernetes, I can't find any places where it's encouraged by fsnotify itself, and honestly since we have gomodules, I'd rather stick to the original path.

dims:
@sagikazarmark to be clear, kubernetes is currently using `github.com/fsnotify/fsnotify`. we are trying to import `github.com/hpcloud/tail` which is using `"gopkg.in/fsnotify.v1"`

https://github.com/hpcloud/tail/blob/master/watch/inotify.go#L13

Thanks for looking at it!

PR #690, opened 2019-04-24
Add deep map logic to config Aliases

### LINKED ISSUE #689: RegisterAlias fails for nested config
If you ReadConfig on a yaml file like:
```
foo:
  bar: helloworld
```
then register a dot delimited alias for a flag, like: 
```
viper.RegisterAlias("foo.bar", "foobar")
```
one would expect retrieving the value by alias or real key like:
```
viper.GetString("foo.bar")
// or
viper.GetString("foobar")
```
to return 'helloworld'.  

However, the actual return is 'helloworld' for foo.bar and '' for foobar.  This is because the RegistAlias function does not honor nested config types.

therealmitchconnors:
Fixes #689 

This logic may need to be extended to other sources of config, but I'm less certain about how nesting is supported outside of config files.

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -1139,6 +1139,12 @@ func (v *Viper) registerAlias(alias string, key string) {
 			// if we alias something that exists in one of the maps to another
 			// name, we'll never be able to get that value using the original
 			// name, so move the config value to the new realkey.
+			path := strings.Split(alias, v.keyDelim)
+			if val := v.searchMapWithPathPrefixes(v.config, path); val != nil {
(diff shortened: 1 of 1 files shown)


sagikazarmark:
@therealmitchconnors thanks a lot for the PR. I will take a look at your changes to see if there is anything else that needs to be done.

spacez320:
Wondering if this PR is worth merging still? Not sure about the `go.sum` changes, but the rest of it seems reasonable.

PR #720, opened 2019-06-19
Add GetIntSlice helper method

zer0tonin:
Kinda follow up this PR : https://github.com/spf13/viper/pull/637

Adds a simple `GetIntSlice` methods that enables parsing int slices from config the same way `GetStringSlice` works.


sagikazarmark:
Thanks @zer0tonin!

PR #721, opened 2019-06-20
BindEnv details specified. Now less confusion.

BruceWangNo1:
This is used to clarify more on BindEnv to keep newcomers from confusion.


sagikazarmark:
Thanks @BruceWangNo1 . I modified your proposed change a bit, but I think the meaning remained the same.

BruceWangNo1:
Fine by me @sagikazarmark

Issue #390, opened 2017-09-22
config file not found


gaoshengL:
hello,i code the example:
         viper.SetConfigType("yaml")
        //viper.SetConfigFile("/etc/config.yaml")
        viper.SetConfigName("config.yaml")
        viper.AddConfigPath("/etc")
        err := viper.ReadInConfig()
        if err != nil {
                fmt.Println(err)
        }
or
        viper.SetConfigType("yaml")
        viper.SetConfigFile("config.yaml")
        //viper.SetConfigName("config.yaml")
        viper.AddConfigPath("/etc")
        err := viper.ReadInConfig()
        if err != nil {
                fmt.Println(err)
        }
i'm sure that i have config.yaml in /etc dirctionary,but the outputs of these  code are:
Config File "config.yaml" Not Found in "[/etc]"
open config.yaml: no such file or directory
what's wrong?

ryanthedev:
I'm not sure but maybe try
``` go
viper.SetConfigType("yaml")
viper.SetConfigName("config")
viper.AddConfigPath("/etc")
err := viper.ReadInConfig()
if err != nil {
  fmt.Println(err)
}
```

or try

``` go
viper.SetConfigFile("config.yaml")
viper.AddConfigPath("/etc")
err := viper.ReadInConfig()
if err != nil {
   fmt.Println(err)
}
```

nad2000:
I confirm the issue with "not found" configuration. The proposed solution doesn't solve it.

nad2000:
It looks that config name should be without any extension (**viper.SetConfigName(...)**) while the actual file should have any supported extension. Take a look into https://github.com/spf13/viper/blob/d9cca5ef33035202efb1586825bdbb15ff9ec3ba/viper.go#L1536

mfridman:
@nad2000 is correct. `searchInPath` concatenates search path, file name and extension. Which appears to be the intended behaviour. `filepath.Join(in, v.configName+"."+ext)`

So viper will search, your filepaths + filename + extension:
```
/Users/example/go/src/github.com/your-app/config.json
/Users/example/go/src/github.com/your-app/config.toml
/Users/example/go/src/github.com/your-app/config.yaml
/Users/example/go/src/github.com/your-app/config.yml
/Users/example/go/src/github.com/your-app/config.properties
/Users/example/go/src/github.com/your-app/config.props
/Users/example/go/src/github.com/your-app/config.prop
/Users/example/go/src/github.com/your-app/config.hcl
```

[...]

Issue #336, opened 2017-04-20
viper.ConfigFileNotFoundError when file most definitely exists.


theshadow:
```
~/P/g/s/e/b/rbaapi ❯❯❯ DEBUG=1 go run main.go register 89a1029f-bdaa-41f0-811e-66ef878faabb
viper.ConfigFileNotFoundError{name:".rbaapi", locations:"[/Users/xguzman]"}
<nil>Error: Unable to convert defined API DSN into a string
Usage:
  rbaapi register <UUID> [flags]

Global Flags:
      --config string   config file (default is $HOME/.rbaapi)

Unable to convert defined API DSN into a string
exit status 255
~/P/g/s/e/b/rbaapi ❯❯❯ file /Users/xguzman/.rbaapi                                                                                                                                                                                                                                                                                                                        ⏎
/Users/xguzman/.rbaapi: ASCII text
~/P/g/s/e/b/rbaapi ❯❯❯ ls -al /Users/xguzman/.rbaapi
-rw-r--r--@ 1 xguzman  staff  80 Apr 20 08:26 /Users/xguzman/.rbaapi
```

theshadow:
I figured it out. The code checks for **filename.ext**. Even when specifying the config file with --config. This is unexpected behavior.

PR #722, opened 2019-06-20
Support config files with no extensions

pedromss:
Support config files with no extension.

Closes: spf13/viper#390
Closes: spf13/viper#336

Similar to https://github.com/spf13/viper/pull/357/files which appears to have been forgotten. 

Difference is: attempt a no extension right away instead of messing with the extension append and iterating through the extensions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,6 +107,8 @@ if err != nil { // Handle errors reading the config file
 }
 ```
 
+*NOTE:* You can also have a file without an extension and specify the format programmaticaly. For those configuration files that lie in the home of the user without any extension like `.bashrc`
+
 ### Writing Config Files
 
 Reading from config files is useful, but at times you want to store all modifications made at run time.
--- a/viper.go
+++ b/viper.go
@@ -1863,6 +1863,10 @@ func (v *Viper) getConfigFile() (string, error) {
 
 func (v *Viper) searchInPath(in string) (filename string) {
 	jww.DEBUG.Println("Searching for config in ", in)
+	if b, _ := exists(v.fs, filepath.Join(in, v.configName)); b {
+		return filepath.Join(in, v.configName)
+	}
+
 	for _, ext := range SupportedExts {
 		jww.DEBUG.Println("Checking for", filepath.Join(in, v.configName+"."+ext))
 		if b, _ := exists(v.fs, filepath.Join(in, v.configName+"."+ext)); b {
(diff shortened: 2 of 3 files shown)


GJRTimmer:
Hi, I really want this, any news when it will be merged ?

@pedromss maybe a update of the README is required ?
Maybe some note that if using a config file without extension `viper.SetConfigType(...)` is required for viper to deserialize the config ?

pedromss:
@GJRTimmer I don't know what else is required to get this approved. Seems like this repo isn't maintained judging by the state of the PR and issue section 

¯\_(ツ)_/¯

dmacvicar:
I like this approach as well.
Please merge either this or my own https://github.com/spf13/viper/pull/357, which is open for more than 2 years.

sagikazarmark:
@pedromss Just for me to understand: you want to explicitly set a config name with an extension?

```go
viper.SetConfigNam("config.yaml")
```

And you want to search for that specific file within the configured search paths?

Is that correct?

pedromss:
@sagikazarmark no. The idea is to be able to have configuration files that don't have an extension in their name. As is many times the case with tools which have config files in the home folder for example.

So supporting `.toolconf` instead of `.toolconf.yaml`  is the goal of this PR

sagikazarmark:
@pedromss  I see. Thanks for the explanation.

pedromss:
@sagikazarmark please take another look now

sagikazarmark:
Thanks @pedromss !

sagikazarmark:
@pedromss actually, I wonder if the condition should be moved after the the loop, to make sure existing behavior isn't broken (eg. there is a config and a config.yaml file in a folder. this PR would break existing code in those cases).

WDYT?

pedromss:
@sagikazarmark I had thought to resolve it immediately without "unnecessarily" walking the files in the folder. But what you say makes more sense.

Changed it.

sagikazarmark:
Awesome, thanks!

This is a one time step, so I wouldn't worry about unnecessary checks for non-existing files.

andig:
I'm a bit confused as to how far this fixes #390. #390 doesn't seem to be about extension-less files?

pedromss:
You're correct 😕 I must have made a mistake when analysing the issue. Reading it again, I would say this comment https://github.com/spf13/viper/issues/390#issuecomment-337650626 made me think the problem was the no extension, which was a problem in it self but not of that issue.

Thank you. I'll edit this description and reopen #390

Issue #430, opened 2017-12-22
Simple way to initialize a config file


michaeleekk:
Hello. I was searching for a way to initialize the config file, but still couldn't find one.

```
viper.SetConfigName("config")
viper.SetConfigType("json")
viper.AddConfigPath("$HOME")
```

And then I tried the new `SafeWriteConfig()` or any of the variety merged two weeks ago, but it didn't create new file or it will overwrite the original config file no matter what. And because of the `$HOME`, I might need to write a function to be cross-platform. I just wish there could be a cleaner way.

Or we could add `O_CREATE` to this line of code ? 
https://github.com/spf13/viper/blob/master/viper.go#L1254

And also is it a good idea to have `SafeWriteConfig()` to create a new file with the first config path and config name ?

revilwang:
Is there anyone working on this issue? 18 months passed and `WriteConfig` still requires the file must exist before writing, which is different from what the docs say.

berlincount:
This seems to be fixed?

kevingentile:
`v1.4.0`: `SafeWriteConfig()` does not initialize a new config file:

```
File "config" Not Found in "[{path hidden}]"
```

errogaht:
as for me it is strange. i want to use viper to manage all cases with config files, but i have to manually create file, why?

malohr:
Hi there, not sure if this is still an issue but here is the simple solution you were looking for:

```
//using the home dir
home, err := os.UserHomeDir()
cobra.CheckErr(err)

configName := ".mycfg"

viper.WriteConfigAs(fmt.Sprintf("%s/%s", home, configName))
```

marcosanchotene:
Adding the call to `viper.SafeWriteConfig()`, after configuring the path and file name, worked for me:

```
func initConfig() {
	if cfgFile != "" {
		// Use config file from the flag.
		viper.SetConfigFile(cfgFile)
	} else {
		// Find home directory.
		home, err := os.UserHomeDir()
		cobra.CheckErr(err)

		// Search config in home directory with name ".config.yaml".
		viper.AddConfigPath(home)
		viper.SetConfigType("yaml")
		viper.SetConfigName(".config")
		viper.SafeWriteConfig()
	}

}
```

PR #723, opened 2019-06-21
Initialize the configuration file during writing if not exist

RyanSiu1995:
Issue #430


metrue:
This would be a pretty handy feature.

dpolicastro:
Why this was not merged?

PR #528, opened 2018-07-09
 Support `.env` format files

mexisme:

Issue #455, opened 2018-02-07
Support for .env files?


cemremengu:
Is there support for a feature like provided by [this library](https://github.com/joho/godotenv)?

Basically if `.env` file exists it will read variables from there, otherwise from OS.

bxcodec:
Someone already submits a PR here, #528 
Look forward to this support

boskiv:
How I can get it? I see that it already merge but not in release.

sagikazarmark:
Adding `master` to a `go.mod` file is not valid indeed.

You should use:

```
go get github.com/spf13/viper@master
```

boskiv:
I can't 
```
➜  go-open-registry git:(git-registry) ✗ go get
go get: disabled by -mod=vendor

```

Jyny:
#756 
add ability to search `.env` or other dotfiles

Jyny:
> -

when will have new release?

Jyny:
@sagikazarmark  @boskiv 
though PR is merged but not in latest release
when will have new release?

axiaoxin:
What is the current situation here? Is there any conclusion or progress?

CSRaghunandan:
+1

Would love to see this feature in viper

sagikazarmark:
Unless I'm mistaken #528 added support for .env files and it's been released for quite some time now.

CSRaghunandan:
> Unless I'm mistaken #528 added support for .env files and it's been released for quite some time now.

@sagikazarmark is there any documentation on how to use it?

I set the config type to `dotenv` and I get an error saying its not a supported type.

sagikazarmark:
.env and .dotenv types are working similarly to any config types. Looking at the code, you'll have to set an explicit config file if you want to use `.env`. A file called `config.env` should automatically be discovered.

CSRaghunandan:
> .env and .dotenv types are working similarly to any config types. Looking at the code, you'll have to set an explicit config file if you want to use `.env`. A file called `config.env` should automatically be discovered.

I see. I will test it out later today and will report if I was able to get it to work.

Thanks for the quick response.

CSRaghunandan:
@sagikazarmark it worked! Now I'm able to read from a file called `config.env` which holds all the secret data.

Thank you :)

abdennour:
what about .env filename ?

abdennour:
is there any documentation on how to use it?

I set the config type to dotenv and I get an error saying its not a supported type.

[...]

PR #756, opened 2019-08-29
Support search `.env` or other dotfiles in path

Jyny:


Jyny:
fix #16 #455

Jyny:
duplicate #528

Issue #643, opened 2019-02-05
environment variables do not follow subtree structure (viper.Sub)


aka-rider:
I think that the following behaviour is misleading.

```go
package main

import (
	"bytes"
	"fmt"
	"os"
	"strings"

	"github.com/spf13/viper"
)

var yamlConf = []byte(`

topic:
    subtopic1:
        name: subtopic1
        value: value1

`)

func main() {
	// try to add subtopic
	os.Setenv("TOPIC_SUBTOPIC2_NAME", "subtopic2")
	os.Setenv("TOPIC_SUBTOPIC2_VALUE", "value2")
	viper.SetConfigType("yaml")
	viper.ReadConfig(bytes.NewBuffer(yamlConf))
	viper.SetEnvKeyReplacer(strings.NewReplacer(".", "_", "-", "_"))
	viper.AutomaticEnv()

	if st1 := viper.Sub("topic.subtopic1"); st1 != nil {
		fmt.Println("We've got subtopic1 as expected")
	}
	if st2 := viper.Sub("topic.subtopic2"); st2 == nil {
		fmt.Println("It's not there")
	}
	if n := viper.Get("topic.subtopic2.name"); n != nil {
		fmt.Println("We still can access it but not like expected")
	}
}

```

In order to get subtopic2, one should add it in the yaml, then environment variables will work as expected.

P.S. Thank you for all the hard work. The library is great.

yashsriv:
There is still an issue even after this (adding subtopic2 in the yaml) as the following doesn't work as expected:

```
var yamlConf = []byte(`

topic:
    subtopic1:
        name: subtopic1
        value: value1
	subtopic2:
		value: value2
`)

...
os.Setenv("TOPIC_SUBTOPIC2_VALUE", "value4")
st2 := viper.Sub("topic.subtopic2")
fmt.Println(viper.GetString("topic.subtopic2.value"))
// the above prints value2 instead of value4
```

See http://github.com/yashsriv/viper-bool-subtree-bug.git for a fully working example.

PR #760, opened 2019-09-02
Read onlu view with sub config

rotium:
Create read only view which can safely been shared in the code.
It also support sub config with environment variables follows (This solve the #643).


rotium:
Thanks for your time.
The .idea files accidentally got in.  I'll remove them.  :)
Using interface provide two things: easy way to mock in tests and ability to wrap the implementation. I actually think that the New() method should return an interface, instead of the implementation.  I didn't made that change since I didn't want to modify the core file.
The R/O view is helpful to provide way to ensure it wont be modify, and "Viper is usually not modified after it's bootstrapped" is right, but it still "usual", and if it isn't the case then it may cause bugs which is hard to find.  

I'll do the changes.
If you think it can provide value to the project, and think more changes have to be done, let me know.  If not just close this PR.

Thanks again.

Issue #549, opened 2018-07-31
SafeWriteConfigAs doesn't work


deinspanjer:
@theherk added this method along with several others related to exporting config.  I don't believe the tests cover this particular method though, and there is an issue with it.

It's intended purpose is to only write out if the file does not exist, however, the writeConfig method only uses `flags = os.O_WRONLY` so the call to `v.fs.OpenFile` fails since it doesn't exist and the  `O_CREATE` flag isn't specified.

itsjamie:
@deinspanjer could you extend a PR related to this?

deinspanjer:
This is in https://github.com/spf13/viper/pull/550

Issue #525, opened 2018-07-04
`SafeWriteConfig` is not working when file not exists.


ryoya-fujimoto:
`SafeWriteConfig` execute `writeConfig` function with `force` flag to false.
But in `writeConfig`, not added `os.O_CREATE` flag when that.

```go
if force == true {
        flags = os.O_CREATE | os.O_TRUNC | os.O_WRONLY
} else {
        if _, err := os.Stat(filename); os.IsNotExist(err) {
                flags = os.O_WRONLY
        } else {
                return fmt.Errorf("File: %s exists. Use WriteConfig to overwrite.", filename)
        }
}
```
https://github.com/spf13/viper/blob/master/viper.go#L1256

For that reason, `SafeWriteConfig` raise this error when file not exists.
```
open conf.toml: no such file or directory
```

Is this the intended behavior?

ryoya-fujimoto:
Sorry, I found PullRequest in https://github.com/spf13/viper/pull/450. But cannot find related issue.
So, I thought should keep this issue was opened.

sureshannadurai:
Use viper.SetConfigFile("test.json") instead of   viper.SetConfigName("test") then SafeWriteConfig and SafeWriteConfigAs() will work

Issue #448, opened 2018-01-28
SafeWriteConfig() fails if file does not exist


TWinsnes:
When calling viper.SafeWriteCofig() it fails with a file not found error. 

Test code attached.

```
package main

import (
	"fmt"
	"os"

	"github.com/spf13/viper"
)

func main() {
	viper.AddConfigPath("/tmp")
	viper.SetConfigName("filename")
	viper.SetConfigType("yaml")
	viper.SetDefault("LogLevel", "Info")

	fmt.Println("Writing Config file!")
	err := viper.SafeWriteConfig()

	if err != nil {
		fmt.Println(err)
		os.Exit(1)
	}

	fmt.Println("wrote config file")
}
```

Output:
```
Writing Config file!
Config File "filename" Not Found in "[/tmp]"
```

Expected output:
```
Write Config file!
wrote config file
```

tskardal:
+1

Currently the documentation for SafeWriteConfig is quite misleading 😅 

> SafeWriteConfig writes current configuration to file only if the file does not exist.

Vaelatern:
This is fixed in https://github.com/spf13/viper/pull/450

sureshannadurai:
Use viper.SetConfigFile("filename.yaml") instead of viper.SetConfigName("filename") then SafeWriteConfig and SafeWriteConfigAs() will work

Issue #433, opened 2018-01-09
WriteConfig fails to write if file does not exist


lllama:
The new `WriteConfig` function ends up calling `findConfigFile`, which in turn checks whether the file exists. If it doesn't then it returns an error, which stops the config being written.

michaeleekk:
I had similar issue too and filed as #430 and right now my workaround is like this, 

```
if err := viper.SafeWriteConfigAs(configPath); err != nil {
	if os.IsNotExist(err) {
		err = viper.WriteConfigAs(configPath)
	}
}
```

I'm just hoping there is a better solution.

parkervcp:
Use `WriteConfigAs` which creates the file before writing.

koooge:
I got same issue.
I think WriteConfig should create configfile as well as WriteConfigAs

mloskot:
@parkervcp 
> Use `WriteConfigAs` which creates the file before writing.

Nope, it does not. So `WriteConfigAs` in https://github.com/spf13/viper/issues/433#issuecomment-356483379 will fail if directory does not exist where the config file is supposed to be written.

PR #763, opened 2019-09-13
Updating SafeWriteConfig and SafeWriteConfigAs to match documented be…

javaducky:
…havior.

 Methods should throw an error if the config file already exists or if no configpath is configured when not explicitly requesting a write path.

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -115,6 +115,22 @@ func (fnfe ConfigFileNotFoundError) Error() string {
 	return fmt.Sprintf("Config File %q Not Found in %q", fnfe.name, fnfe.locations)
 }
 
+// ConfigFileAlreadyExistsError denotes failure to write new configuration file.
+type ConfigFileAlreadyExistsError string
+
+// Error returns the formatted error when configuration already exists.
+func (faee ConfigFileAlreadyExistsError) Error() string {
+	return fmt.Sprintf("Config File %q Already Exists", string(faee))
+}
+
+// MissingConfigurationError denotes a required configuration setting has not been provided.
+type MissingConfigurationError string
+
+// Error returns the formatted error when a required configuration element has not been provided.
+func (mce MissingConfigurationError) Error() string {
+	return fmt.Sprintf("Missing Configuration for %q", string(mce))
+}
+
 // A DecoderConfigOption can be passed to viper.Unmarshal to configure
 // mapstructure.DecoderConfig options
 type DecoderConfigOption func(*mapstructure.DecoderConfig)
@@ -1331,11 +1347,10 @@ func (v *Viper) WriteConfig() error {
 // SafeWriteConfig writes current configuration to file only if the file does not exist.
 func SafeWriteConfig() error { return v.SafeWriteConfig() }
 func (v *Viper) SafeWriteConfig() error {
-	filename, err := v.getConfigFile()
-	if err != nil {
-		return err
+	if len(v.configPaths) < 1 {
+		return MissingConfigurationError("configPath")
 	}
-	return v.writeConfig(filename, false)
+	return v.SafeWriteConfigAs(filepath.Join(v.configPaths[0], v.configName+"."+v.configType))
 }
 
 // WriteConfigAs writes current configuration to a given filename.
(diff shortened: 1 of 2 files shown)


javaducky:
**I apologize for yet another PR** on this issue; I hadn't performed enough due diligence before working up a patch.  I'd still like to provide this version as I am providing custom errors to note the specific reason for failure during file initialization.

From my review of open issues, it would appear this may address the following issues: #549, #525, #448 and possibly clear up #433. 

Thanks!

Paul

javaducky:
Hoping to get this PR looked at.  As it is now, the current descriptions for the SafeWrite* methods in the README.md do not match reality.  This PR will correct that so the expected exceptions are encountered and verifies this as well with the included unit tests.

Here's the description which has the intended use:
image

javaducky:
@sagikazarmark @spf13 

I'm sure this project is encountering the same issue (lack of time, etc.) as Cobra ([cobra#959](https://github.com/spf13/cobra/issues/959)), but I'd be willing to help out.  I'm relatively new to the GoLang scene, but have been a software developer longer than I'd like to admit.  I appreciate the efforts of all committers and maintainers and would love to pitch in!

sagikazarmark:
@javaducky thanks for the offer. It would be a great help if you could review submitted PRs.

The primary reason why the development of Viper stalled is that it became insanely complex, many projects rely on it, so avoiding breaks is not just important, but relatively hard as well.

Another problem is that (obviously) sometimes people want to add features that they want to use, but not necessarily fit into the current Viper ecosystem.

And obviously there is time...

So for now I would say rigorously reviewing PRs is the easiest way to jump on the project.

viper.go
```diff
@@ -115,6 +115,22 @@ func (fnfe ConfigFileNotFoundError) Error() string {
...
+// ConfigFileAlreadyExistsError denotes failure to write new configuration file.
+type ConfigFileAlreadyExistsError string
+
+// Error returns the formatted error when configuration already exists.
+func (faee ConfigFileAlreadyExistsError) Error() string {
+	return fmt.Sprintf("Config File %q Already Exists", string(faee))
+}
+
+// MissingConfigurationError denotes a required configuration setting has not been provided.
+type MissingConfigurationError string
```
sagikazarmark: I wonder if it makes sense to have a separate error type for this. Not found and already exists kinda make sense to me, but this is hardly a recoverable error, so I would probably replace the usage of this error with `errors.New`.

WDYT?

viper.go
```diff
@@ -1347,6 +1362,10 @@ func (v *Viper) WriteConfigAs(filename string) error {
 // SafeWriteConfigAs writes current configuration to a given filename if it does not exist.
 func SafeWriteConfigAs(filename string) error { return v.SafeWriteConfigAs(filename) }
 func (v *Viper) SafeWriteConfigAs(filename string) error {
+	handle, err := v.fs.Stat(filename)
```
sagikazarmark: Afero has an exists call, tsh might look better here: https://godoc.org/github.com/spf13/afero#Afero.Exists

viper_test.go
```diff
@@ -1349,6 +1349,86 @@ func TestWriteConfigYAML(t *testing.T) {
...
 
+func TestSafeWriteConfig(t *testing.T) {
+	v := New()
+	fs := afero.NewMemMapFs()
+	v.SetFs(fs)
+	v.AddConfigPath("/test")
+	v.SetConfigName("c")
+	v.SetConfigType("yaml")
+	err := v.ReadConfig(bytes.NewBuffer(yamlExample))
+	if err != nil {
```
sagikazarmark: We tend to use `require.NoError` here (same for bellow).

viper_test.go
```diff
@@ -1349,6 +1349,86 @@ func TestWriteConfigYAML(t *testing.T) {
...
+}
+
+func TestSafeWriteConfigWithMissingConfigPath(t *testing.T) {
+	v := New()
+	fs := afero.NewMemMapFs()
+	v.SetFs(fs)
+	v.SetConfigName("c")
+	v.SetConfigType("yaml")
+	err := v.SafeWriteConfig()
+	if err == nil {
```
sagikazarmark: We tend to use `require.NoError` here.

viper_test.go
```diff
@@ -1349,6 +1349,86 @@ func TestWriteConfigYAML(t *testing.T) {
...
+		t.Fatal("Expected exception")
```
sagikazarmark: This is  not Java 🙂 There are no exceptions, but errors.

javaducky: > This is not Java 🙂 There are no exceptions, but errors.

LOL...you caught me.

viper.go
javaducky: Seems reasonable to me!  I'll update that.

PR #766, opened 2019-09-18
[WIP]: Fix Unmarshalling when only PFlags or Envs are set

### LINKED ISSUE #764: PR #673 broke my code. The nested structures don't get the default values correctly
I wrote a short example to illustrate the problem.

```Go
package main

import (
	"fmt"

	"github.com/spf13/cobra"
	"github.com/spf13/viper"
)

type Config struct {
	Foo struct {
		Bar string
	}
}

func main() {
	config := &Config{}
	cmdRoot := &cobra.Command{
		Run: func(*cobra.Command, []string) {
			viper.Unmarshal(config)
		},
	}
	flags := cmdRoot.PersistentFlags()
	flags.String("foo.bar", "hello-world", "test default value fo foo.bar")
	viper.BindPFlags(flags)
	cmdRoot.Execute()
	fmt.Printf("foo.bar: [%v]\n", config.Foo.Bar)
}
```

The above code I expected to get the output is:

```
foo.bar: [hello-world]
```

This has worked very well in the past. But since PR #673, it can't work properly, it will output:

```
foo.bar: []
```

I checked the contents of the PR #673 , I can't fully understand the changes. But I believe the changes blocked my previous usage.

How should I modify my code to fit the new version of viper?

inkychris:
Fixes #764 - Regression introduced by #673

The `keyComponents()` function was using the `Get()` method in a way similar to how `IsSet()` works. Neither of these were returning in the expected way for partial keys when only pflags had been bound.

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -1789,14 +1789,35 @@ outer:
 	return shadow
 }
 
+// Checks to see if a key may exist as a parent node.
+// Similar to v.IsSet(), but checks pflags as a fallback.
+func (v *Viper) isComponent(key string) bool {
+	lcaseKey := strings.ToLower(key)
+	val := v.find(lcaseKey)
+	if val != nil {
+		return true
+	}
+	for k := range v.pflags {
+		if strings.HasPrefix(k, lcaseKey) {
+			return true
+		}
+	}
+	for k := range v.env {
+		if strings.HasPrefix(k, lcaseKey) {
+			return true
+		}
+	}
+	return false
+}
+
 // Converts a fully qualified map key into a list of relative
 // map keys, allowing for keys to contain the delimiter themselves
 func keyComponents(v *Viper, key string) []string {
 	var result []string
 	components := strings.Split(key, v.keyDelim)
-	for index := 0; index < len(components); index++ {
+	for index := 1; index < len(components); index++ {
 		potentialKey := strings.Join(components[0:index], v.keyDelim)
-		if v.Get(potentialKey) != nil {
+		if v.isComponent(potentialKey) {
 			result = append(result, potentialKey)
 		}
 	}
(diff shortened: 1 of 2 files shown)


sagikazarmark:
Checking pflag is not enough IMHO. The same issue will cause `Set` and `SetDefault` not working. For that reason I think we will have to revert that PR and implement key delimiter instead.

inkychris:
I'm not seeing the same behaviour for `Set` and `SetDefault` because they create an interface for `"foo"` which is then found by `Get` or `IsSet`. The problem before was that while these were returning true, the pflags, and as I just suspected env, were returning `nil` for `"foo"` instead, hence the new modified version of `IsSet`.

sagikazarmark:
Indeed, you are right. :)

Here is a breaking test though:

```
func TestUnmarshalOnlyPFlagSet2(t *testing.T) {
	flags := pflag.NewFlagSet("test", pflag.ContinueOnError)
	flags.String("foo.bar", "cobra_flag", "")

	v := New()
	assert.NoError(t, v.BindPFlags(flags))

	v.SetConfigType("yaml")
	v.ReadConfig(bytes.NewBuffer([]byte(`foo.bar: file_content`)))

	config := &struct {
		Foo struct {
			Bar string
		}
	}{}

	assert.NoError(t, v.Unmarshal(config))
	assert.Equal(t, "cobra_flag", config.Foo.Bar)
	t.Log(v.AllSettings())
}
```

inkychris:
Yes, I'd noticed that. This change is at least consistent with pre #673 behviour but I guess if we also want a solution to the config override then reverting #673 and implementing a key delimter setter is the only viable way forward.

sagikazarmark:
I'm afraid that any breaking change will cause trouble since this library is widely used. I hope we can clear things up a bit in a v2, but for now, I think reverting and allowing to set the key delimiter is a better option.

PR #768, opened 2019-09-18
Fix Dropped Errors

alrs:
A `:=` assignment inside an if-block caused two `error` variables to be shadowed. Here is a fix.

### DIFF
--- a/remote/remote.go
+++ b/remote/remote.go
@@ -77,11 +77,12 @@ func getConfigManager(rp viper.RemoteProvider) (crypt.ConfigManager, error) {
 	var err error
 
 	if rp.SecretKeyring() != "" {
-		kr, err := os.Open(rp.SecretKeyring())
-		defer kr.Close()
+		var kr *os.File
+		kr, err = os.Open(rp.SecretKeyring())
 		if err != nil {
 			return nil, err
 		}
+		defer kr.Close()
 		if rp.Provider() == "etcd" {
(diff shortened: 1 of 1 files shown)


remote/remote.go
```diff
@@ -75,13 +75,14 @@ func (rc remoteConfigProvider) WatchChannel(rp viper.RemoteProvider) (<-chan *vi
 func getConfigManager(rp viper.RemoteProvider) (crypt.ConfigManager, error) {
 	var cm crypt.ConfigManager
 	var err error
+	var kr *os.File
```
sagikazarmark:
```suggestion
```

remote/remote.go
```diff
@@ -75,13 +75,14 @@ func (rc remoteConfigProvider) WatchChannel(rp viper.RemoteProvider) (<-chan *vi
...
 
 	if rp.SecretKeyring() != "" {
-		kr, err := os.Open(rp.SecretKeyring())
-		defer kr.Close()
+		kr, err = os.Open(rp.SecretKeyring())
```
sagikazarmark:
```suggestion
		var kr *os.File
		kr, err = os.Open(rp.SecretKeyring())
```

alrs: @sagikazarmark If I read your suggestion correctly, the only change was to move the declaration of the `kr` variable, which I've done and squashed into the original commit.

sagikazarmark: @alrs that's correct, thank you!

PR #771, opened 2019-09-27
Revert unmarshal keys containing dots

### LINKED ISSUE #764: PR #673 broke my code. The nested structures don't get the default values correctly

sagikazarmark:
This PR reverts #673 because it introduced a breaking change.

Fixes #764

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -1789,24 +1789,6 @@ outer:
 	return shadow
 }
 
-// Converts a fully qualified map key into a list of relative
-// map keys, allowing for keys to contain the delimiter themselves
-func keyComponents(v *Viper, key string) []string {
-	var result []string
-	components := strings.Split(key, v.keyDelim)
-	for index := 0; index < len(components); index++ {
-		potentialKey := strings.Join(components[0:index], v.keyDelim)
-		if v.Get(potentialKey) != nil {
-			result = append(result, potentialKey)
-		}
-	}
-	result = append(result, key)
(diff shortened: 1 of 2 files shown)


sagikazarmark:
@inkychris would you mind taking a quick look at this PR? 🙂

inkychris:
It looks good to me!

sagikazarmark:
Thanks!

PR #773, opened 2019-09-27
Add Go 1.13

sagikazarmark:


sagikazarmark:
@spf13 what do you think about switching to GitHub Actions for building Viper? In my experience it's much faster than Travis.

PR #805, opened 2019-12-06
Improve tooling

sagikazarmark:
Improve project tooling

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
+/bin/
+/build/
+
 # Compiled Object files, Static and Dynamic libs (Shared Objects)
 *.o
 *.a
--- a/.gitignore
+++ b/.gitignore
@@ -1,32 +1,20 @@
 /bin/
 /build/
+/var/
+/vendor/
 
-# Compiled Object files, Static and Dynamic libs (Shared Objects)
-*.o
-*.a
-*.so
-
-# Folders
-_obj
-_test
-
-# Architecture specific extensions/prefixes
-*.[568vq]
-[568vq].out
-
-*.cgo1.go
-*.cgo2.c
-_cgo_defun.c
-_cgo_gotypes.go
-_cgo_export.*
-
-_testmain.go
-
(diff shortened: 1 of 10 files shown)


aeneasr:
Why was `.idea` commited here? This gives nasty merge conflicts and other people's IDE configs do not belong into a source repo. If certain things should be enforced, we can use a linter, but no one should mess with my IDE workspace config.

Can this please be reverted?

sagikazarmark:
If you look closely, you will see that workspace config is not actually committed. Only the files that are mentioned in the relevant [documentation](https://intellij-support.jetbrains.com/hc/en-us/articles/206544839). It won't affect your workspace config, it will give you a few project specific settings though (like go imports settings, test tasks and warn you if a plugin is recommended to be installed).

I don't think it needs to be reverted.

aeneasr:
It still caues merge conflicts if you have other settings in your workspace. These files do not belong in a repository and there are no open source projects that I know of that do this.

sagikazarmark:
I'm not really sure how they can cause merge conflicts. They are project specific files. You probably shouldn't change them. Or if there is a reason to change them, a PR should be submitted.

This is the official recommendation from JetBrains to share project specific settings. The fact that other projects don't do it does not sound like an argument to me. 🙂 

I always commit them and never had any merge conflicts before.

aeneasr:
I use very specific workspace settings which I like to use, and I got a merge conflict when rebasing some commits, which is just annoying to deal with (because XML).

Other projects don't use it because it's not needed and because it's annoying. If you want to enforce a certain style, use a linter and a formatter that do what you expect the code to look like, but don't mess with my IDE settings. There is not a single Go project I have ever encountered that enforces IDE settings.

Sharing IDE settings is cool for small personal projects or enterprise where your boss kinda hates you, but not for open source projects with many contributors ;)

Go, Kubernetes, Helm, Skaffold, Moby, Traefik - you name it. You will not find a single project that does this!

PR #813, opened 2019-12-15
Add support to save file with no extension

gssbzn:
The suppport introduced for files with no file extension is only partial as trying to save the config file would fail with `<file name> equires valid extensio`
This adds support to saving such files

### DIFF
--- a/viper_test.go
+++ b/viper_test.go
@@ -1311,64 +1311,97 @@ func TestWriteConfig(t *testing.T) {
 	fs := afero.NewMemMapFs()
 	testCases := map[string]struct {
 		configName      string
-		configType      string
+		inConfigType    string
+		outConfigType   string
 		fileName        string
 		input           []byte
 		expectedContent []byte
 	}{
 		"hcl with file extension": {
 			configName:      "c",
-			configType:      "hcl",
+			inConfigType:    "hcl",
+			outConfigType:   "hcl",
 			fileName:        "c.hcl",
 			input:           hclExample,
 			expectedContent: hclWriteExpected,
 		},
 		"hcl without file extension": {
 			configName:      "c",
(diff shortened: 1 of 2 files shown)


gssbzn:
@sagikazarmark any chance to give this a look? it would be great to fully support config files with no extensions

viper_test.go
```diff
@@ -1436,24 +1307,203 @@ hobbies:
...
-	v := New()
+func TestWriteConfig(t *testing.T) {
 	fs := afero.NewMemMapFs()
-	v.SetFs(fs)
-	v.SetConfigName("c")
-	v.SetConfigType("yaml")
-	err := v.ReadConfig(bytes.NewBuffer(yamlExample))
-	if err != nil {
-		t.Fatal(err)
+	testCases := map[string]struct {
```
sagikazarmark: Nice refactor!

Currently, the config type is always inferred from the file name, not the original config type. So if the config type was yaml, but I provided a json file name, the output content was in JSON.

Can you please add a test case that verifies this is still the case?

Because (unfortunately) as far as I can tell, it is not.

gssbzn: Just making sure I got the right idea here, are you saying to default on `writeConfig` to the extension file and if this is blank then use the setting for config type?

sagikazarmark: Yup. At least, that's compatible with the current behaviour, taking the original configType by default is not.

gssbzn: Gotcha 👍

gssbzn:
@sagikazarmark I added the requested changes, this complicated the test a bit as the read depends always of the `configType` and I don't know if we should keep the error of undefined type or just relay on the extensions check

gssbzn:
@sagikazarmark any update on this PR?

Issue #818, opened 2019-12-25
^1.6: Unsupported Config Type "" when config file not found and falling back to extension-less filename


andig:
1.6.0 has introduced a breaking change that will lead to failure if not config file is found. I used to do this:

	// If a config file is found, read it in
	if err := viper.ReadInConfig(); err == nil {
		// using config file
		cfgFile = viper.ConfigFileUsed()
	} else if _, ok := err.(viper.ConfigFileNotFoundError); !ok {
		// parsing failed - exit
		fmt.Println(err)
		os.Exit(1)
	} else {
		// not using config file
		cfgFile = ""
	}

Starting with 1.6.0, `viper.ReadInConfig` will return 

    Unsupported Config Type ""

if the config file does not exist. This is highly confusing as it should really be `ConfigFileNotFoundError` instead of `UnsupportedConfigError`. Please advise.

sagikazarmark:
@andig unfortunately I cannot reproduce the issue with your code sample.

I tried to simplify it a little bit:

```
package main

import (
	"fmt"

	"github.com/spf13/viper"
)

func main() {
	viper.AddConfigPath(".")
	viper.SetConfigName("config")

	// If a config file is found, read it in
	if err := viper.ReadInConfig(); err != nil {
		fmt.Println(err)
	}
}
```

But I'm unable to reproduce it with 1.6.0 and 1.6.1.

There was indeed a breaking change introduced in 1.6.0, which the above code sample causes to return `Unsupported Config Type`: when there is a directory with the same name (eg. `config` is the config name and there is a directory with that name). But that was fixed in 1.6.1.

Please try to provide a breaking test, or a simple code sample that reproduces the issue. It might have environmental components (eg. existing file/directory, envorinment variable, etc), so please take those into account as well (eg. what config file are you trying to read? is there a directory with the same name?, etc).

sagikazarmark:
@andig any chance you can provide an example reproducing this bug?

andig:
Will do but need some time for debugging to find the faulty case.

andig:
Closing this one. I have the sneaking suspicion that I had a cobra `PersistentFlags().StringVarP` in somewhere for an unrelated option pointing to the config file setting. Please forgive me :/

sagikazarmark:
No worries :)

andig:
I've finally found what happens and why I didn't see it every time:

[...]

PR #827, opened 2020-01-12
Extensionless files only allowed when config type is set

pedromss:
This PR addresses #818 

cc @andig

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -1976,7 +1976,7 @@ func (v *Viper) searchInPath(in string) (filename string) {
 		}
 	}
 
-	if b, _ := exists(v.fs, filepath.Join(in, v.configName)); b {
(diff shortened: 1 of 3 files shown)


viper.go
```diff
@@ -1976,7 +1976,7 @@ func (v *Viper) searchInPath(in string) (filename string) {
 		}
 	}
 
-	if b, _ := exists(v.fs, filepath.Join(in, v.configName)); b {
+	if b, _ := exists(v.fs, filepath.Join(in, v.configName)); b && v.configType != "" {
```
andig: It's a minor point for performance-wise I'd do the single string check first. Otherwise LGTM.

andig:
@sagikazarmark could we get this in and preferably release to unbreak 1.6 for this use case?

Issue #742, opened 2019-08-02
WatchConfig and WriteConfig are conflict


ywh254:
```
viper.WatchConfig()
viper.OnConfigChange(func(e fsnotify.Event) {
	log.Logger().Printf("Config file changed: %s", e.Name)
})
```
```
if err := viper.WriteConfig(); err != nil {
	return err
}
```
WatchConfig and WriteConfig are conflict. WatchConfig and WriteConfig concurrent execute， this cause writing a not correct config file content.

muXxer:
Was the fix in #834 ever merged or this issue fixed?

PR #577, opened 2018-10-10
Implement CancelWatchConfig which allows one to turn off the watcher

deinspanjer:

PR #834, opened 2020-01-25
Fix dangling WatchConfig goroutine, add viper.CancelWatchConfig()

7d4b9:
* properly handle the WatchConfig goroutines workflow
* add viper.CancelWatchConfig()
*fixes the concurrency issue #742 between `WatchConfig()`and `WriteConfig()`
* added tests from #577

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -996,11 +996,6 @@ func (v *Viper) BindFlagValues(flags FlagValueSet) (err error) {
 }
 
 // BindFlagValue binds a specific key to a FlagValue.
-// Example (where serverCmd is a Cobra instance):
-//
-//	 serverCmd.Flags().Int("port", 1138, "Port to run Application server on")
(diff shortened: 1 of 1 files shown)


7d4b9:
@sagikazarmark any chance to give this a look ?
* **refactor WatchConfig**: use only one cancellable goroutine: The current code make use of two nested goroutine with subtil intrication (the surrounding goroutine does not wait for the inner goroutine (`init.WG.Done()` called before `eventWG().Wait()`), making it dangling. The inner goroutine was not stoppable and the surrounding goroutine had to wait for an "init phase" of the inner one before returning, thus making requirement of 2 waitGroup + 2 goroutines "complex"...  
This have been breakdown to one _cancellable_ goroutine and make the code clearer imho.
  *  also fixes the concurrency issue #742 between `WatchConfig()`and `WriteConfig()`
* added tests from #577

PR #835, opened 2020-01-25
WatchConfig: Remove unused eventsWG waitgroup

7d4b9:
eventsWG.Wait() is called after initWG.Done(), so the inner goroutine is dangling and cannot be waited by the surrounding one.
So remove the useless waitGroup


7d4b9:
Done with #834
Properly handle asynchronous WatchConfig goroutines

Issue #844, opened 2020-02-18
Should find config file which with extension first.


SmallTianTian:
```go
package main

import (
	"fmt"
	"os"

	"github.com/spf13/viper"
)

func main() {
	viper.SetConfigName("main")
	viper.SetConfigType("yaml")
	viper.AddConfigPath(".")
	viper.AddConfigPath("/tmp")

	if err := viper.ReadInConfig(); err != nil {
		fmt.Println(err)
		os.Exit(1)
	}

	fmt.Println("File path:", viper.ConfigFileUsed())
}
```

When it's a `go` file, it can read `main.yaml` for config file. But when you build this go file(`main`), you will find the executable file will be reading first. I'm set config type and config name, but I couldn't get this file.
I've read #818, extension-less filename is neeeded. So, I think the extension config file should be high weight.

wfernandes:
I'm running into a similar issue.
I have the file `~/.myconfdir/myconf.yaml`
When I do 
```
viper.SetConfigName("myconf")
viper.AddConfigPath("~/.myconfdir")
```
And then call `viper.ConfigFileUsed()` I get an empty string because it doesn't set `v.configFile`.

I propose that the following suggestion
```
func (v *Viper) ConfigFileUsed() string {
  file, _ := v.getConfigFile()
  return file
}
```

The current implementation is 
```
func (v *Viper) ConfigFileUsed() string {
  return v.configFile
}
```

winder:
I noticed this issue today as well, but don't really like the proposed solution. Some users may not have their config file be the same as the binary, in which case it's completely reasonable to want/expect a config file without an extension in the CWD/bindir to override some system config.

Another way to do this is to ignore executable files. There isn't a great way to do this in Windows, but on that system the binary generally has the `.exe`  extension so I think this is a general solution:
```diff
  func exists(fs afero.Fs, path string) (bool, error) {
  	stat, err := fs.Stat(path)
  	if err == nil {
- 		return stat.IsDir(), nil
+ 		if stat.IsDir() {
+ 			return false, nil
+ 		}
+ 		return stat.Mode().Perm()&0111 != 0, nil
  	}
  	if os.IsNotExist(err) {
  		return false, nil
  	}
  	return false, err
  }
```

bearsh:
another way would be to compare it against the path of the application's binary

drsybren:
Thanks @SmallTianTian for writing this report. I also had issues with the way Viper tries to find a suitable config file to load.

[...]

PR #845, opened 2020-02-18
find config file which with extension first.

SmallTianTian:
for #844.


sagikazarmark:
I'm not sure this is gonna work for everyone. I would assume, that any configuration file takes precedence within a single location, before moving onto another.

SmallTianTian:
In current version, you only need to do three steps, and you can replicated 100%.
### step 1: Use the same name for `config` and `project`, and set config type.   
EX.   
```go
viper.SetConfigName("test") // the `test` is the project name
viper.SetConfigType("yaml") // must set config type.
```
### step 2: Add current path first.
EX.
```go
viper.AddConfigPath(".") // this statement must come first
viper.AddConfigPath(os.Getenv("CONFIG_PATH"))
```
### step 3: Build this project just by `go build`, don't use `-o`, you will get a executable file with the same name as the config(extensions are not included).

Run the executable and you'll see `viper` will read this executable, not true config.

viper.go
```diff
@@ -1971,12 +1971,6 @@ func (v *Viper) searchInPath(in string) (filename string) {
 		}
 	}
 
-	if v.configType != "" {
-		if b, _ := exists(v.fs, filepath.Join(in, v.configName)); b {
```
SmallTianTian: When the program starts looking for the current directory, `v.configType` not `""`. The executable file is same as `v.configName`, `b` is `true`.

PR #847, opened 2020-02-19
Remove .idea directory

aeneasr:
The `.idea` directory does not belong into the viper project. Forcefully overwriting that in `.gitignore` doesn't make sense (it overrides my global gitignore for example).

This just leads to someone messing with my IDE settings, and causes merge conflicts because we might be using different IDE versions.

If standards should be enforced, this must not be done on a per-IDE base but with golangci-lint or something like that.


sagikazarmark:
My answer from #805:

If you look closely, you will see that workspace config is not actually committed. Only the files that are mentioned in the relevant [documentation](https://intellij-support.jetbrains.com/hc/en-us/articles/206544839). It won't affect your workspace config, it will give you a few project specific settings though (like go imports settings, test tasks and warn you if a plugin is recommended to be installed).

I don't think it needs to be reverted.

aeneasr:
I strongly disagree with your conclusion, as commented in the other issue.

sagikazarmark:
Let's agree to disagree then. 🙂 

As I said, I never had any conflicts before, but if that happens I'm happy to reevaluate my position. So if you had any conflicts in Viper, please provide some details, so i can check.

Again, this is what JetBrains recommends to commit. Nothing more.

sagikazarmark:
Let's keep the discussion here.

> I use very specific workspace settings which I like to use, and I got a merge conflict when rebasing some commits, which is just annoying to deal with (because XML).
> 
> Other projects don't use it because it's not needed and because it's annoying. If you want to enforce a certain style, use a linter and a formatter that do what you expect the code to look like, but don't mess with my IDE settings. There is not a single Go project I have ever encountered that enforces IDE settings.
> 
> Sharing IDE settings is cool for small personal projects or enterprise where your boss kinda hates you, but not for open source projects with many contributors ;)
> 
> Go, Kubernetes, Helm, Skaffold, Moby, Traefik, all of Ory - you name it. You will not find a single project that does this!

This is just ranting and it's not helpful at all. 🙁 You disagree, I get that.

I'd much rather take a look at the actual issue, so if you have a diff or something generated by your environment that conflicts with the latest master, that would be helpful.

sagikazarmark:
Thanks for your answer!

Indeed, in case of forks this might cause some problems. Especially in case of Viper, since it's quite undermaintained, so the number of forks is probably large.

Well, this looks like a real issue to me and I don't have a better answer than removing those files for now, they are mostly there for convenience anyway.

aeneasr:
Thank you, I appreciate that! I'm not sure if viper is now better maintained (it looks like it?) but we've had several performance issues with reads which we have/are fixing in our fork and also some tooling around reloading and changed values. If there's a chance to merge this back into upstream let me know, I'd be really happy to get these things back into the original project!

sagikazarmark:
Definitely open to performance improvements. As for reloading values: interested in that as well, but I'm currently working on what could go into Viper 2 and new features might fit that better.

aeneasr:
Makes sense, I'll cherrypick the perf improvements only then :)

PR #848, opened 2020-02-19
Add GetStringPtr

satyap:
Returns a string pointer. Returns `nil` if GetString() would have returned
`""`. If you want a pointer to the empty string, assign the string returned by
GetString and take its pointer:

```golang
v = viper.GetString("key")
ptr := &v
```

I have a handful of projects that need the config value as a pointer, but if it's not set or if it's set as empty string, the pointer needs to be nil (there are external APIs that want it that way). Got tired of funcs like `utils.NilForEmptyString(viper.GetString("some.key"))`

### DIFF
--- a/README.md
+++ b/README.md
@@ -507,6 +507,7 @@ The following functions and methods exist:
  * `GetInt(key string) : int`
  * `GetIntSlice(key string) : []int`
  * `GetString(key string) : string`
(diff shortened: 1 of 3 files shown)


sagikazarmark:
Thanks for sending a PR, but we won't accept it for now. See more details in #1112

PR #852, opened 2020-02-25
Add example regarding required flags with Cobra + Viper

migueleliasweb:


edwarnicke:
Perhaps I'm doing this wrong, but this does not seem to work for me:
```
cobra.OnInitialize(func(){
		viper.AutomaticEnv()
		viper.SetEnvPrefix("NSM")
		viper.SetEnvKeyReplacer(strings.NewReplacer("-","_"))
		viper.BindPFlags(cmd.Flags())
		cmd.Flags().VisitAll(func(f *pflag.Flag) {
			logrus.Infof("Trying to set %s from env variable",f.Name)
			if viper.IsSet(f.Name) && viper.GetString(f.Name) != "" {
				logrus.Infof("Setting %s from env variable",f.Name)
				cmd.Flags().Set(f.Name, viper.GetString(f.Name))
			}
			cmd.MarkFlagRequired(f.Name)
		})
	})
```

results in a log:
```
INFO[0000] Trying to set tunnel-ip from env variable    
INFO[0000] Setting tunnel-ip from env variable 
```

which tells me that I should be properly setting that (meaning I've avoided issues like failing to set the right ENV variable).  But yet I still get the complaint on running:

```
Error: required flag(s) "base-dir", "connect-to-url", "help", "listen-on-url", "name", "tunnel-ip", "vppagent-url" not set
```

I *definitely* have missing required parameters, but 'tunnel-ip' should not be among them.

Is there something *obvious* I've borked here?

migueleliasweb:
Hi @edwarnicke . 

The only obvious difference to the working example I have on my computer is the fact you set the `cmd.MarkFlagRequired(f.Name)` after setting the value.

I try to do all cobra related config in the `func init(){}` of my module (including setting the required flag). In the end of the init function, I call `cobra.OnInitialize(viperExtraInitConfig)` then I run pretty much what is set in the code example on the PR.

edwarnicke:
OK... beat this into a shape very close to @v-braun 's example and it works now :)

edwarnicke:
Would it make sense to include proper error reporting in the docs here as well?

```
cmd.Flags().VisitAll(func(f *pflag.Flag) {
		if viper.IsSet(f.Name) && viper.GetString(f.Name) != "" {
			err := cmd.Flags().Set(f.Name, viper.GetString(f.Name))
			if err != nil {
				// If root command has SilentErrors flagged,
				// all subcommands should respect it
				if !cmd.SilenceErrors && !cmd.Root().SilenceErrors{
					cmd.Println("Error:", err.Error())
				}

				// If root command has SilentUsage flagged,
				// all subcommands should respect it
				if !cmd.SilenceUsage && !cmd.Root().SilenceUsage {
					cmd.Println(cmd.UsageString())
				}
			}
		}
	})
```

README.md
```diff
@@ -282,6 +282,16 @@ os.Setenv("SPF_ID", "13") // typically done outside of the app
 id := Get("id") // 13
 ```
 
+#### When working with *required flags with Cobra+Viper*, you must set the flag values after binding to them. Otherwise Cobra won't be aware of the value from any source other tha the CLI itself.
```
mitar, line 285: Typo: `tha` -> `than`.

nicop311:
Hey @migueleliasweb ,

I tried your suggestion on my project https://github.com/nicop311/cobravsviper/commit/b544d936be08d6daead02a2e35aa05e3c15f177e: it seem to work! But I did not do much testing.

It seem your fix to viper allow viper to support both cobra's `MarkFlagsOneRequired` and `MarkFlagsMutuallyExclusive`.

Notice on my project, I have my own fix for viper to support nested cobra subcommand better (when binding cobra with viper).

Issue #196, opened 2016-06-06
Addresses of Items in Slice with `Get()`


jakdept:
Let's say I have a json config that looks like the following, and I load it in:

``` json
{
  "slice": [
    {
      "name": "apple",
      "type": "fruit",
      "serving": 1
    },
    {
      "name": "banana",
      "type": "fruit",
      "serving": 1
    },
    {
      "name": "grape",
      "type": "fruit",
      "serving": 6
    },
    {
      "name": "carrot",
      "type": "veggie",
      "serving": 1
    },
    {
      "name": "broccoli",
      "type": "veggie",
      "serving": 1
    }
  ]
}
```

How, using `Get()`, would I access `slice.2.serving` - the serving size for grapes? I would expect that, but I honestly don't know. Is the same string used for `GetInt()` to access that same location? I didn't see any tests that looked to be referencing an index either.

I've looked through the docs, and I don't really see where it's specified how to get at items within a slice - I saw a lot of named stuff, lots of `map[string]interface{}`, but no slices.

torbensky:
I just ran into this as well. It really surprised me, but it looks like Viper doesn't handle lists of objects. You have to call `Get` and then use reflection to cast it to a slice of interfaces. At least I think so, I hope I'm wrong because this is a really annoying thing to code around.

torbensky:
The title of this issue is confusing though. I still don't really understand it. IMO the issue title would be more understandable to others if it was called something like:

> Support getting list of objects from config
or
> Add GetMapSlice function

torbensky:
Given the YAML:
```
foos:
  - name: foo1
    val: 1
  - name: foo2
    val: 2
  - name: foo3
    val: 3
  - name: foo4
    val: 4
```

Looking at the type returned by `Get`
```
fmt.Printf("%T\n", viper.Get("foos"))
```
Outputs:
> []interface {}

torbensky:
Ah-hah, I figured this was too big of a hole in the library to not have a solution. There is a solution!

```
foos:
  - name: foo1
    val: 1
  - name: foo2
    val: 2
  - name: foo3
    val: 3
  - name: foo4
    val: 4
```

can be accessed like so:

```
var foos []struct{
  Name string
  Val int
}
viper.UnmarshalKey("foos", &foos)
fmt.Printf("%v\n", ps) // voila!
```

torbensky:
[...]

PR #861, opened 2020-03-05
Added support for accessing slices

dylandreimerink:
Hi,
I added support for accessing values within a slice as mentioned in #196. For example, if we have a config file which is structured like this:
```json
{
  "server_name": "example.com",
  "interfaces": [
    {
        "interface": "192.168.1.1",
        "port": 80
    },
    {
        "interface": "127.0.0.1",
        "port": 8080
    }
  ]
}
```

We can now get the second interface config like so `viper.Get("interfaces.1")` or specifically the port like so `viper.GetInt("interfaces.1.port")`

I have chosen to implement the index numbers like path parts since it is way more difficult to try to parse the go like syntax `viper.Get("interfaces[1]")` and still respect the other functionality like shadowing checks.

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -583,7 +583,7 @@ func (v *Viper) searchMap(source map[string]interface{}, path []string) interfac
 	return nil
 }
 
-// searchMapWithPathPrefixes recursively searches for a value for path in source map/slice.
+// searchIndexableWithPathPrefixes recursively searches for a value for path in source map/slice.
 //
 // While searchMap() considers each path element as a single map key or slice index, this
 // function searches for, and prioritizes, merged path elements.
@@ -594,7 +594,7 @@ func (v *Viper) searchMap(source map[string]interface{}, path []string) interfac
 // in their keys).
 //
 // Note: This assumes that the path entries and map keys are lower cased.
-func (v *Viper) searchMapWithPathPrefixes(source interface{}, path []string) interface{} {
+func (v *Viper) searchIndexableWithPathPrefixes(source interface{}, path []string) interface{} {
 	if len(path) == 0 {
 		return source
 	}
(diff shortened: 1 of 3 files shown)


snail007:
bro, what's next plan about this feature ?

dylandreimerink:
Well, this is one of 102 open pull requests at the moment, it most likely is not on anyone's radar. I think @sagikazarmark should take a look at it and give some feedback or close the pull request. Until then I can't really do much as far as I know.

viper.go
```diff
@@ -588,36 +589,58 @@ func (v *Viper) searchMap(source map[string]interface{}, path []string) interfac
...
-func (v *Viper) searchMapWithPathPrefixes(source map[string]interface{}, path []string) interface{} {
+func (v *Viper) searchMapWithPathPrefixes(source interface{}, path []string) interface{} {
 	if len(path) == 0 {
 		return source
 	}
 
 	// search for path prefixes, starting from the longest one
 	for i := len(path); i > 0; i-- {
 		prefixKey := strings.ToLower(strings.Join(path[0:i], v.keyDelim))
+		if sourceSlice, ok := source.([]interface{}); ok {
```
sagikazarmark, line 600: How about making this a type switch instead?

I'd also consider moving the bodies of these conditionals to functions like `searchSliceWithPathPrefixes` and `searchMapWithPathPrefixes`, so that this function is a bit shorter and better separated.

viper.go
```diff
@@ -588,36 +589,58 @@ func (v *Viper) searchMap(source map[string]interface{}, path []string) interfac
...
 
-		next, ok := source[prefixKey]
-		if ok {
-			// Fast path
-			if i == len(path) {
-				return next
+				var val interface{}
+				switch n := next.(type) {
+				case map[interface{}]interface{}:
+					val = v.searchMapWithPathPrefixes(cast.ToStringMap(next), path[i:])
```
sagikazarmark, line 613:
```suggestion
					val = v.searchMapWithPathPrefixes(cast.ToStringMap(n), path[i:])
```

viper.go
```diff
@@ -588,36 +589,58 @@ func (v *Viper) searchMap(source map[string]interface{}, path []string) interfac
...
-			default:
-				// got a value but nested key expected, do nothing and look for next prefix
-			}
-			if val != nil {
-				return val
+				// Nested case
+				var val interface{}
+				switch n := next.(type) {
+				case map[interface{}]interface{}:
+					val = v.searchMapWithPathPrefixes(cast.ToStringMap(next), path[i:])
```
sagikazarmark, line 635:
```suggestion
					val = v.searchMapWithPathPrefixes(cast.ToStringMap(n), path[i:])
```

viper.go
```diff
@@ -588,36 +589,58 @@ func (v *Viper) searchMap(source map[string]interface{}, path []string) interfac
...
+func (v *Viper) searchMapWithPathPrefixes(source interface{}, path []string) interface{} {
 	if len(path) == 0 {
 		return source
 	}
 
 	// search for path prefixes, starting from the longest one
 	for i := len(path); i > 0; i-- {
 		prefixKey := strings.ToLower(strings.Join(path[0:i], v.keyDelim))
+		if sourceSlice, ok := source.([]interface{}); ok {
+			//if the prefixKey is a number which is not out of bounds of the slice
```
sagikazarmark, line 601:
```suggestion
			// if the prefixKey is a number which is not out of bounds of the slice
```

viper.go
```diff
@@ -588,36 +589,58 @@ func (v *Viper) searchMap(source map[string]interface{}, path []string) interfac
...
+			if index, err := strconv.Atoi(prefixKey); err == nil && len(sourceSlice) > index {
```
sagikazarmark, line 602: By inverting this condition and returning early, we can avoid the arrow pattern.

viper.go
```diff
@@ -588,36 +589,58 @@ func (v *Viper) searchMap(source map[string]interface{}, path []string) interfac
...
+				default:
+					// got a value but nested key expected, do nothing and look for next prefix
+				}
+				if val != nil {
+					return val
+				}
 			}
+		} else if sourceMap, ok := source.(map[string]interface{}); ok {
+			next, ok := sourceMap[prefixKey]
+			if ok {
```
sagikazarmark, line 625: See my above comment about the arrow pattern.

sagikazarmark (changes requested):
Thanks @dylandreimerink

I reviewed your PR and had a few comments.

viper.go
```diff
@@ -602,35 +603,100 @@ func (v *Viper) searchMapWithPathPrefixes(source map[string]interface{}, path []
...
+	}
+
+	var val interface{}
+	switch n := next.(type) {
+	case map[interface{}]interface{}:
+		val = v.searchIndexableWithPathPrefixes(cast.ToStringMap(n), path[pathIndex:])
+	case map[string]interface{}, []interface{}:
+		val = v.searchIndexableWithPathPrefixes(n, path[pathIndex:])
+	default:
+		// got a value but nested key expected, do nothing and look for next prefix
```
sagikazarmark, line 652: I think you can just return the values here.

viper.go
```diff
@@ -602,35 +603,100 @@ func (v *Viper) searchMapWithPathPrefixes(source map[string]interface{}, path []
...
+
+	// Nested case
+	var val interface{}
+	switch n := next.(type) {
+	case map[interface{}]interface{}:
+		val = v.searchIndexableWithPathPrefixes(cast.ToStringMap(n), path[pathIndex:])
+	case map[string]interface{}, []interface{}:
+		val = v.searchIndexableWithPathPrefixes(n, path[pathIndex:])
+	default:
+		// got a value but nested key expected, do nothing and look for next prefix
```
sagikazarmark, line 690: I think you can just return the values here.

sagikazarmark (changes requested):
Thanks @dylandreimerink 

One final thing then it's good to go.

dylandreimerink:
You are correct, missed that after the refactor. Should be all good now.

snail007:
hi bro , do you consider for sets value for slices, i hava a toml config file bellow, in go code i want to set 
through such as `Set("cache.memory[0].enable",true)`, i think this feature is a basic requirement.

```ini
[cache]
default="redis"

[[cache.redis]]
debug=true
enable=true
id="default"
address="127.0.0.1:6379"
prefix=""
password=""
# seconds
timeout=10
dbnum=0
maxidle=10
maxactive=30
# seconds
idletimeout=300
# seconds
maxconnlifetime=3600
wait=true

[[cache.memory]]
enable=false
id="default"
# seconds
cleanupinterval=30
```

sagikazarmark (approved):
Thanks @dylandreimerink

sagikazarmark:
@snail007 you should be able to do

```go
Set("cache.memory", []interface{}{
    map[string]interface{}{
        "enabled": true,
    },
})
```

snail007:
@sagikazarmark thanks

snail007:
> @snail007 you should be able to do
> 
> ```go
> Set("cache.memory", []interface{}{
>     map[string]interface{}{
>         "enabled": true,
>     },
> })
> ```

bro, i tried it, the result not crrect, other keys was lost :
code:

```go
        cfg.SetConfigFile("app.toml")
	cfg.ReadInConfig()
	fmt.Println(cfg.Get("cache"))
	cfg.Set("cache.memory",[]interface{}{
		map[string]interface{}{
			"enable":true,
		},
	})
	fmt.Println(cfg.Get("cache"))
```
output:

```text
map[default:redis memory:[map[cleanupinterval:30 enable:false id:default]] redis:[map[address:127.0.0.1:6379 dbnum:0 debug:true enable:true id:default idletimeout:300 maxactive:30 maxconnlifetime:3600 maxidle:10 password: prefix: timeout:10 wait:true]]]
map[memory:[map[enable:true]]]
```

Issue #858, opened 2020-03-04
`WriteConfig` panics for extension `ini`


dionysius:
Consider this example: https://play.golang.org/p/MHL5QwfsKmx

Copy this example locally, as you can't write files in playground AFAIK, and you can't write the config to a writer #856 

Running this file results into the following panic:
```bash
$ go run test.go 
panic: runtime error: slice bounds out of range [:-1]

goroutine 1 [running]:
github.com/spf13/viper.(*Viper).marshalWriter(0xc00011c900, 0x9ada80, 0xc000010098, 0xc000024671, 0x3, 0x9ada80, 0xc000010098)
        /home/user/go/pkg/mod/github.com/spf13/viper@v1.6.2/viper.go:1617 +0x1254
github.com/spf13/viper.(*Viper).writeConfig(0xc00011c900, 0xc000024660, 0x14, 0xc000024601, 0x0, 0x0)
        /home/user/go/pkg/mod/github.com/spf13/viper@v1.6.2/viper.go:1442 +0x1ed
github.com/spf13/viper.(*Viper).WriteConfigAs(...)
        /home/user/go/pkg/mod/github.com/spf13/viper@v1.6.2/viper.go:1406
github.com/spf13/viper.WriteConfigAs(...)
        /home/user/go/pkg/mod/github.com/spf13/viper@v1.6.2/viper.go:1404
main.glob..func2(0xcaaca0, 0xc0000212d0, 0x1, 0x1, 0x0, 0x0)
        /home/user/repos/repo/test.go:26 +0x216
github.com/spf13/cobra.(*Command).execute(0xcaaca0, 0xc000021280, 0x1, 0x1, 0xcaaca0, 0xc000021280)
        /home/user/go/pkg/mod/github.com/spf13/cobra@v0.0.6/command.go:840 +0x453
github.com/spf13/cobra.(*Command).ExecuteC(0xcaaa00, 0xc00000e700, 0x0, 0x0)
        /home/user/go/pkg/mod/github.com/spf13/cobra@v0.0.6/command.go:945 +0x317
main.main()
        /home/user/repos/repo/test.go:56 +0xa2
exit status 2

```

While other extensions work just fine.

gandarez:
Having the same problem here.

W1M0R:
I'm having the same issue.

PR #864, opened 2020-03-09
Fix panic while writing invalid ini

gopherclass:
Each item of .ini consists of section and key. If section is present and key is missing, writing in ini format should return an error. But it panics now.

```go
// This is an error when written in ini format.
viper.Set("only_section", "anything")
```


gopherclass:
The third commit (a925fa4) is just a tiny code improvement. I removed two useless parenthesis. If it is inappropriate, it's totally okay to reject the commit.

gopherclass:
I found this fixes #858.

gandarez:
No intention to merge it?

PR #866, opened 2020-03-12
Make Viper.Sub() inherit values from the bound pflags of the parent

blagojts:
Sub creates a new instance from a sub-tree, but bound flags
are not passed down to the new instance. This commit adds
the bound flags that are prefixed as the provided k to the
new instance.


bogdandrutu:
In the same notes https://github.com/spf13/viper/issues/871

Issue #840, opened 2020-02-10
INI files save error


alex-bacart:
Hi, I have a file `./config/config.ini` with the following content:
```
[param]
key=1
```

Here is my code:
```
func main()  {
	v := viper.New()

	v.SetConfigName("config")
	v.SetConfigType("ini")
	v.AddConfigPath("./config")

	err := v.ReadInConfig()
	if err != nil {
		panic(fmt.Errorf("Fatal error config v: %s \n", err))
	}

	fmt.Println(v.Get("param.key")) // prints 1 as expected

	v.Set("param.key", "123")

	err = v.WriteConfigAs("./config/new.ini")
	if err != nil {
		panic(fmt.Errorf("Fatal error config v: %s \n", err))
	}
}
```

On  execution I get panic:
>panic: interface conversion: interface {} is nil, not string

If I replace `ini` with `yaml` everything works as expected.

PR #868, opened 2020-03-15
fix bug while saving ini file

flow00:
While new Viper with ini file and call method v.WriteConfig or v.WriteConfigAs, the method will return err "interface conversion: interface {} is nil, not string"

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -1621,7 +1621,7 @@ func (v *Viper) marshalWriter(f afero.File, configType string) error {
 			if sectionName == "default" {
 				sectionName = ""
 			}
-			cfg.Section(sectionName).Key(keyName).SetValue(Get(key).(string))
+			cfg.Section(sectionName).Key(keyName).SetValue(v.Get(key).(string))
 		}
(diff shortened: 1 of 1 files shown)


flow00:
I found this fixes #840

freeasbird:
viper.go  line: 1624
```
			kk := cfg.Section(sectionName).Key(keyName)

			var value string
			inter := v.Get(key)
			switch inter.(type) {
			case string:
				value = inter.(string)
				break
			case int:
				value = strconv.Itoa(inter.(int))
				break
			case int64:
				value = strconv.FormatInt(inter.(int64), 10)
				break
			case float64:
				value = strconv.FormatFloat(inter.(float64), 'f', -1, 64)
				break
			case bool:
				value = strconv.FormatBool(inter.(bool))
				break
			}

			kk.SetValue(value)
```
test cases:
	
```
        v.Set("daemon.port", 12346777)
	v.Set("daemon.bool", false)
	v.Set("daemon.flot", 1.56456456456456)
	v.Set("daemon.int64", int64(6546544536))
```

Issue #869, opened 2020-03-16
Unable to set custom viper.StringReplacer using SetEnvKeyReplacer


dan-j:
I'm trying to create a string replacer which will replace both `.` and `-` characters with `_`. I have a custom struct implementing `viper.StringReplacer` which achieves this, however it's not possible to apply this to the singleton `viper` instance.

Note that the signature of `SetEnvKeyReplacer` only accepts `*strings.Replacer`, which is a struct so cannot provide a custom implementation:

https://github.com/spf13/viper/blob/97ee7adfef4882d78c0ef26e22a0c8a7a8bf6776/viper.go#L1196-L1202

Whereas the using the `Option` method and creating a new `Viper` instance would work:

https://github.com/spf13/viper/blob/97ee7adfef4882d78c0ef26e22a0c8a7a8bf6776/viper.go#L267-L271

On first glance, I don't think changing `SetEnvKeyReplace` to use your interface would be a breaking change, could someone confirm?

I'd be happy to provide a PR

PR #870, opened 2020-03-16
allow SetEnvKeyReplacer to take viper.StringReplacer

dan-j:
PR for issue: https://github.com/spf13/viper/issues/869

Signed-off-by: dan-j <5727701+dan-j@users.noreply.github.com>

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -1196,8 +1196,8 @@ func (v *Viper) AutomaticEnv() {
 // SetEnvKeyReplacer sets the strings.Replacer on the viper object
 // Useful for mapping an environmental variable to a key that does
 // not match it.
-func SetEnvKeyReplacer(r *strings.Replacer) { v.SetEnvKeyReplacer(r) }
-func (v *Viper) SetEnvKeyReplacer(r *strings.Replacer) {
+func SetEnvKeyReplacer(r StringReplacer) { v.SetEnvKeyReplacer(r) }
+func (v *Viper) SetEnvKeyReplacer(r StringReplacer) {
 	v.envKeyReplacer = r
 }
 


dan-j:
are there any active maintainers on this project? Not saying this PR is particularly important, but there's a lot of outstanding PRs 🤔

@spf13

sagikazarmark:
Yes, there are. 🙂 

Technically, this PR is a breaking change.

dan-j:
@sagikazarmark, can you explain why it's a breaking change? I'm struggling to find how so.

The `*strings.Replacer` struct implements the `viper.StringReplacer` interface, so no existing consumers of this project should result in and compiler errors or changed behaviour?

Apologies if I'm missing something obvious 🧐

sagikazarmark:
Because of Go's type system, the used types are part of your API, even if technically the same inputs would always work.

If someone were to use Viper through an interface like this:

```go
type MyViper interface {
    SetEnvKeyReplacer(r *strings.Replacer)
}
```

The change in this PR would break their code.

dan-j:
Ah, yes I see now.

Although whilst looking into an alternative solution I've just noticed I can solve my problem using `*strings.Replacer` anyway.

In case anyone else goes through the same thought process as me and ends up here: My problem was I have yaml config with nested keys and I couldn't use env vars to map to a config property like `parent.child-key` because of multiple characters that need converting to `_`, i.e. `PARENT_CHILD_KEY`. So I created a custom replacer, `CompositeReplacer`, which implemented the `viper.StringReplacer` interface which could be given multiple `*string.Replacer` values.

However, I've just noticed I think (not tested it yet) that one can do `strings.NewReplace(".", "_", "-", "_")` which would achieve the same result.

I do think this PR provides a better API to users, so if you're happy to put this into v2 then great, however I won't mind if this got closed either.

PR #891, opened 2020-04-11
Add custom metrics for viper.Get calls

kmhagan:
We are using viper at my company and it has been working great! One of the common issues we face is people having a typo in one of their environment variable keys causing them to go down a troubleshooting rabbit hole until it's finally discovered that they didn't override one of their settings. I'd like to have some generic hooks into when a config value is called to allow for custom implementations that can help with troubleshooting. In our case an internal endpoint where we can see the environment variables being called and a list of keys available but not being called.

This pull request would allow people to set up custom metrics for whenever a viper.Get call is made.

Example:
```
        customMetric  := func(key string, val interface{}) {
		fmt.Printf("Config get call made '%v: %+v'\n", key, val)
	}
	v := viper.New()
	v.AddGetMetric(customMetric)

       isPretty := v.GetBool("PRETTY_OUTPUT")
       logLevel := v.GetString("LOG_LEVEL")
       port := v.GetString("server_port")
```

Output:
```
Config get call made 'PRETTY_OUTPUT: <nil>'
Config get call made 'LOG_LEVEL: debug'
Config get call made 'server_port: :8080'
```

Please let me know your thoughts and if you have an idea for a better approach for this problem or recommendation on variable/method names

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -213,7 +213,8 @@ type Viper struct {
 	// This will only be used if the configuration read is a properties file.
 	properties *properties.Properties
 
-	onConfigChange func(fsnotify.Event)
+	onConfigChange  func(fsnotify.Event)
+	onGetCallMetric func(key string, value interface{})
 }
 
 // New returns an initialized Viper instance.
@@ -537,6 +538,11 @@ func (v *Viper) AddSecureRemoteProvider(provider, endpoint, path, secretkeyring 
 	return nil
 }
 
+func AddGetMetric(getCallMetric func(key string, val interface{})) { v.AddGetMetric(getCallMetric) }
+func (v *Viper) AddGetMetric(getCallMetric func(key string, val interface{})) {
+	v.onGetCallMetric = getCallMetric
+}
+
 func (v *Viper) providerPathExists(p *defaultRemoteProvider) bool {
 	for _, y := range v.remoteProviders {
 		if reflect.DeepEqual(y, p) {
@@ -726,6 +732,9 @@ func Get(key string) interface{} { return v.Get(key) }
 func (v *Viper) Get(key string) interface{} {
 	lcaseKey := strings.ToLower(key)
 	val := v.find(lcaseKey, true)
+	if v.onGetCallMetric != nil {
+		v.onGetCallMetric(key, val)
+	}
 	if val == nil {
 		return nil
 	}


sagikazarmark:
Interesting idea. I think this problem hits many users.

Here are my first thoughts:

- My first instinct is just to use a debugger. That will tell you exactly where Get was called, with what parameters and what it returned
- My second best idea is to just fetch everything with `AllSettings`. That will give you everything stored in Viper, helps with debugging.
- Last, but not least you could create an interface for Viper, use it inside your application (instead of the actual Viper type) and implement a wrapper around the actual Viper type (implementing the same interface). That would give you exactly the described behavior. I find this solution slightly better, because composition can keep the Viper core clean.

WDYT?

kmhagan:
Hi @sagikazarmark,

Thanks for your feedback!

- Unfortunately this is for issues that usually arise while promoting to production where we don't have access to a debugger.
- `AllSettings` is very helpful for debugging but doesn't give you the full picture.
With `AllSettings` you could develop a debugging endpoint that returns the following:
```
All Settings available:
HTTP_DISABLED: true
LOG_LEVEL: debug
SERVER_PORT: :8080
```
Using both `customMetric` and `AllSettings` function you would be able to create the following:
```
Settings called:
HTTPS_DISABLED: <nil>
SERVER_PORT: :8080

Settings available and not called:
HTTP_DISABLED: true
LOG_LEVEL: debug
```
The second endpoint would save more time by being able to quickly see that HTTPS_DISABLED is not set and the reason is because there's a typo in the environment variables. You could also quickly point out that while LOG_LEVEL is set to debug the code never actually calls it.
- Creating an interface and using that is certainly a viable option and would help keep the Viper core lean. However it goes against one of the goals of the SDK we provide to our developers. We try to use the libraries directly so they have full control of how they want to use the library and don't have to worry about our wrappers becoming out of date for new functionality or not implementing a method they would want to use. When a feature is requested or certain monitoring needs are lacking I'm first trying to first make a PR to the open source library so they can continue to use the library directly and it also gives me a chance to contribute back to the projects we rely on.

Thanks for reading! I intentionally made this feature more of an opt-in functionality where a developer can fill in the blanks and implement how they see fit or not use it and have practically no performance impact. If you think the disadvantages are greater than the advantages that this PR would provide and decide to decline I will most likely just leverage what `AllSettings` provides for my debugging endpoint. The second example is a nice to have but not a blocker to the point of introducing a wrapper.

sideyewa (approved):

Issue #898, opened 2020-04-21
[Vulnerability] gorilla/websocket 1.4.0


jpmcb:
This issue is related to an issue in `spf13/cobra`: https://github.com/spf13/cobra/issues/1091

There appears to be a vulnerability in [gorilla 1.4.0](https://github.com/gorilla/websocket/security/advisories/GHSA-jf24-p9p9-4rjh)

```
$ go mod why -m github.com/gorilla/websocket
...
# github.com/gorilla/websocket
github.com/spf13/viper/remote
github.com/bketelsen/crypt/config
github.com/bketelsen/crypt/backend/etcd
github.com/coreos/etcd/client
github.com/coreos/etcd/client.test
github.com/coreos/etcd/integration
github.com/coreos/etcd/embed
github.com/tmc/grpc-websocket-proxy/wsproxy
github.com/gorilla/websocket
```

sagikazarmark:
Fixed in #899

DarthHater:
Hi there! Can y'all cut a new release of Viper that pins to the newer version? It's great this got merged but upstream consumption is still stuck on `1.6.3` which is still pointed at `1.4.0` for gorilla/websocket.

sagikazarmark:
@DarthHater it shouldn't matter. Because of MVS this only means your app needs minimum 1.4.0, but ideally it should install the latest.

DarthHater:
That's not what I'm experiencing locally. For example:

Output of `go list -m all`:
```
$ go list -m all
github.com/sonatype-nexus-community/hashbrowns
cloud.google.com/go v0.26.0
github.com/BurntSushi/toml v0.3.1
github.com/Flaque/filet v0.0.0-20190209224823-fc4d33cfcf93
github.com/Masterminds/semver v0.0.0-20180403130225-3c92f33da7a8
github.com/Masterminds/vcs v1.13.1
github.com/OneOfOne/xxhash v1.2.2
github.com/alecthomas/template v0.0.0-20160405071501-a0175ee3bccc
github.com/alecthomas/units v0.0.0-20151022065526-2efee857e7cf
github.com/armon/consul-api v0.0.0-20180202201655-eb2c6b5be1b6
github.com/armon/go-radix v1.0.0
github.com/beevik/etree v1.1.0
github.com/beorn7/perks v1.0.0
github.com/boltdb/bolt v1.3.1
github.com/cespare/xxhash v1.1.0
github.com/client9/misspell v0.3.4
[...]

PR #899, opened 2020-04-21
Update gorilla dependency

jpmcb:
This dependency bump is in reference to #898 and https://github.com/spf13/cobra/issues/1091

Bumps to the latest patch version of `gorilla/websocket` 1.4

Let me know if there are questions!


DarthHater:
@sagikazarmark it's just good to know someone cares :)

Thanks for doing this!

PR #901, opened 2020-04-22
Fix ConfigFileUsed to return config in search path

### LINKED ISSUE #900: ConfigFileUsed returns empty after setting config name and config path
I have the file `~/.myconfdir/myconf.yaml`
When I do 
```
viper.SetConfigName("myconf")
viper.AddConfigPath("~/.myconfdir")
```
And then call `viper.ConfigFileUsed()` I get an empty string because it doesn't set `v.configFile`.
The current implementation is 
```
func (v *Viper) ConfigFileUsed() string {
  return v.configFile
}
```

I propose that the following suggestion
```
func (v *Viper) ConfigFileUsed() string {
  file, _ := v.getConfigFile()
  return file
}
```

wfernandes:
Returns the config file found in search paths as well instead of just
the configFile property.

Fixes #900

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -459,8 +459,11 @@ func (v *Viper) getEnv(key string) (string, bool) {
 }
 
 // ConfigFileUsed returns the file used to populate the config registry.
-func ConfigFileUsed() string            { return v.ConfigFileUsed() }
-func (v *Viper) ConfigFileUsed() string { return v.configFile }
+func ConfigFileUsed() string { return v.ConfigFileUsed() }
+func (v *Viper) ConfigFileUsed() string {
+	f, _ := v.getConfigFile()
+	return f
+}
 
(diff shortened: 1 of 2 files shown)


DanielRuf (approved):

sagikazarmark:
This is a little bit problematic. Consider the following (probably unrealistic, but valid example):

```golang
viper.ConfigFileUsed()
viper.AddConfigPath("/etc/myapp")
viper.ReadInConfig()
```

Before this change, the above code will properly find a config file in `/etc/myapp`. After applying this change, it won't.

Generally speaking, mixing read-only operations with write operations is not a good idea.

Here is my suggestion: call `findConfigFile` function directly in `ConfigFileUsed` when the `configFile` field is unset. It'll duplicate a little bit of code, but it'll be more correct IMHO.

PR #916, opened 2020-05-21
Skipping error != nil checks in favor of the fast return.

butuzov:
Checking for `nil` error doesn't make a scene, since we return it (or `nil`).

All `func (v *VIper) Unmarshal*` methods affected.


butuzov:
JIC - it also can be useful to add `//go:noinline` directive. Should I do it?

Issue #921, opened 2020-06-01
Can't save no string value in ini config file.


zhangyongding:
panic: interface conversion: interface {} is bool, not string

goroutine 1 [running]:
github.com/spf13/viper.(*Viper).marshalWriter(0xc0000847e0, 0xe1b8a0, 0xc0000108f8, 0xc0001782c9, 0x3, 0xe1b8a0, 0xc0000108f8)
        /home/constantd/go/pkg/mod/github.com/spf13/viper@v1.7.0/viper.go:1652 +0x125a
github.com/spf13/viper.(*Viper).writeConfig(0xc0000847e0, 0xc0001782a0, 0x2c, 0x1, 0x0, 0x0)
        /home/constantd/go/pkg/mod/github.com/spf13/viper@v1.7.0/viper.go:1472 +0x203
github.com/spf13/viper.(*Viper).WriteConfig(0xc0000847e0, 0xced380, 0xb)
        /home/constantd/go/pkg/mod/github.com/spf13/viper@v1.7.0/viper.go:1414 +0x61
github.com/spf13/viper.WriteConfig(...)
        /home/constantd/go/pkg/mod/github.com/spf13/viper@v1.7.0/viper.go:1408

PR #923, opened 2020-06-02
fix bug save value for ini config file

zhangyongding:


zhangyongding:
#921

sagikazarmark:
Can you please write a test demonstrating the faulty behavior?

zhangyongding:
```
package main

import (
	"fmt"

	"github.com/spf13/viper"
)

func main() {
	viper.SetConfigFile("config.ini")
	viper.AddConfigPath(".")
	viper.SetConfigType("ini")

	viper.Set("set.boolKey", true)
	viper.Set("set.intKey", 1)

	viper.WriteConfig()
	fmt.Println(viper.ConfigFileUsed())
}

sagikazarmark:
I meant an automated test to the existing set of test suites.

zhangyongding:
Let me try

sagikazarmark:
@zhangyongding can you add some tests [here](https://github.com/spf13/viper/blob/master/viper_test.go#L1375) or probably in a separate test similar to [toml](https://github.com/spf13/viper/blob/master/viper_test.go#L1498) proving the bug and that this patch fixes it?

Let me know if you need any help with it!

Thanks!

Issue #339, opened 2017-04-24
How to get a array config value from env?


wangxufire:
e.g I want get the hobbies(in README) value from env.

gbolo:
This is a good question. So we can retrieve this list like this: `viper.GetStringSlice("hobbies")`.

Example Code:
```go
package main

import (
	"fmt"
	"github.com/spf13/viper"
	"bytes"
	"strings"
)

var yamlExample = []byte(`
name: steve
hobbies:
- skateboarding
- snowboarding
`)

func main(){
	viper.SetConfigType("yaml")
	viper.SetEnvKeyReplacer(strings.NewReplacer(".", "_"))
	viper.AutomaticEnv()
	viper.ReadConfig(bytes.NewBuffer(yamlExample))

	t := viper.GetStringSlice("hobbies")
	fmt.Printf( "Type: %T,  Size: %d \n", t, len(t) )
	for i, v := range t {
		fmt.Printf("Index: %d, Value: %v\n", i, v )
	}
}
```

When I run this without setting an environment variable override I get:
```
$ go run main.go 
Type: []string,  Size: 2 
Index: 0, Value: skateboarding
Index: 1, Value: snowboarding
```

Overriding this through env:
```
$ HOBBIES="devops go docker ansible" go run main.go 
Type: []string,  Size: 4 
Index: 0, Value: devops
Index: 1, Value: go
Index: 2, Value: docker
Index: 3, Value: ansible
```

@spf13 perhaps we should include this in the main **README.md**? Also what if we have a list of maps, how can we override that?

Kyroy:
> Also what if we have a list of maps, how can we override that?
Is it possible?

fopina:
> @spf13 perhaps we should include this in the main **README.md**? Also what if we have a list of maps, how can we override that?

Definitely, just spent quite some time going through the code to find how to pass stringslices through ENV...
Also some way to support spaces in the slices would be nice

kchristidis:
> Also what if we have a list of maps, how can we override that?

Has anyone figured this out?

fopina:
> > Also what if we have a list of maps, how can we override that?
> 
> Has anyone figured this out?

going back to code to find that as I needed as well (but haven't tested yet) https://github.com/spf13/cast/blob/master/caste.go#L876

So if we're using `GetStringMap` it ends there where, if value is string, it will be parsed as JSON... 👍 for doc updates again...

manya0393:
[...]

PR #946, opened 2020-08-10
Add BindEnvSliceValue()

leventov:
Add `BindEnvSliceValue()` to support auto binding of environment variables into slice fields. Related to https://github.com/spf13/viper/issues/339.

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -206,6 +206,7 @@ type Viper struct {
 	kvstore        map[string]interface{}
 	pflags         map[string]FlagValue
 	env            map[string]string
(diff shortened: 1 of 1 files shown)


navi89mai:
Any plan to merge and release new version?

leventov:
@sagikazarmark could you please take a look at this?

sagikazarmark:
I don't think this is the right solution. I would instead use `Unmarshal` or `UnmarshalKey` with a decode hook.

Read more about the topic in this post: https://sagikazarmark.hu/blog/decoding-custom-formats-with-viper/

PR #949, opened 2020-08-11
BindPFlag() detect nil flag before wrapping in pflagValue

bhamail:
I ran into a nasty problem while changing a Cobra command from `flag` to a `persistentFlag`. The old code to bind the flag to viper did not error out after the new change, and allowed me to bind a wrapped empty flag. (details below)

This PR adds a check to make sure the passed in flag value is not `nil`. Without this check, a `nil` value gets wrapped in a new `pflagValue`, and later nil checks do not detect the problem.

More details: 

```
cmd.Flags().StringVarP(&myConfig, "myFlagName", ...
...
if err := viper.BindPFlag("myViperKey", cmd.Flags().Lookup("myFlagName")); err != nil { ...
```

When I changed the Cobra code to use a Persistent flag (as shown below), the viper.Bind call didn't notice any issue. The problem was the cmd.Flags.Lookup() was now returning nil. I needed to use cmd.PersistentFlags.Lookup() instead. It would have been helpful if Viper detected the nil parameter.
```
cmd.PersistentFlags().StringVarP(&myConfig, "myFlagName", ...
...
// would have been helpful if this call errored out
if err := viper.BindPFlag("myViperKey", cmd.Flags().Lookup("myFlagName")); err != nil { ...
// fixed call below
//if err := viper.BindPFlag("myViperKey", cmd.PersistentFlags().Lookup("myFlagName")); err != nil {...
```

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -962,6 +962,9 @@ func (v *Viper) BindPFlags(flags *pflag.FlagSet) error {
 //
 func BindPFlag(key string, flag *pflag.Flag) error { return v.BindPFlag(key, flag) }
 func (v *Viper) BindPFlag(key string, flag *pflag.Flag) error {
+	if flag == nil {
+		return fmt.Errorf("flag for %q is nil", key)
+	}
 	return v.BindFlagValue(key, pflagValue{flag})
 }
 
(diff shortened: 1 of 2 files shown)


jwriteclub:
👍 

This issue is insidious and leads to apparent issues far away in viper if undetected.

sagikazarmark (approved):
Thanks!

PR #958, opened 2020-08-19
add parser

legenove:
Add the interface of Parser. If the user wants to identify the user-defined configuration, such as XML, then the user can customize a parser and register it to process the customized configuration.

### DIFF
--- a/parsers.go
+++ b/parsers.go
@@ -4,6 +4,9 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"io"
+	"strings"
+
 	"github.com/hashicorp/hcl"
 	"github.com/hashicorp/hcl/hcl/printer"
 	"github.com/magiconair/properties"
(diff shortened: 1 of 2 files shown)


sagikazarmark:
@legenove thanks a lot for putting the effort into this PR. I really appreciate it.

Sadly, I cannot merge it as it is:

- there is already a PR in the works implementing the same thing, related to Viper 2. There are a couple rough edges to iron out around backward compatibility
- the parsing logic needs to be file agnostic
- the PR contains backward incompatible changes

Let's keep this open until a similar PR is submitted/merged, but ultimately I don't think this is the right path forward.

But huge thanks for the effort!

legenove:
> @legenove thanks a lot for putting the effort into this PR. I really appreciate it.
> 
> Sadly, I cannot merge it as it is:
> 
> * there is already a PR in the works implementing the same thing, related to Viper 2. There are a couple rough edges to iron out around backward compatibility
> * the parsing logic needs to be file agnostic
> * the PR contains backward incompatible changes
> 
> Let's keep this open until a similar PR is submitted/merged, but ultimately I don't think this is the right path forward.
> 
> But huge thanks for the effort!

legenove:
It's OK. I'm looking forward to viper2. @sagikazarmark

PR #972, opened 2020-09-10
Allow BindEnv to register multiple environment variables.

### LINKED ISSUE #971: Allow multiple env. var. keys in BindEnv
It sometimes can happen that a project updates its conventions for environment variables, deprecating others while still wanting to offer support for them. For this reason it would be nice if `BindEnv` would allow this. The good news is that `BindEnv`'s function signature takes a variadic input so adding even more arguments to it would not be a breaking change.

An example of how it would work:

```go
viper.BindEnv("my.key", "APP_MY_KEY", "MY_KEY")
```

Then, when running `viper.Get("my.key")`, the first environment variable from the list would win. Basically the first entry is the latest valid form and the subsequent ones are the deprecated ones we want to keep support for.

Would it be ok to open a PR for this? It's a fairly small change with full backwards compatibility.

gbbr:
This change modifies BindEnv to permit a list of environment variable
names in order to support multiple env. vars. for the same config key.
When this form is used, env. keys take precedence in the written order.

Closes #971

### DIFF
--- a/README.md
+++ b/README.md
@@ -245,9 +245,10 @@ using `SetEnvPrefix`, you can tell Viper to use a prefix while reading from
 the environment variables. Both `BindEnv` and `AutomaticEnv` will use this
 prefix.
 
-`BindEnv` takes one or two parameters. The first parameter is the key name, the
-second is the name of the environment variable. The name of the environment
-variable is case sensitive. If the ENV variable name is not provided, then
+`BindEnv` takes one or more parameters. The first parameter is the key name, the
+rest are the name of the environment variables to bind to this key. If more than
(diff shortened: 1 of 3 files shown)


sagikazarmark (approved):
Thanks!

sagikazarmark (changes requested):
Actually @gbbr would you mind adding some documentation about this feature:
- to the `BindEnv` function header
- to the README

Thanks!

sagikazarmark:
Ping @gbbr

gbbr:
Done. Let me know what you think.

sagikazarmark (approved):
LGTM, thanks!

PR #978, opened 2020-09-14
update package and add write file with old data

olongfen:


sagikazarmark:
There are a bunch of changes that are not necessary. Please rebase from master.

Also, can you please explain what this PR is about. It doesn't seem to make sense to be honest. We use afero as a filesystem abstraction.

olongfen:
thank , this PR i want to use when i run code if there is no file that create file and write by default value, else overwite file by old file value and if the old file no filed will  be creat filed by default  value.

PR #979, opened 2020-09-15
writeConfig operations are atomic

ItalyPaleAle:
This PR modifies the behavior of `writeConfig` so all write operations are atomic. It works by first writing the data to a temporary file (in the same directory, using `afero.TempFile`), then calling `Rename` to move the file to the desired path.

Because rename calls are atomic, this guarantees that if the app crashes while the file is being written, the config file will not be corrupted.

The interface for the method did not change, and tests are passing (at least, they do in my Codespace)


ItalyPaleAle:
Hi is there any update on this PR?

Issue #819, opened 2020-01-01
Keys with nil values are ignored


eskriett:
I'm currently running viper `1.6.1`.

Consider the following yaml config, `config.yaml`:

```yaml
---
entries:
  a:
    foo: bar
  b: ~
```

My intention is to unmarshal this into a Config struct of the form:

```go
type Entry struct {
    Foo string
}

type Config struct {
    Entries map[string]*Entry
}
```

Where the expected output when printing would look like: `map[a:<pointer> b:<nil>]`. However, it appears keys with nil values are lost during the unmarshalling process.

The following POC demonstrates loading the `config.yaml` file above using viper and yaml.v2. The yaml.v2 is what I would expect.

```go
package main

import (
    "fmt"
    "io/ioutil"
    "log"

    "github.com/spf13/viper"
    "gopkg.in/yaml.v2"
)

type Entry struct {
    Foo string
}

type Config struct {
    Entries map[string]*Entry
}

func main() {
    viper.AddConfigPath(".")
    viper.SetConfigName("config")

    if err := viper.ReadInConfig(); err != nil {
        log.Fatal(err)
    }

    viperConfig := Config{}
    if err := viper.Unmarshal(&viperConfig); err != nil {
        log.Fatal(err)
    }

    data, err := ioutil.ReadFile("config.yaml")
    if err != nil {
        log.Fatal(err)
    }

    yamlConfig := Config{}
    if err := yaml.Unmarshal(data, &yamlConfig); err != nil {
        log.Fatal(err)
    }

    fmt.Println("AllKeys():", viper.AllKeys())
    fmt.Println("AllSettings():", viper.AllSettings())
    fmt.Println("viper config:", viperConfig)
    fmt.Println("yaml.v2 config:", yamlConfig)
}
```

When running the output is:

```
AllKeys(): [entries.a.foo entries.b]
AllSettings(): map[entries:map[a:map[foo:bar]]]
viper config: {map[a:0xc00009f430]}
yaml.v2 config: {map[a:0xc00009f550 b:<nil>]}
```

So viper is clearly aware the "b" key exists given the `AllKeys()` output, but it looks to be lost during the call to `AllSettings()`.

[...]

PR #982, opened 2020-09-18
Allow nil values to be returned by AllSettings

bogdandrutu:
Fixes https://github.com/spf13/viper/issues/819


sagikazarmark:
I'm not sure this is a good idea as it's most likely a break in behavior. See my comment on the original issue.

bogdandrutu:
@sagikazarmark done, changed to allow an optional config. Not sure if you want to use options in the ctor or setters. I see both, and I picked a random one.

bogdandrutu:
Please take another look, I added a ctor option to allow nil values to be returned

bogdandrutu:
@sagikazarmark any chance to merge this?

viper_test.go
```diff
@@ -2258,6 +2258,37 @@ func TestKeyDelimiter(t *testing.T) {
...
 }
 
+var yamlExampleWithNilValues = []byte(`Hacker: true
+name: steve
+hobbies:
+clothing:
+  jacket:
+`)
+
+func TestNilValues(t *testing.T) {
```
sagikazarmark, line 2268: Can you please extend this test with (unset) environment variables bound to a custom key?

Like this: `BindEnv("age", "AGE")`

Just to make sure, that env vars don't mess up anything (and `AllKeys` really returns keys with explicitly set values)

viper.go
```diff
@@ -1960,9 +1969,8 @@ func (v *Viper) AllSettings() map[string]interface{} {
 	// start from the list of keys, and construct the map one value at a time
 	for _, k := range v.AllKeys() {
 		value := v.Get(k)
-		if value == nil {
-			// should not happen, since AllKeys() returns only keys holding a value,
-			// check just in case anything changes
```
sagikazarmark, line 1965: Can we retain some of this comment with the reference to AllKeys behavior?

sagikazarmark (changes requested):
Thanks!

Looks good to me, but I had a few minor requests.

peter-mogensen:
The proposed fix only fixes half of the problem.
AllKeys not only ignores nil values, but also empty maps/objects. It could of course be argued that that is a different problem, but it would seem weird if changing a key from nil to {} would make it go away.

Here's a test to illustrate what I mean.

```
var jsonExampleWithEmptyObjects = []byte(`
{
  "Hacker": true,
  "name": "steve",
  "hobbies": null,
  "clothing": {
    "jacket": null,
    "pocket": {}
  }
}  
`)

func TestEmptyObjects(t *testing.T) {
	vNil := NewWithOptions(AllowNilValues(true))
	vNil.SetConfigType("json")
	require.NoError(t, vNil.ReadConfig(strings.NewReader(string(jsonExampleWithEmptyObjects))))
	expected := map[string]interface{}{
		"hacker":  true,
		"name":    "steve",
		"hobbies": nil,
		"clothing": map[string]interface{}{
			"jacket": nil,
			"pocket": map[string]interface{}{},
		},
	}
	assert.EqualValues(t, expected, vNil.AllSettings())
}

```

PR #983, opened 2020-09-18
Use ristretto to improve find performance

bogdandrutu:
Cloned from https://github.com/spf13/viper/pull/849. Resolved merge commits, upgrade to ristretto 0.0.3

This patch significantly improves `v.find` performance by using
ristretto cache.

Ristretto was chosen because it is a very performant, thread-safe
cache that does not suffer form congestion or long GC pauses.

```
% go test -bench=. .
goos: darwin
goarch: amd64
pkg: github.com/spf13/viper
BenchmarkGetBool-8               5431267               211 ns/op
BenchmarkFind/cache=false-8       289495              4060 ns/op
BenchmarkFind/cache=true-8       6980976               171 ns/op
BenchmarkGet-8                   6332091               193 ns/op
BenchmarkGetBoolFromMap-8       235071828                5.09 ns/op
PASS
```


sagikazarmark:
@bogdandrutu CI is failing because you haven't run `go mod tidy` and it tries to modify `go.sum`.

bogdandrutu:
Not yet able to fix the lint issue (ran the `go mod tidy`), but there is an interesting test failure which may make this impossible to merge:

```
func TestBindPFlag(t *testing.T) {
	testString := "testing"
	testValue := newStringValue(testString, &testString)

	flag := &pflag.Flag{
		Name:    "testflag",
		Value:   testValue,
		Changed: false,
	}

	BindPFlag("testvalue", flag)

	assert.Equal(t, testString, Get("testvalue"))

	flag.Value.Set("testing_mutate")
	flag.Changed = true // hack for pflag usage

	assert.Equal(t, "testing_mutate", Get("testvalue"))
}
```

This test will fail because the flag is updated after the value is cached, so I am not sure what is the expected behavior.

bogdandrutu:
I can probably do that, but would like to understand what is the prefer way to set options via setters or ctor options? Asked same question in #982

sagikazarmark:
> would like to understand what is the prefer way to set options via setters or ctor options?

Sorry, missed that: ctor options is what I try to enforce in Viper these days.

bogdandrutu:
@sagikazarmark managed to pass the lint, I had to actually clean my go mod cache:

```
rm go.sum
go clean -modcache
go mod tidy
```

Also I fixed the problem, by default `New` will not create a cache, so only if users manually sets a cache instance this will work, added documentation about the caveat of using cache as well.

QeyCoder:
any update on this, are you going to release cache support?

sagikazarmark:
After thinking about this PR: I'm not going to accept it as is. I will add support for caching in a separate PR defining a custom interface instead of directly depending on ristretto.

I will probably add a ristretto implementation in a separate module, probably in https://github.com/go-viper

Issue #888, opened 2020-04-09
Rewrite the encoding layer as a library


sagikazarmark:
Currently file encoding is hard-wired into Viper which makes every consumer download the dependencies of all supported formats, even if they only want to use one or two (json, yaml, toml to name the most popular ones). These dependencies even make their way to the final binary.

In Viper 2 we want to [lower the number of dependencies](https://github.com/spf13/viper/issues/887). One way of doing that is getting rid of all the encoding libraries from the Viper core, and requiring consumers to manually import the formats they want to use:

```go
package main

import (
	"github.com/spf13/viper"
	_ "github.com/spf13/viper/encoding/yaml" // Add YAML support to Viper
	_ "github.com/spf13/viper/encoding/json" // Add JSON support to Viper
)

// ...
```

Although moving the encoding libraries to separate modules would be a breaking change, there is no reason against moving the code to separate packages inside the main module in Viper 1 and move them to separate modules to Viper 2 if we decide to break backwards compatibility.

Until there is actually need for people to import those packages manually, they should remain _internal_: `github.com/spf13/viper/internal/encoding/...`.

andig:
The main pain I feel and the biggest impact on dependency and code size is the remote config though. Remote config is already half-separated from the core and fully splitting it would already get large part of the job done if making everything optional is not achievable short-term.

PR #986, opened 2020-09-23
New encoding layer

sagikazarmark:
This PR adds an encoding layer to Viper, separating direct calls to encoding libraries to different packages and ultimately moving out most of the logic from the main Viper package.

The package implements the proposal from #888 with the exception of moving them out from the module to preserve backwards compatibility. That cannot happen in the v1 branch as it would be a breaking change, but introducing an encoding layer this way allows us to move things out of the module easily in the future.

Although this is a sane proposal and I believe it's the way forward, the abstraction is not perfect and we can already see some shortcomings:

- TOML does not support `map[string]interface{}` well (compared to JSON and YAML anyway), so it contains some conditional logic specific to that type.
- Many supported formats (INI, dotenv, Java properties) don't have a JSON/YAML style "unmarshal" API. In fact, some of them are completely different:
    - In case of dotenv, the env var names are generated by formatting the keys
    - In case of INI sections are created from keys

All this points towards a much higher level abstraction where we pass and return `map[string]interface{}` directly. The current implementation is a low level interface based on the Marshal/Unmarshal style of `encoding/json`.

Something like this would probably be better:

```go
// Decoder decodes the contents of b into a map[string]interface{}.
type Decoder interface {
	Decode(b []byte) (map[string]interface{}, error)
}

// Encoder encodes the contents of v into a raw byte representation.
type Encoder interface {
	Encode(v map[string]interface{}) ([]byte, error)
}
```

There are also some other missing pieces from this PR:
- Proper tests for the different codecs (currently we rely on Viper's tests which are...not necessarily good)
- Some codecs might have additional config options

I'm not sure whether this change should make its way to Viper as it is. On one hand it would be nice to start experimenting, on the other hand this is not a perfect solution and will require some changes for the long-term. I'll leave this PR here for now, wait for some potential feedback and merge it if it still feels right after a few days/weeks.

### DIFF
--- a/internal/encoding/decoder.go
+++ b/internal/encoding/decoder.go
@@ -1,0 +1,61 @@
+package encoding
+
+import (
+	"sync"
+)
+
+// Decoder decodes the contents of b into a v representation.
+// It's primarily used for decoding contents of a file into a map[string]interface{}.
+type Decoder interface {
+	Decode(b []byte, v interface{}) error
+}
+
+const (
+	// ErrDecoderNotFound is returned when there is no decoder registered for a format.
+	ErrDecoderNotFound = encodingError("decoder not found for this format")
+
+	// ErrDecoderFormatAlreadyRegistered is returned when an decoder is already registered for a format.
+	ErrDecoderFormatAlreadyRegistered = encodingError("decoder already registered for this format")
+)
+
+// DecoderRegistry can choose an appropriate Decoder based on the provided format.
+type DecoderRegistry struct {
+	decoders map[string]Decoder
+
+	mu sync.RWMutex
+}
+
+// NewDecoderRegistry returns a new, initialized DecoderRegistry.
+func NewDecoderRegistry() *DecoderRegistry {
+	return &DecoderRegistry{
+		decoders: make(map[string]Decoder),
+	}
+}
+
+// RegisterDecoder registers a Decoder for a format.
+// Registering a Decoder for an already existing format is not supported.
+func (e *DecoderRegistry) RegisterDecoder(format string, enc Decoder) error {
+	e.mu.Lock()
+	defer e.mu.Unlock()
+
+	if _, ok := e.decoders[format]; ok {
+		return ErrDecoderFormatAlreadyRegistered
+	}
+
+	e.decoders[format] = enc
+
+	return nil
+}
+
+// Decode calls the underlying Decoder based on the format.
+func (e *DecoderRegistry) Decode(format string, b []byte, v interface{}) error {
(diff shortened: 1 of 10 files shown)


internal/encoding/decoder.go
```diff
@@ -0,0 +1,58 @@
...
+const (
+	ErrDecoderNotFound                = encodingError("decoder not found for this format")
+	ErrDecoderFormatAlreadyRegistered = encodingError("decoder already registered for this format")
+)
+
+// DecoderRegistry can choose an appropriate Decoder based on the provided format.
+type DecoderRegistry struct {
+	decoders map[string]Decoder
+
+	mu sync.RWMutex
```
bogdandrutu, line 22: Not sure if matters, but can use sync.Map to avoid explicitly synchronize access to the map.

bogdandrutu (approved):

internal/encoding/decoder.go
sagikazarmark: Hm, yeah, interesting idea. For now, I think type safety is more important than performance, especially since this mutex won't be locked more than 5-10 times during an application's lifecycle. But I will certainly play with it in the next round of changes (this is not the final form of the encoding package).

Thanks!

internal/encoding/decoder.go
```diff
@@ -0,0 +1,58 @@
...
+// NewDecoderRegistry returns a new, initialized DecoderRegistry.
+func NewDecoderRegistry() *DecoderRegistry {
+	return &DecoderRegistry{
+		decoders: make(map[string]Decoder),
+	}
+}
+
+// RegisterDecoder registers a Decoder for a format.
+// Registering a Decoder for an already existing format is not supported.
+func (e *DecoderRegistry) RegisterDecoder(format string, enc Decoder) error {
```
chrisdoherty4, line 34: Would defining `type Key string` and replacing `format string` with `format Key` both here and in the `Decode()` method be appropriate? 

It's a trivial change, but cognitively it highlights the relationship between the parameters across the different methods while still allowing consumers to use strings.

sagikazarmark: I'm not sure it'd be. What does `Key` mean? Format I understand.

chrisdoherty4: Perhaps `Key` is the wrong type name. The core of the suggestion is to just better highlight the relationship across methods rather than relying on variable names; perhaps it's unnecessary.

sagikazarmark:
I'm going to merge this PR for now. I've created a note in the Viper 2 board with the next steps (will probably convert it to an issue).

PR #988, opened 2020-09-26
doc: notic check instance return by viper.Sub

j2gg0s:


sagikazarmark:
Thanks @j2gg0s ! I decided to rewrite this section entirely.

PR #998, opened 2020-10-01
Fix: Top level JSON Array `ReadInConfig`

baybaraandrey:
Hi, i want to fix some issues in your project
Please review my code and give me feedback
Thanks

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -1528,8 +1528,19 @@ func (v *Viper) unmarshalReader(in io.Reader, c map[string]interface{}) error {
 		}
 
 	case "json":
-		if err := json.Unmarshal(buf.Bytes(), &c); err != nil {
-			return ConfigParseError{err}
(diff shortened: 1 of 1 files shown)


sagikazarmark:
Hi @baybaraandrey

Thanks a lot for the contribution!

Unfortunately, I don't really think this fits Viper's style of loading configuration. It's quite arbitrary. Not to mention we would have to add this to other sources as well.

Instead, I'd suggest loading the JSON content yourself, parsing it as you see fit into `map[string]interface{}` chunks and use [MergeConfigMap](https://pkg.go.dev/github.com/spf13/viper@v1.7.1#MergeConfigMap).

In Viper 2, you'll be able to use your own loading mechanisms much-much more easily, so you will be able to hook into Viper with a custom logic, like this.

With that, I'm declining this change. Thanks a lot anyway!

PR #1005, opened 2020-10-15
Add env prefix accessor method

chrisdoherty4:
At present, there's no way of retrieving the Viper instances environment variable prefix. This adds a simple accessor method to retrieve the prefix.

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -428,6 +428,15 @@ func (v *Viper) SetEnvPrefix(in string) {
 	}
 }
 
+// EnvPrefix returns the ENV prefix set with SetEnvPrefix for the global
+// Viper instance.
+func EnvPrefix() string { return v.envPrefix }
+
+// EnvPrefix returns the env prefix set with SetEnvPrefix.
+func (v *Viper) EnvPrefix() string {
+	return v.envPrefix
+}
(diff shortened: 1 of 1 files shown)


sagikazarmark:
Hi @chrisdoherty4 !

Thanks for the contribution!

I'd argue that this is is an internal detail for Viper. In the future, Viper will expose much less information about itself (eg. only `Get*` methods).

Why do you need this information? More importantly, why do you need this from Viper?

chrisdoherty4:
Hey @sagikazarmark, thanks for getting back to me.

I totally appreciate the point of view that this could be an internal detail. The specific use-case I had was reconstructing the environment variables for a help menu. More generally, if 2 parts of an application are interfacing with Viper it seemed appropriate to have Viper be queryable so I don't need a higher level construct that tracks that information. 

In hindsight, perhaps a higher level construct would be appropriate. What do you think?

sagikazarmark:
I'd say a higher level construct sounds like a better design choice, at least I wouldn't build anything directly on Viper. Ideally, Viper should never leave your `main` func.

chrisdoherty4:
Sounds good. Thanks.

Issue #1019, opened 2020-11-06
problem getting config from environement if key has a dash (-) in it.


flyhard:
I have a key with a - in the name. 

Since this works everywhere except if reading from the ENV, I suppose this is a bug.

PR #1020, opened 2020-11-06
Fixing env names #1019

flyhard:


sagikazarmark:
Thanks @flyhard for the contribution.

Environment variable names are preserved as is on purpose. You can use [`viper.SetEnvKeyReplacer`](https://pkg.go.dev/github.com/spf13/viper#Viper.SetEnvKeyReplacer) for this purpose.

PR #1027, opened 2020-11-18
Update README.md

darlandieterich:
Added live example


sagikazarmark:
Thanks for the idea @darlandieterich !

I came up with a slightly different version and added it to the README.

PR #1043, opened 2020-12-07
Add ini load options argument

gandarez:
This PR solves a problem where in some cases we need to specify some behaviours to ini parser (gopkg.in/ini.v1). It also bumps it to latest version as it was extremely outdated.

```ini
[settings]
apikey = 1234

[projectmap]
projects/foo = new project name
^/home/user/projects/bar(\d+)/ = project{0}

[git]
submodules_disabled = 
  .*secret.*
  fix.*
```

Without passing `AllowPythonMultilineValues` to ini parsing function it breaks with error `While parsing config: key-value delimiter not found: .*secret.*`. I believe it helps other developers looking for using viper with integrated configuration files as we've been developing at WakaTime.

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -199,6 +199,9 @@ type Viper struct {
 	configPermissions os.FileMode
 	envPrefix         string
 
+	// Specific commands for ini parsing
+	iniLoadOptions ini.LoadOptions
+
 	automaticEnvApplied bool
 	envKeyReplacer      StringReplacer
 	allowEmptyEnv       bool
@@ -1637,7 +1640,7 @@ func (v *Viper) unmarshalReader(in io.Reader, c map[string]interface{}) error {
 		}
 
 	case "ini":
-		cfg := ini.Empty()
+		cfg := ini.Empty(v.iniLoadOptions)
 		err := cfg.Append(buf.Bytes())
 		if err != nil {
 			return ConfigParseError{err}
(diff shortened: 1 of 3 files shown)


alanhamlett:
Ping on this.

gandarez:
Any update here?

viper.go
```diff
@@ -2079,6 +2082,13 @@ func (v *Viper) SetConfigPermissions(perm os.FileMode) {
 	v.configPermissions = perm.Perm()
 }
 
+// SetIniLoadOptions sets the load options for ini parsing.
+func SetIniLoadOptions(in ini.LoadOptions) { v.SetIniLoadOptions(in) }
```
sagikazarmark, line 2086: Please use the new functional option interface: https://github.com/spf13/viper/blob/d10c856f6c7be1fcdfc692b912e1d047c3ac0036/viper.go#L243

sagikazarmark (changes requested):
Thanks a lot for the contribution!

Please see the requested changes. Also, please make sure to run `go mod tidy`.

viper.go
gandarez: sure!

gandarez:
@sagikazarmark fixes applied.

gandarez:
Any update here?

gandarez:
ping here

gandarez:
ping

gandarez:
ping

sagikazarmark (approved):
It took me some time to decide whether this is a good idea to do. I realized that in the v1 API there is probably no better alternative, especially because of the size of that config struct.

In Viper v2 the INI encoding will have to be configured separately.

sagikazarmark:
@gandarez can you please rebase your PR?

gandarez:
@sagikazarmark rebased!

sagikazarmark (approved):
Thanks!

PR #964, opened 2020-08-26
Update crypt to address CVE-2020-15114

louiedp3:
Addresses [CVE-2020-15114](https://nvd.nist.gov/vuln/detail/CVE-2020-15114) by upgrading crypt. `nancy` returns no vulnerabilities:
```
go list -json -m all | nancy
 __  __
/\ \/\ \
\ \ `\\ \      __       ___      ___    __  __
 \ \ , ` \   /'__`\   /' _ `\   /'___\ /\ \/\ \
  \ \ \`\ \ /\ \L\.\_ /\ \/\ \ /\ \__/ \ \ \_\ \
   \ \_\ \_\\ \__/.\_\\ \_\ \_\\ \____\ \/`____ \
    \/_/\/_/ \/__/\/_/ \/_/\/_/ \/____/  `/___/> \
                                            /\___/
                                            \/__/
  _        _                           _    _
 /_)      /_` _  _  _ _/_     _  _    (/   /_` _ . _  _   _/  _
/_) /_/  ._/ /_// //_|/  /_/ /_//_'  (_X  /   / / /_'/ //_/ _\
    _/                   _/ /
Nancy version: 0.3.1

Non Vulnerable Packages

[1/155]	pkg:golang/cloud.google.com/go@0.46.3
[2/155]	pkg:golang/cloud.google.com/go/bigquery@1.0.1
[3/155]	pkg:golang/cloud.google.com/go/datastore@1.0.0
[4/155]	pkg:golang/cloud.google.com/go/firestore@1.1.0
[5/155]	pkg:golang/cloud.google.com/go/pubsub@1.0.1
[6/155]	pkg:golang/cloud.google.com/go/storage@1.0.0
[7/155]	pkg:golang/dmitri.shuralyov.com/gpu/mtl@0.0.0-20190408044501-666a987793e9
[8/155]	pkg:golang/github.com/BurntSushi/toml@0.3.1
[9/155]	pkg:golang/github.com/BurntSushi/xgb@0.0.0-20160522181843-27f122750802
[10/155]	pkg:golang/github.com/OneOfOne/xxhash@1.2.2
[11/155]	pkg:golang/github.com/alecthomas/template@0.0.0-20160405071501-a0175ee3bccc
[12/155]	pkg:golang/github.com/alecthomas/units@0.0.0-20151022065526-2efee857e7cf
[13/155]	pkg:golang/github.com/armon/circbuf@0.0.0-20150827004946-bbbad097214e
[...]

PR #961, opened 2020-08-19
Upgrade crypt and etcd

sagikazarmark:
Fixes #865
Fixes #956
Fixes #957
Fixes #964
Fixes #1049

PR #957, opened 2020-08-18
fix vulnerability: CVE-2020-15114 in etcd v3.3.13+incompatible

bhamail:
We're using Viper in [nancy](https://github.com/sonatype-nexus-community/nancy). Thanks so much for a great tool!

During a recent CI build, nancy discovered a vulnerability in a transitive dependency of Viper. This PR includes a `replace` directive to use a newly released update of that transitive dependency `etcd`. This transitive dep is pulled in by `github.com/bketelsen/crypt`.
```
$ go mod graph | grep coreos/etcd
github.com/bketelsen/crypt@v0.0.3-0.20200106085610-5cbc8cc4026c github.com/coreos/etcd@v3.3.13+incompatible
```

I submitted a [PR to crypt](https://github.com/bketelsen/crypt/pull/10), so hopefully there will soon be a non-vulnerable version of `crypt` available, to which Viper can upgrade. Meanwhile, this PR provides a way to use a non-vulnerable version of `etcd` if you like.

PR #1049, opened 2020-12-17
Update bketelsen/crypt to fix CVE-2020-15114.

dlorenc:
This picks up https://github.com/bketelsen/crypt/pull/10. It doesn't look like they've done
a new tag yet, so going to the latest commit (as of 17DEC2020) is about all we can do.


dlorenc:
I see a few other PRs that also try to address this:

https://github.com/spf13/viper/pull/964
https://github.com/spf13/viper/pull/957

Let me know if there's anything else I can do to help here - this update will allow all the importers of spf13/viper to update and remove the deprecated/vulnerable jwt-go library.

sagikazarmark:
As mentioned in #961 we are waiting for etcd 3.5 to be tagged for proper modules support.

dlorenc:
Thanks for the reply!

Just to add some context, keeping this at the older version causes any other repo that uses viper to be flagged by vulnerability scanners as having the jwt-go CVE.

sagikazarmark:
Yeah, I'm aware. That's actually a false alert, because it's not actually compiled in the final binary.

As far as I know 3.5 is pretty close and it'll improve modules support. Unfortunately, the currently it's not that good (and became somewhat worse with the last version)

dlorenc:
> Yeah, I'm aware. That's actually a false alert, because it's not actually compiled in the final binary.

Understood. I think it still could be problematic in things further up the dependency hierarchy that do use that codepath and viper at the same time, though.



> As far as I know 3.5 is pretty close and it'll improve modules support. Unfortunately, the currently it's not that good (and became somewhat worse with the last version)

I'm not sure what the etcd relation is here. I'm sure I'm misunderstanding something though. This PR was simply the result of updating the crypt library then running "go mod tidy". Is there a downside to merging early updates like these rather than waiting?

I'm trying to stay in the habit of keeping my projects at zero flagged CVEs, even for false positives like this one. I figure it'll be easier to notice a real one when it happens if I'm at 0, rather than if I have a bunch of false positives to wade through.

sagikazarmark:
> I'm not sure what the etcd relation is here

It's twofold:

- The jwt library is a dependency because of etcd
- The particular etcd version pulled in by crypt is a bit problematic from modules point of view. At least in my tests, go getting Viper didn't always work. (A bit of backstory: etcd was moved to a new organization and go modules was added, but not as a new major version which caused a whole lot of trouble

> I'm trying to stay in the habit of keeping my projects at zero flagged CVEs, even for false positives like this one.

Agreed. The reason I was holding this back is the one I explained above. I'll look into it again to see if things are better now, but I think without etcd 3.5 modules will not work perfectly.

> github.com/bketelsen/crypt v0.0.4-0.20201105235121-f295231a095e

This is also not going to fly: it's not a tagged version.

PR #1053, opened 2020-12-23
Allow merging configs with different types of leaf values

illarion:
In some environments, it is useful to be able to replace certain values with values of another type, for example when parsing toml as a base config and then parsing json as a merge config, numbers in json are stored as float64, and therefore cannot be merged to integer defaults specified in toml config. This PR fixes such situation and adds more type flexibility.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yaml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yaml
@@ -30,7 +30,7 @@ body:
   attributes:
     label: Go Version
     description: What version of Go are you using?
-    placeholder: 1.16
+    placeholder: "1.16"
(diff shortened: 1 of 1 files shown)


sagikazarmark:
Thanks @illarion !

TBH I don't know why that type check was in place, probably to avoid merging different collection data structures (maps, slices). I'll take a closer look later, but I think it makes sense.

illarion:
@sagikazarmark yes, it seems to be an attempt to protect maps from merging with something else, and it did too much extra protection! For instance, there is a type switch case that allows merging map[string]interface{} with a map[interface{}]interface{}, but the protection happens before it, therefore skipping it.

Issue #745, opened 2019-08-08
BUG: Sub Viper is not reflecting config value updates via cli-option or ENV


marcofranssen:
I have following cobra command

```go
func init() {
	rootCmd.AddCommand(serveCmd)

	serveCmd.Flags().String("listenAddr", ":5000", "server listen address")
	viper.BindPFlag("server.listenaddr", serveCmd.Flags().Lookup("listenAddr"))
}
```

Now when I run the command I try to get the config value.

```go
listenAddr := viper.Sub("server").GetString("listenaddr")
```

When running this command I see that the config value is not updated based on Env variable or commandline option.

```bash
$ ./my-app serve
listenAddr = :5000
$ ./my-app serve --listenAddr :6000
listenAddr = :5000
$ SERVER_LISTENADDR=:6000 ./my-app serve
listenAddr = :5000
```

However when changing the code to read the value as following it does work as expected.

```go
listenAddr := viper.GetString("server.listenaddr")
```

```bash
$ ./my-app serve
listenAddr = :5000
$ ./my-app serve --listenAddr :6000
listenAddr = :6000
$ SERVER_LISTENADDR=:8888./my-app serve
listenAddr = :8888
```

It would be excepted that the Sub viper will also have the correct values for the config. So I can pass the subviper to the package that depends on those configs.

Issue #1012, opened 2020-11-02
UnmarshalKey / Sub only uses read config, neglects environment variables 


manuelstein:
**Description**: 
A structured configuration is read using `ReadConfig` or `ReadInConfig`. Subsection keys are bound to environment variables. Viper correctly updates its config, i.e. key access `Get` / `AllSettings` works correctly. But when trying to use `Unmarshal`, only the config file seems to be used.

**Code**:
```golang
package main

import (
	"bytes"
	"fmt"
	"os"

	"github.com/spf13/viper"
)

func main() {
	v := viper.New()

	type Foo struct {
		Bar string `mapstructure:"bar"`
	}
	foo := Foo{
		Bar: "A",
	}
	var yamlExample = []byte(`
foo:
  bar: "B"
`)
	// Read config
	fmt.Println("Reading config")
	v.SetConfigType("yaml")
	v.ReadConfig(bytes.NewBuffer(yamlExample))
	fmt.Printf("%#v\n\n", v.AllSettings())

	// BindEnv
	fmt.Println("Setting env")
	os.Setenv("FOO_BAR", "C")
	v.BindEnv("foo.bar", "FOO_BAR")
	fmt.Printf("%#v\n\n", v.AllSettings())

	fmt.Println("Unmarshal")
	if err := v.Sub("foo").Unmarshal(&foo); err != nil {
		panic(err)
	}
	fmt.Printf("%#v\n\n", foo)
}
```

**Output**:
```
Reading config
map[string]interface {}{"foo":map[string]interface {}{"bar":"B"}}

Setting env
map[string]interface {}{"foo":map[string]interface {}{"bar":"C"}}

Unmarshal
main.Foo{Bar:"B"}
```

**Expected Output**:
```
...
Unmarshal
main.Foo{Bar:"C"}
```

KalleDK:
Ran into the same issue

KalleDK:
The error seems to come from the Get() method. If I do the same but change Sub to Get, I also "miss" the env values.

KalleDK:
Example expanded with Get also

**code**
```go
package main

import (
	"bytes"
	"fmt"
	"os"

	"github.com/spf13/viper"
)

func main() {
	v := viper.New()

	type Bar struct {
		Val string `mapstructure:"val"`
	}
	type Foo struct {
		Bar Bar `mapstructure:"bar"`
	}
	foo := Foo{
		Bar: Bar{
			Val: "A",
		},
	}
[...]

Issue #801, opened 2019-11-27
viper and cobra sub commands


lburgazzoli:
I'm trying to write a tool for which I would let user set defaults or custom CLI arguments through a configuration file or env vars. What I've done so far is something like: 

```go
package main

import (
	"errors"
	"log"
	"strings"

	"github.com/spf13/cobra"
	"github.com/spf13/pflag"
	"github.com/spf13/viper"
)

func bindPFlagsHierarchy(cmd *cobra.Command) error {
	for _, c := range cmd.Commands() {
		if err := bindPFlags(c); err != nil {
			return err
		}

		if err := bindPFlagsHierarchy(c); err != nil {
			return err
		}
	}

	return nil
}

func bindPFlags(cmd *cobra.Command) error {
	prefix := cmd.Name()

	for current := cmd.Parent(); current != nil; current = current.Parent() {
		name := current.Name()
		name = strings.ReplaceAll(name, "_", "-")
		name = strings.ReplaceAll(name, ".", "-")
		prefix = name + "." + prefix
	}

	cmd.Flags().VisitAll(func(flag *pflag.Flag) {
		name := flag.Name
		name = strings.ReplaceAll(name, "_", "-")
		name = strings.ReplaceAll(name, ".", "-")

		if err := viper.BindPFlag(prefix+"."+name, flag); err != nil {
			log.Fatalf("error binding flag %s with prefix %s to viper", flag.Name, prefix)
		}
	})

	return nil
}

type MainOptions struct {
}

type SubOptions struct {
	Info string `mapstructure:"info"`
}

func main() {
	mopt := MainOptions{}
	sopt := SubOptions{}

	viper.AutomaticEnv()
	viper.SetEnvKeyReplacer(strings.NewReplacer("_", ".", "-", "."))

	mainCmd := &cobra.Command{
		Use:   "main",
		PreRunE: func(cmd *cobra.Command, args []string) error {
			if err := viper.UnmarshalKey("main", &mopt); err != nil {
				return nil
			}

			return nil
		},
		RunE: func(cmd *cobra.Command, args []string) error {
			log.Printf("mopt: %+v", mopt)
			return nil
		},
	}

	subCmd1 := &cobra.Command{
		Use:   "sub1",
		PreRunE: func(cmd *cobra.Command, args []string) error {
			v := viper.Sub("main.sub1")
			if v == nil {
				return errors.New("no main.sub1")
			}

			if err := v.Unmarshal(&sopt); err != nil {
				return nil
			}

[...]

PR #1056, opened 2020-12-24
Sub parent inheritance

TaylorOno:
This PR fixes two issues with Sub():
-    Options were not copied from parent (e.g, automaticEnvApplied).
 -   Environment variable names were wrong: Sub("x").Sub("y").Get("z") should look up env var x.y.z but was instead looking up z

Fixes https://github.com/spf13/viper/issues/745
Fixes https://github.com/spf13/viper/issues/507

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -952,6 +952,7 @@ func (v *Viper) Sub(key string) *Viper {
 	if reflect.TypeOf(data).Kind() == reflect.Map {
 		subv.parents = append(v.parents, strings.ToLower(key))
 		subv.automaticEnvApplied = v.automaticEnvApplied
+		subv.envPrefix = v.envPrefix
 		subv.envKeyReplacer = v.envKeyReplacer
 		subv.config = cast.ToStringMap(data)
 		return subv
--- a/viper.go
+++ b/viper.go
@@ -206,6 +206,7 @@ type Viper struct {
 	envKeyReplacer      StringReplacer
 	allowEmptyEnv       bool
 
+	parents        []string
 	config         map[string]interface{}
 	override       map[string]interface{}
 	defaults       map[string]interface{}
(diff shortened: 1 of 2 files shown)


afirth:
Probably fixes:
#745
#507
#1012 (partial)
#801 (partial)

viper.go
```diff
@@ -847,6 +849,9 @@ func (v *Viper) Sub(key string) *Viper {
 	}
 
 	if reflect.TypeOf(data).Kind() == reflect.Map {
+		subv.parents = append(v.parents, strings.ToLower(key))
+		subv.automaticEnvApplied = v.automaticEnvApplied
```
afirth, line 853: envPrefix must also be preserved

afirth: otherwise `1181:                 if val, ok := v.getEnv(v.mergeWithEnvPrefix(envKey)); ok {` breaks

afirth: fixed in fd570e1 :100:

afirth (approved):
I hope @sagikazarmark will consider shipping this. If it needs some more work I would be happy to help. Otherwise sub is very much impaired.

If there's no chance of shipping it, then I would also volunteer to make a readme update and list the known caveats, and hopefully save someone else the pain of assuming `Sub()` is complete as you could reasonable assume reading https://github.com/spf13/viper#extracting-a-sub-tree: "Viper instance to the constructor that represents a subset of the configuration"

lightglitch:
@sagikazarmark any news on the possibility of merging this? Just found this inconsistency and this would be very helpful to solve it.

sagikazarmark:
The PR needs rebasing, otherwise it looks good to me.

TaylorOno:
This feels like it could be a breaking change, but using the `Sub` function to look up a top level value is something I hope nobody is doing

sagikazarmark:
@TaylorOno can you explain? (Or better yet, write a test?)

TaylorOno:
Im just saying it is possible some people maybe relying on the old behavior e.g. https://xkcd.com/1172/

sagikazarmark:
Fixes https://github.com/spf13/viper/issues/745
Fixes https://github.com/spf13/viper/issues/507

sagikazarmark (approved):
Thanks!

PR #1099, opened 2021-03-24
Sub inerits from automaticEnvApplied and envKeyReplacer

afirth:
parents are tracked so sub(x).sub(y).GetString(z) looks up x.y.z

-from upstream https://github.com/spf13/viper/pull/1056


afirth:
sorry, wrong fork

PR #1066, opened 2021-01-19
Update github.com/gogo/protobuf (CVE-2021-3121)

imander:
Update github.com/gogo/protobuf to v1.3.2 in response to [CVE-2021-3121](https://nvd.nist.gov/vuln/detail/CVE-2021-3121)

PR #1103, opened 2021-03-29
add nancy scan to CI build

bhamail:
Runs [nancy](https://github.com/sonatype-nexus-community/nancy) to scan for vulnerabilities as part of the CI build.

relates to PR #961 and PR #1066 in that those PR's should fix the build failures.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,3 +33,9 @@ jobs:
 
       - name: Test
         run: make test
+
+      - name: WriteGoList
+        run: go list -json -m all > go.list
+
+      - name: Nancy scan
+        uses: sonatype-nexus-community/nancy-github-action@main


bhamail:
Ping! Has anybody had a chance to look this over - is adds a nancy scan to the CI build?

I'm wondering if it fell through the cracks due to the CI build failing - but the failures appear to be due to security vulns - so sort of expected tail chasing going on.

Please let me know if there is anything I can do to help move it along.

sagikazarmark:
It's on my mental todo list, but I'd like to combine it with other tools (like snyk and dependabot). I need some time to figure things out. Till then, we can go with manual updates.

Until we can upgrade to etcd 3.5 (#1115) which will likely only happen in June, we can't do much about dependencies anyway as it pins quite a few dependencies to relatively old versions.

PR #1112, opened 2021-04-09
Get String Else Default

david-doit-intl:
Adds a fallback for get string

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -860,6 +860,16 @@ func (v *Viper) GetString(key string) string {
 	return cast.ToString(v.Get(key))
 }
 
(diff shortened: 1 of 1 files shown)


sagikazarmark:
@david-doit-intl Thanks for your contribution!

We have turned these kind of PRs down in the past for multiple reasons:

- it's trivial to write a wrapper function for `Get*` in your own code base
- the use of `Get*` functions is discouraged anyway
- there are a bunch more `Get*` functions and we should keep their behavior consistent

Let me write up an RFC for v2 about additional `Get*` functions and a potential automation for generating that code.

PR #1125, opened 2021-04-26
Add SetViper() function

fuwensun:
Usage
```
func TestNew(t *testing.T) {

        //  1, Gets global Viper instance
	vp:=viper.GetViper()
	t.Cleanup(func(){
                // 2, Waiting for recovery
		viper.SetViper(vp)  // Added by this PR
	})

        // 3, New global Viper instance for test
	viper.Reset()
	viper.SetConfigName("config")
	viper.SetConfigType("yaml")
	viper.AddConfigPath("./testdata/config")
	viper.ReadInConfig()
        // 4, Do test
        ......
}
```

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -781,6 +781,11 @@ func GetViper() *Viper {
 	return v
 }
 
+// SetViper sets the global Viper instance.
+func SetViper(vp *Viper) {
+	v=vp
+}
+
 // Get can retrieve any value given the key to use.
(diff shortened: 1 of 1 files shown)


sagikazarmark:
Thanks @fuwensun for the contribution!

Using the global Viper instance is discouraged, but even if you do use it: are there any reasons why Reset is not enough?

PR #1128, opened 2021-05-02
Minor changes

fziglio:
Mainly this changes `WatchConfig` using a single goroutine instead of 2.
See commit messages for details.
To simplify the difference ignore space changes when reviewing.


fziglio:
ping

sagikazarmark:
I'm going to need a little bit of time to properly review this PR.

fziglio:
ping

fziglio:
ping

sagikazarmark:
I wonder why there were two goroutines used in the original solution in the first place. It seems very intentional, but I can't figure out why and I'm not comfortable merging this until then, especially because the watching functionality of Viper is quite fragile.

PR #1133, opened 2021-05-11
Make SetEnvKeyReplacer similar to EnvKeyReplacer in signature so you can actually write a different implementation

scr-oath:


sagikazarmark:
Thanks @scr-oath ! We already have a patch like this in #870, but since technically a BC break (even if it's an edge case), it hasn't been merged yet.

In the meantime, you can use the [EnvKeyReplacer](https://pkg.go.dev/github.com/spf13/viper#EnvKeyReplacer) functional option when initializing Viper.

Issue #884, opened 2020-04-09
InConfig behavior is inconsistent


Serhioromano:
If I use GET

```go
viper.Get("profiles.local.delete_table")
```

It does return nested parameter but if I check

```go
viper.InConfig("profiles.local.delete_table")
```

It always return `false`

Serhioromano:
Now I did this way. Is this right?

```go
// Package cmd CLI tool package
// Copyright © 2020 Sergey Romanov aka Serhioromano <serg4172@mail.ru>
package cmd

import (
	"fmt"
	"os"

	"github.com/spf13/cobra"
	"github.com/spf13/viper"
)

var cfgFile string
var cnfLoaded bool

// rootCmd represents the base command when called without any subcommands
var rootCmd = &cobra.Command{
	Use:   "mysqlsync",
	Short: "Synchronize MySQL DB Model from one DB to another",
	Long: `Tool to migrate DB from development to productoion without a pain.	
It works though generating DB model snapshot JSON files.`,
}

// Execute adds all child commands to the root command and sets flags appropriately.
// This is called by main.main(). It only needs to happen once to the rootCmd.
func Execute() {
	if err := rootCmd.Execute(); err != nil {
		fmt.Println(err)
		os.Exit(1)
	}
}

func init() {
	cobra.OnInitialize(initConfig)

	rootCmd.PersistentFlags().StringVar(&cfgFile, "config", "", "config file (default is $PWD/.mysqlsync.json)")
	rootCmd.PersistentFlags().StringP("profile", "p", "", "Name of connection profile in configuration file. By adding profile you do not need to add any other flag.")
	rootCmd.PersistentFlags().String("db", "", "DB scheme name")
	rootCmd.PersistentFlags().String("user", "root", "DB user name")
	rootCmd.PersistentFlags().String("pass", "", "DB password")
	rootCmd.PersistentFlags().String("host", "localhost", "DB host")
	rootCmd.PersistentFlags().String("port", "3306", "DB port")
	rootCmd.PersistentFlags().String("prefix", "", "DB table prefix. Will be deleted on snapshot and added on restore, thus you can have dev tables with one prefix or without and prod tables with other prefix.")
	rootCmd.PersistentFlags().String("file", "", "File to save snapshot or to restore from (default to DB scheme name)")
	rootCmd.PersistentFlags().Bool("d-table", true, "Delete deleted tables")
	rootCmd.PersistentFlags().Bool("d-index", true, "Delete deleted indexes")
[...]

PR #1147, opened 2021-06-01
fix: made `InConfig` process paths correctly

VOvchinnikov:
Fixes https://github.com/spf13/viper/issues/884

`InConfig` starts to process paths instead of treating them strictly as keys to the `v.config` map


sagikazarmark (approved):
Thanks!

PR #1167, opened 2021-07-16
Improve encoding layer

sagikazarmark:
This is an attempt to further improve the encoding layer specified in #888 and introduced in #986.

Major changes introduced by this PR:

- Encoder/Decoder interfaces now accept `map[string]interface{}` instead of just `interface{}`. The rationale behind that change is that the encoding library is supposed to be used with Viper (it's not a generic library), so it's okay to make the interface Viper specific.
- Two additional codec implementations:
  - INI
  - Java properties
- Encoder/decoder registries are now instantiated per Viper instance. It's required for some codecs (INI, Java props)

The PR also adds some basic tests to the existing encoder implementations.

A couple notes and concerns regarding this change:

Both INI and Java properties require the key delimiter to do some key mutation magic. It's required because both file formats contain key-value pairs (unlike the more common formats like YAML or JSON where you can have more nested configurations). As a result, the codecs have to be instantiated per Viper instance (since the key delimiter can be changed on instantiation).

In Viper 2 this should be a separate configuration of the codec, not tied to Viper's internal key delimiter config to allow storing keys in different formats.

---

Another problem introduced by the Java properties codec is that it's actually stateful: Viper saves the properties structure so it can write back properties files preserving comments and order. This is problematic, because the interface itself is designed for stateless codec instances.

For Viper 2 I think we should drop this feature, because it makes things a lot more complicated. We can hack things around for Viper 1 (eg. create a codec for Viper 1 specifically or make statefullness optional). I can't think of any viable alternative than dropping this feature for Java props. It's not like the most actively used file format and we want to discourage saving config from Viper anyway.

## To Do

- [x] Make tests more robust based on examples in Viper tests (eg. test more INI features)
- [x] Improve the Java properties codec
- [x] ~~Java properties will probably have to be added back to Viper, since codec registries can be reset. Not sure if it's a problem though.~~ No need to reset codecs (for now).
- [x] Implement a dotenv codec

## Misc

Fixes #923

### DIFF
--- a/internal/encoding/decoder.go
+++ b/internal/encoding/decoder.go
@@ -4,10 +4,10 @@ import (
 	"sync"
 )
 
-// Decoder decodes the contents of b into a v representation.
+// Decoder decodes the contents of b into v.
 // It's primarily used for decoding contents of a file into a map[string]interface{}.
 type Decoder interface {
-	Decode(b []byte, v interface{}) error
+	Decode(b []byte, v map[string]interface{}) error
 }
 
 const (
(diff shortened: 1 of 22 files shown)


sagikazarmark:
I decided to merge this PR. We can always make changes to it before we finalize the API.

PR #1180, opened 2021-08-04
Fix overriding of keys in slices with environment variables

AriehSchneier:
This ensures that keys in slices are expanded into the results of AllKeys().
But when merged back into a map in AllSettings() it will convert it back to a slice - this allows it to still be handled correctly by mapstructure.


tunglt-sm:
hi guys, are there any update about this PR?

Issue #1163, opened 2021-06-28
Choosing config file with correct extension/configType


jqk:
**Expected behavior (what you expected to happen)**:

search config file according to specified configType.

**Actual behavior (what actually happened)**:

I have 2 config files config.json and config.yaml, in the same path at the same time. Only file extensions are different.

I got json config file when I set empty string or 'json' for configType. That's what I expected.
But I still got json file when I set configType to 'yaml'.

**Repl.it link**:

**Code reproducing the issue**:

```go
viper.SetConfigName("config") 
viper.SetConfigType("yaml")
```
For above code, I got config.json opened. That's wrong for me.

I can open config.yaml when running code below:

```go
viper.SetConfigName("config.yaml") 
viper.SetConfigType("yaml")
```

I found what's happened in source code.

```go
func (v *Viper) searchInPath(in string) (filename string) {
	jww.DEBUG.Println("Searching for config in ", in)
	for _, ext := range SupportedExts {
		jww.DEBUG.Println("Checking for", filepath.Join(in, v.configName+"."+ext))
		if b, _ := exists(v.fs, filepath.Join(in, v.configName+"."+ext)); b {
			jww.DEBUG.Println("Found: ", filepath.Join(in, v.configName+"."+ext))
			return filepath.Join(in, v.configName+"."+ext)
		}
	}

	if v.configType != "" {
		if b, _ := exists(v.fs, filepath.Join(in, v.configName)); b {
			return filepath.Join(in, v.configName)
		}
	}

	return ""
}
```

Put the code below prior to 'for _, ext' section will solve the problem.

```go
	if v.configType != "" {
		if b, _ := exists(v.fs, filepath.Join(in, v.configName+"."+v.configType)); b {
			return filepath.Join(in, v.configName+"."+v.configType)
		}
	}
```

**Environment**:
- Viper version: 1.8.1
- Config source: 
- File format: 

**Anything else we should know?**:

None

sagikazarmark:
The current workaround for this problem is reordering the `SupportedExts` that contains the file extensions Viper looks for.

The configuration for searching files is already complicated enough and I'm not sure adding another setter is a good solution. Let me think about it. Till then, see my suggested workaround.

photowey:
Ah, I had the same question. 
version: v1.11.0.

the search folder contains toml and yml files with the same name.

configType: yml

search result: xxx.toml
unmarshal error: cannot unmarshal !!seq into map[string]interface {}

[...]

PR #1198, opened 2021-08-27
Add ConfigExtension

phinnaeus:
This forces Viper to look for config files with a given extension
which short circuits the supported extensions checks.

See https://github.com/spf13/viper/issues/1163


sagikazarmark:
Thanks for your contribution!

I believe #1849 is a better long-term approach to the problem.

Let me know what you think!

phinnaeus:
LGTM!

PR #1204, opened 2021-09-03
upgrade gopkg.in/yaml.v2 to gopkg.in/yaml.v3

inteon:


sagikazarmark:
Looks like upgrading breaks some of the tests.

inteon:
@sagikazarmark the failing tests were due to indentation issues.
I updated the behavior to match the old indentation as close as possible & changed the indentation in the remaining tests.

internal/encoding/yaml/codec.go
```diff
@@ -1,12 +1,20 @@
...
+)
 
 // Codec implements the encoding.Encoder and encoding.Decoder interfaces for YAML encoding.
 type Codec struct{}
 
 func (Codec) Encode(v interface{}) ([]byte, error) {
-	return yaml.Marshal(v)
+	var buf bytes.Buffer
+	encoder := yaml.NewEncoder(&buf)
+	encoder.SetIndent(1)
```
sagikazarmark, line 15: What's the default value for this? Would it make sense to use it instead?

sagikazarmark (changes requested):
Can you please rebase from master?

internal/encoding/yaml/codec.go
inteon: @sagikazarmark I created a new PR using these defaults & rebased on master: #1264

PR #1212, opened 2021-09-13
Update README.md

ZhengHe-MD:
~~convert tab to space in line 738~~
convert spaces to tab in all code segments to correct the alignment, for example

before: 

```go
type config struct {
	Chart struct{
        Values map[string]interface{}
    }
}
```

after:
```go
type config struct {
    Chart struct{
        Values map[string]interface{}
    }
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -735,7 +735,7 @@ v.SetDefault("chart::values", map[string]interface{}{
 })
 
 type config struct {
-	Chart struct{
+    Chart struct{
         Values map[string]interface{}
     }
 }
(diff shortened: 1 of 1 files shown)


sagikazarmark:
I don't think this is right, Go code is always tab-indented.

ZhengHe-MD:
@sagikazarmark Thanks for pointing that out, I was trying to keep consistent with the code segment, other lines use spaces. Yeah, I agree, Go code is always tab-indented. Ultimately I just wanted to make the lines aligned correctly. I can also convert all spaces in that segment to tabs instead. : )

ZhengHe-MD:
@sagikazarmark I open the option 'show whitespaces' of my IDE, and see both spaces and tabs used. I convert all spaces in Go segments to tabs, but feel free to check that yourself.

sagikazarmark:
Thanks @ZhengHe-MD !

PR #1235, opened 2021-10-03
chore(linter): enable errorlint

mmorel-35:


sagikazarmark:
Thanks!

This will likely conflict with #1167
Let's wait for that landing first

PR #1237, opened 2021-10-14
docs: fix error handling for file-not-found

fileformat:
I'm not a go expert, so I'm not sure exactly what the example was supposed to do before my change, but it definitely did not work.

Running on go1.17.2

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,7 +127,7 @@ You can handle the specific case where no config file is found like this:
 
 ```go
 if err := viper.ReadInConfig(); err != nil {
-	if _, ok := err.(viper.ConfigFileNotFoundError); ok {
+	if errors.Is(err, os.ErrNotExist) {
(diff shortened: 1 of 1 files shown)


sagikazarmark:
I believe `os.ErrNotExist` should only be returned when you use `SetConfigFile` and the file doesn't exist, so the documentation is technically correct. It might make sense to clean up the confusion with `SetConfigFile`.

fileformat:
I'm a go beginner, so I may be missing something, but `findConfigFile` isn't behaving the way I expect and `ReadInConfig` is definitely returning `os.ErrNotExist`.

I created a [gist](https://gist.github.com/fileformat/719f8558987c5a537a1b2db78eda12a0) that demonstrates what I think it should do vs what it actually does.

The output is in the gist as stderr.txt.  This is with go v1.17.2.

PR #1301, opened 2022-02-28
Add MustBindEnv

meowfaceman:
Adds in a MustBindEnv function which will panic if BindEnv fails.

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -1197,6 +1197,17 @@ func (v *Viper) BindEnv(input ...string) error {
 	return nil
 }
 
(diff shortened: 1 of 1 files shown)


meowfaceman:
I find myself using this pattern fairly frequently. It's small and didn't take much to implement, so if it's antithetical to the way viper wants to do things that's fine. :-) But figured I'd push up the PR and get feedback/try to get it in if there's interest.

sagikazarmark (approved):
Thanks!

PR #1307, opened 2022-03-06
Fix viper_go1_15.go if two files with the same name but different types in the same directory

XYZ0901:
If i have two config files in the same path, one's type is yaml and another is toml. But i set the type: `viper.SetConfigType("yaml")`, old viper_go1_15.go will return me the toml's config. Now the bug is fixed.

### DIFF
--- a/viper_go1_15.go
+++ b/viper_go1_15.go
@@ -27,17 +27,18 @@ func (v *Viper) findConfigFile() (string, error) {
 
 func (v *Viper) searchInPath(in string) (filename string) {
 	v.logger.Debug("searching for config in path", "path", in)
+	
+	if v.configType != "" && stringInSlice(v.configType,SupportedExts) {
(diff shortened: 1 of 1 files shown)


viper_go1_15.go
```diff
@@ -27,6 +27,13 @@ func (v *Viper) findConfigFile() (string, error) {
 
 func (v *Viper) searchInPath(in string) (filename string) {
 	v.logger.Debug("searching for config in path", "path", in)
+	
+	if v.configType != "" && stringInSlice(v.configType,SupportedExts) {
```
sagikazarmark, line 31: Can you please provide some tests providing that it works and that it doesn't break previous behavior? (I believe it actually does thanks to that second condition)

sagikazarmark (changes requested):

Issue #944, opened 2020-08-06
Missing entries after loading a .properties file


htmfilho:
I'm loading the following `.properties` file:

    item.value.cd.inconsistency=aasdasd
    btn.add.item.value=yffdsd
    title.add.item.value=yfffdd
    title.edit.item.value=yewee
    item.value.name=ndfaso
    item.value.points.caption=nqqqo
    item.value.display.order.caption=nddddo
    item.value.created.msg=novfff
    item.value.updated.msg=noxxxxx
    item.value.deleted.msg=noaaaaa
    item.value.not.deleted.msg=nodsasd
    item.value=noffsdf
    complianceItemValue.nameLang1.blank=dfgg
    visit.item.type.value.invalid=yesjgh
    complianceItemValue.nameLang1.unique=yrttres
    inactive.item.value.caption=yehghjs
    item.value.not.active=yohjkes

and showing the list of loaded keys using this code:

    config := viper.New()
    config.SetConfigFile("test.properties")
    config.SetConfigType("properties")
    keys := config.AllKeys()
    sort.Strings(keys)
    fmt.Printf("Size: %v\n", len(keys))
    for _, key := range keys {
        fmt.Printf("%v\n", key)
    }

printing this result:

    Size: 8
    btn.add.item.value
    complianceitemvalue.namelang1.blank
    complianceitemvalue.namelang1.unique
    inactive.item.value.caption
    item.value.not.active
    title.add.item.value
    title.edit.item.value
    visit.item.type.value.invalid

As you can see, there are several entries missing. Can you explain what is happening? Can you point me to the code that does it?

Thank you very much in advance.

htmfilho:
Something happens in the `for` loop at `viper.go:1515`. It kind of builds a tree of entries and at some point branches of the tree are lost.

htmfilho:
Ok. I found the problem. When I try to load these entries:

    item.value.name=ndfaso
    item.value.points.caption=nqqqo
    item.value.display.order.caption=nddddo
    item.value.created.msg=novfff
    item.value.updated.msg=noxxxxx
    item.value.deleted.msg=noaaaaa
    item.value.not.deleted.msg=nodsasd
    item.value=noffsdf

Viper gets loaded a single entry:

    item.value=noffsdf

Why? Because while it loads the tree it reaches the key `item.value` and loses all previously attached nodes (`name`, `points`, `display`, `created`, `updated`, `deleted`, and `not`).

I don't think this is desirable behavior, is it?

Issue #640, opened 2019-01-29
deepSearch overide properties value


shalk:
hello , I have a properties file `1.properties` . I want to  get  value `foo` from key `test`, 
But  viper 's method  `deepSearch`  will replace it with a new map when i have a map under the path
does it make sense or there are other workaround? 
Thank you

1.properties
```
test=foo
test.1=abc
test.2=efg
```
and code 

```golang
package main

import (
	"bytes"
	"fmt"
	"github.com/spf13/viper"
)

func main()  {
	typ := "properties"
	viper.SetConfigType(typ)
	str := `
test=foo
test.1=abc
test.2=efg
`
	viper.ReadConfig(bytes.NewBuffer([]byte(str)))
	// expect: foo , actual: empty  ;  override foo
	fmt.Println(viper.GetString("test"))
	// get map
	fmt.Println(viper.Get("test"))
}
```
result:
```

map[1:abc 2:efg]
```

PR #1315, opened 2022-03-24
Preventing properties to override each other with Java properties

bu3:
Using Viper in one of our team projects, I noticed that Java properties config has issues at loading a configuration like the following one:
```
app.env=development
app.consumerbroker=first.example.com:9092
app.producerbroker=second.example.com:9092
app.producerbroker.foo=foo
```
Viper returns only:
```
app.env : development
app.producerbroker :
app.producerbroker.foo : foo
app.consumerbroker : first.example.com:9092
```
This PR is meant to solve that.

Let me know what you think about it.

I found open issues related to this topic:
https://github.com/spf13/viper/issues/944
https://github.com/spf13/viper/issues/640

### DIFF
--- a/internal/encoding/javaproperties/codec.go
+++ b/internal/encoding/javaproperties/codec.go
@@ -2,11 +2,9 @@ package javaproperties
 
 import (
 	"bytes"
-	"sort"
-	"strings"
-
 	"github.com/magiconair/properties"
 	"github.com/spf13/cast"
+	"sort"
 )
 
 // Codec implements the encoding.Encoder and encoding.Decoder interfaces for Java properties encoding.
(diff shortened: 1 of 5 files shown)


bu3:
Hey @sagikazarmark, sorry to bother you directly but I noticed you being quite active on this repo's PRs.
Would you mind taking a look at this change I sent? Thanks a million

bu3:
Sorry to bother you again @sagikazarmark! Any chances to have my PR reviewed?

sagikazarmark:
Hi @bu3 !

Thanks for writing a PR, although I have to say I don't really see how this could end up in Viper:

- It's a massive BC break
- It changes something that's not really a bug, but a feature for this library

Also, java props support will most probably removed in Viper 2, so I'd suggest switching to a different config language.

sagikazarmark:
Although on the surface Viper might seem like a key-value store, it really isn't. The internal data representation is mostly a nested map, not a flat one (although unfortunately some data sources are read as flat maps, eg. flags).

Nested maps are closer to the best practices of Viper which is to unmarshal data to config structs.

Here is what I can do for you (although I don't know when that's going to happen): we plan to open up the encoder interface soon, meaning you could register your own encoders with your own logic in Viper.

When that happens, you can use this implementation instead of the current one.

PR #1316, opened 2022-03-26
Allow marshal to bytes buffer

nobiit:


nobiit:
Thanks for the idea of @JulienBreux ,
I have updated the code. For your example you can refer to this https://github.com/nobiit/viper/blob/master/viper_test.go#L2003-L2015

jerome-diver:
Hi @nobiit , @JulienBreux , i can not find these "new" method you did created in the viper package. We are one year half later from your commit. What's happening in the process to be integrated to the official viper package, i do not understand ? Or maybe i have something special at import time to get use your commited code ? (Thank you for your commit, hope it will be accessible soon)

PR #1337, opened 2022-04-27
Correct GetStringMap inconsistency, fixes #708

### LINKED ISSUE #708: Incosistant merging with GetStringMapString and environment variables
When getting a value with GetStringMapString it produced a different value to AllSettings and Get/GetString when the value is set with an environment variable.

```package main
 
import "strings"
import "log"
import "github.com/spf13/viper"

func main() {
    viper.SetEnvPrefix("test")
    viper.SetEnvKeyReplacer(strings.NewReplacer(".", "_"))
    viper.AutomaticEnv()
    
    viper.SetDefault("some.key", "default")
    
    log.Printf("AllSettings: %#v", viper.AllSettings())
    
    log.Printf("GetStringMapString('some.key'): %#v", viper.GetStringMapString("some"))
    log.Printf("GetString('some.key'): %#v", viper.GetString("some.key"))
    log.Printf("Get('some.key'): %#v", viper.Get("some.key"))
}
```
When executing produces:

```
$ TEST_SOME_KEY=env go run main.go
2019/05/28 12:27:01 AllSettings: map[string]interface {}{"some":map[string]interface {}{"key":"env"}}
2019/05/28 12:27:01 GetStringMapString('some.key'): map[string]string{"key":"default"}
2019/05/28 12:27:01 GetString('some.key'): "env"
2019/05/28 12:27:01 Get('some.key'): "env"
```
When I expected the `GetStringMapString` to output the same value as `GetString`, `Get` and `AllSettings`, instead it uses the default value.

When you throw a config into the mix you get the value from the config rather than the default:

```
package main
 
import "strings"
import "log"
import "fmt"  
import "github.com/spf13/viper"

func main() {
    viper.SetEnvPrefix("test")
    viper.SetEnvKeyReplacer(strings.NewReplacer(".", "_"))
    viper.AutomaticEnv()
    
    viper.SetDefault("some.key", "default")
    
    viper.SetConfigFile("./config.yml")  
    err := viper.ReadInConfig()  
    if err != nil {  
            log.Fatal(fmt.Errorf("fatal error config file: %s", err))
    }    
    
    log.Printf("AllSettings: %#v", viper.AllSettings())
    
    log.Printf("GetStringMapString('some.key'): %#v", viper.GetStringMapString("some"))
    log.Printf("GetString('some.key'): %#v", viper.GetString("some.key"))
    log.Printf("Get('some.key'): %#v", viper.Get("some.key"))
}
```
```
$ TEST_SOME_KEY=env go run main.go
2019/05/28 12:28:26 AllSettings: map[string]interface {}{"some":map[string]interface {}{"key":"env"}}
2019/05/28 12:28:26 GetStringMapString('some.key'): map[string]string{"key":"config"}
2019/05/28 12:28:26 GetString('some.key'): "env"
2019/05/28 12:28:26 Get('some.key'): "env"
```

rikkuness:
I don't know if this is the best way to tackle this, adding the whole new method `allSettingsUnderParent()` which is almost a straight copy of `AllSettings()` but filtering to a given sub key. I'm open to suggestions if there's a cleaner way to achieve this.

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -1041,21 +1041,21 @@ func (v *Viper) GetStringSlice(key string) []string {
 func GetStringMap(key string) map[string]interface{} { return v.GetStringMap(key) }
 
 func (v *Viper) GetStringMap(key string) map[string]interface{} {
-	return cast.ToStringMap(v.Get(key))
+	return cast.ToStringMap(v.allSettingsUnderParent(key))
 }
 
 // GetStringMapString returns the value associated with the key as a map of strings.
 func GetStringMapString(key string) map[string]string { return v.GetStringMapString(key) }
 
 func (v *Viper) GetStringMapString(key string) map[string]string {
-	return cast.ToStringMapString(v.Get(key))
+	return cast.ToStringMapString(v.allSettingsUnderParent(key))
 }
 
 // GetStringMapStringSlice returns the value associated with the key as a map to a slice of strings.
 func GetStringMapStringSlice(key string) map[string][]string { return v.GetStringMapStringSlice(key) }
 
 func (v *Viper) GetStringMapStringSlice(key string) map[string][]string {
-	return cast.ToStringMapStringSlice(v.Get(key))
+	return cast.ToStringMapStringSlice(v.allSettingsUnderParent(key))
 }
 
 // GetSizeInBytes returns the size of the value associated with the given key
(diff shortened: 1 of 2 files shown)


Ran-Xing:
@rikkuness I don't think the problem is solved

PR #1341, opened 2022-05-05
chore: fix Error log calls in mergeMaps

wwade:
The logger interface uses k,v pairs, but it was being called with a
format string, here.

Changed to use k,v pairs.

I didn't find any other instances of this problem.


wwade:
@sagikazarmark mind having a look? just a small fix.

sagikazarmark (approved):
Thanks!

PR #1353, opened 2022-05-25
DONTMERGE: Initial thread-safe implementation

david-imola:
Adds a `RWLock` to viper and simple `Lock()/Unlock()` calls to most
relevant routines. Full coverage for remote config and config watching
is TBD.


david-imola:
Sorry, ignore

PR #1356, opened 2022-05-26
Add etcd3 support to remote

sagikazarmark:


Nebell:
Remote provider "etcd3" is not in SupportedRemoteProviders of [viper.go](https://github.com/spf13/viper/blob/b1b8b3d4d484989aceb420deaf55e2a702780acb/viper.go#L422)

PR #1379, opened 2022-06-10
:warning: removing infected yaml.v2 which suffers from CVE-2022-28948

AlmogBaku:
Removing infected `yaml.v2` which suffers from CVE-2022-28948


sagikazarmark:
Thanks for the PR @AlmogBaku !

While I understand why you want to see it gone, YAML v3 is not fully backwards compatible with YAML v2. As a result, we need a longer waiting period before we can safely remove it.

Rest assured, unless you apply the `yaml2` build tag no v2 code will end up in your binary.

I'm going to close this for now, but you can track the status of this issue in #1399

PR #1393, opened 2022-06-27
Use jsonc in markdown codeblocks for better readability

HurSungYun:
## AS-IS

스크린샷 2022-06-28 오전 1 11 56

## TO-BE

스크린샷 2022-06-28 오전 1 15 42

use `jsonc` instead of `json` to improve readability


sagikazarmark (approved):
Thanks!

PR #1396, opened 2022-07-04
remove dependency to go-toml/v1

majst01:
Found in my project [metalctl](https://github.com/metal-stack/metalctl/blob/master/go.mod#L72) two indirect dependencies to the same go-toml library, one with v1, and one with v2. 

Current core is compatibly to v2, so lets remove the v1 variant.


majst01:
Already recognized that this is by design, closing and sorry for the noise

PR #1405, opened 2022-07-19
Adds support for uint16 with `GetUint16`

oxisto:
We have encountered numerous places where it is convenient to have viper return a `uint16` value, especially in combination with the new `netip` package  that represents a port correctly as `uint16` rather than just an `int`. cobra already supports this, but we need a conversion from the existing `GetUint` method in viper.

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -991,6 +991,13 @@ func (v *Viper) GetUint(key string) uint {
 	return cast.ToUint(v.Get(key))
 }
 
+// GetUint16 returns the value associated with the key as an unsigned integer.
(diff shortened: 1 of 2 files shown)


sagikazarmark (approved):
LGTM, thanks @oxisto

sagikazarmark (changes requested):
@oxisto actually, can you please rebase your PR to trigger CI?

oxisto:
> @oxisto actually, can you please rebase your PR to trigger CI?

done.

sagikazarmark (approved):
Thanks!

Issue #896, opened 2020-04-20
Debug should accept a Writer to specify output location besides Stdout


cixel:
When using Viper in a client application, it can be useful to log configuration registries as a diagnostic, to understand what the client did. In situations where we don't want to log to stdout, it would be nice to be able to provide an `io.Writer` for Viper to use instead.

sagikazarmark:
Fixed in #1414

PR #1414, opened 2022-08-11
Add a DebugTo convenience funtion

bcrochet:
One might want to write the debug information somewhere other than
Stdout. This patch adss a DebugTo function and method, that accepts
an io.Writer. It changes the original Debug implementation to call
this new function with a default of os.Stdout, which maintains
backward compatibility.

Signed-off-by: Brad P. Crochet <brad@redhat.com>

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -2135,14 +2135,17 @@ func (v *Viper) getConfigFile() (string, error) {
 
 // Debug prints all configuration registries for debugging
 // purposes.
-func Debug() { v.Debug() }
+func Debug()              { v.Debug() }
+func DebugTo(w io.Writer) { v.DebugTo(w) }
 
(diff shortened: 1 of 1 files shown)


bcrochet:
I see that #897 also addresses this, just with a different name. Happy to close this one if that name is preferred. Otherwise, it would be nice to see one of these make it in.

sagikazarmark:
I prefer `DebugTo` (for reasons I can't explain), so fixes #897

sagikazarmark (approved):
Thanks @bcrochet

sagikazarmark:
Fixes #896

sagikazarmark:
@bcrochet can you please take a look at the breaking lint?

sagikazarmark (approved):

Issue #935, opened 2020-07-06
Is it possible to rewrite the value in the structure slice from the environment variable /w automatic env?


tmarcus87:
Hello everyone.

Currently, I use viper to configure the k8s app, and pass sensitive information (DB password etc.) from environment variables.
However, when the configuration is a structure slice, I cannot successfully overwrite the value with the environment variable.

Is this something my settings are missing?
Or, in case of structure slice, can't it be overwritten from environment variable?

sample code here
```
package main

import (
	"encoding/json"
	"fmt"
	"github.com/mitchellh/mapstructure"
	"github.com/spf13/viper"
	"os"
	"strings"
)

var ConfigYaml = `
foo:
  host: host0
  port: 3306
  user: user0
  pass: ""
bar:
  - host: host1
    port: 3306
    user: user1
    pass: ""
  - host: host2
    port: 3306
    user: user2
    pass: ""
`

type Config struct {
	Foo ConfigValue   `yaml:"foo"`
	Bar []ConfigValue `yaml:"bar"`
}

type ConfigValue struct {
	Host string `yaml:"host"`
    Port uint16 `yaml:"port"`
    User string `yaml:"user"`
    Pass string `yaml:"pass"`
}

func main() {
	v := viper.New()

	wd, _ := os.Getwd()
	v.AddConfigPath(wd)
	v.SetConfigType("yaml")
	v.SetConfigName("config")

	v.AutomaticEnv()
	v.SetEnvKeyReplacer(strings.NewReplacer(".", "_"))

	if err := v.ReadConfig(strings.NewReader(ConfigYaml)); err != nil {
		panic(err)
	}

	cfg := Config{}
	if err := v.Unmarshal(&cfg, func(cfg *mapstructure.DecoderConfig) { cfg.TagName = "yaml" }); err != nil {
		panic(err)
	}

	bs, _ := json.MarshalIndent(cfg, "", "  ")
	fmt.Println(string(bs))
}
```

run with environment variable.
```
mylaptop $ FOO_PASS=pass0  BAR_0_PASS=pass1  BAR_1_PASS=pass2 go run main.go
{
  "Foo": {
    "Host": "host0",
    "Port": 3306,
    "User": "user0",
    "Pass": "pass0"            // <- override OK
  },
  "Bar": [
    {
      "Host": "host1",
      "Port": 3306,
      "User": "user1",
      "Pass": ""               // <- override NG
    },
    {
      "Host": "host2",
      "Port": 3306,
[...]

PR #1429, opened 2022-09-06
Implement viper.BindStruct for automatic unmarshalling from environment variables

### LINKED ISSUE #188: Consider env vars when unmarshalling
Is this possible?

**UPDATE: yes it is, see https://github.com/spf13/viper/issues/188#issuecomment-255519149**

``` go

type Config struct {
    BindPort int `mapstructure:"port" yaml:"port,omitempty"`
}

var c Config

// ...
viper.AutomaticEnv()

    // ...
}

if err := viper.Unmarshal(&c); err != nil {
    // ...
}

// I want c.BindPort == viper.Get("PORT") == os.Getenv("PORT")
```

### LINKED ISSUE #584: Does viper require a config file in order to use ENV variables?
Hi, I have an application that can use both a config file and env variable.

My `initConfig()` method looks like this:
```
if err := viper.ReadInConfig(); err == nil {
	fmt.Println("Using config file:", viper.ConfigFileUsed())
} else {
	fmt.Println("Config file not found:", cfgFile)
}

viper.SetEnvPrefix("GC")
viper.AutomaticEnv()

err := viper.Unmarshal(&config)
if err != nil {
	fmt.Println(err.Error())
	os.Exit(1)
}
```

`unmarshal` drops the config into a struct, and it works well. 

I am facing an issue where if my config file does not exist (or is empty), viper seems to completely ignore any environment variables. I have verified that viper can see the env variables, by printing them to standard out. My workaround is to deploy the config file, with all the correct keys but no values. Viper then correctly overrides the config file with my environment variables. 

Am I wrong to assume that Viper should be able to use environment variables when no config file is found?

### LINKED ISSUE #688: Unmarshal using ENV without binding
Hi all,

Having this code:
```go
type RpcConfig struct {
	ServerAddr string `mapstructure:"ADDR"`
}

func main() {
	v := viper.New()
	v.SetEnvKeyReplacer(strings.NewReplacer(".", "_"))
	v.AutomaticEnv()

	//v.BindEnv("ADDR")

	c := &RpcConfig{}
	if err := v.Unmarshal(c); err != nil {
		panic(err)
	}

	log.Printf("conf=%+v", c)
}
```
and running in this way:
```
ADDR="test" go run main.go
```
I got config with empty `ServerAddr`. If I remove comment from binding line it works but why viper couldn't get it from `mapstructure` tag itself?

Please let me know if I doing something wrong.

### LINKED ISSUE #725: Commented out property in YAML yields inconsistencies between Get and Unmarshal when overriding using environment variables
Unmarshal does not resolve the correct value when using environment variable and a yaml config file with commented out property.
To reproduce, see the following unit test:

```go
func Test(t *testing.T) {
	type FooConfig struct {
		Bar string
	}
	type Config struct {
		Foo FooConfig
	}

	yamlWithValueCommented := `---
foo:
#  bar: "value from file"`

	// Overriding with OS Environment Variable
	os.Setenv("FOO_BAR", "value from env")

	v := viper.New()
	v.SetEnvKeyReplacer(strings.NewReplacer(".", "_"))
	v.AutomaticEnv()
	v.SetConfigType("yaml")
	v.SetDefault("foo.bar", "default value")
	v.ReadConfig(strings.NewReader(yamlWithValueCommented))
	c := Config{}
	v.Unmarshal(&c)
	fmt.Println(v.Get("foo.bar")) // As expected: "value from env"
	fmt.Println(c.Foo.Bar) // Expected: "value from env", Actual: "default value"
}
```
Output:
```
=== RUN   Test
value from env
default value
--- PASS: Test (0.00s)
```

----
The full test, which includes some sanity tests:

```go
func Test(t *testing.T) {
	type FooConfig struct {
		Bar string
	}
	type Config struct {
		Foo FooConfig
	}

	yamlWithValue := `---
foo:
  bar: "value from file"`

	yamlWithValueCommented := `---
foo:
#  bar: "value from file"`

	v := viper.New()
	v.SetConfigType("yaml")
	v.SetDefault("foo.bar", "default value")
	v.ReadConfig(strings.NewReader(yamlWithValue))
	c := Config{}
	v.Unmarshal(&c)
	fmt.Println(v.Get("foo.bar")) // As expected: "value from file"
	fmt.Println(c.Foo.Bar) // As expected: "value from file"

	v = viper.New()
	v.SetConfigType("yaml")
	v.SetDefault("foo.bar", "default value")
	v.ReadConfig(strings.NewReader(yamlWithValueCommented))
	c = Config{}
	v.Unmarshal(&c)
	fmt.Println(v.Get("foo.bar")) // As expected: "default value"
	fmt.Println(c.Foo.Bar) // As expected: "default value"

	// Now overriding with OS Environment Variable
	os.Setenv("FOO_BAR", "value from env")

	v = viper.New()
	v.SetEnvKeyReplacer(strings.NewReplacer(".", "_"))
	v.AutomaticEnv()
	v.SetConfigType("yaml")
	v.SetDefault("foo.bar", "default value")
	v.ReadConfig(strings.NewReader(yamlWithValue))
	c = Config{}
	v.Unmarshal(&c)
	fmt.Println(v.Get("foo.bar")) // As expected: "value from env"
	fmt.Println(c.Foo.Bar) // As expected: "value from env"

	v = viper.New()
	v.SetEnvKeyReplacer(strings.NewReplacer(".", "_"))
	v.AutomaticEnv()
	v.SetConfigType("yaml")
	v.SetDefault("foo.bar", "default value")
	v.ReadConfig(strings.NewReader(yamlWithValueCommented))
	c = Config{}
	v.Unmarshal(&c)
	fmt.Println(v.Get("foo.bar")) // As expected: "value from env"
	fmt.Println(c.Foo.Bar) // Expected: "value from env", Actual: "default value"
}
```
Output:
```
=== RUN   Test
value from file
value from file
default value
default value
value from env
value from env
value from env
default value
--- PASS: Test (0.00s)
```

### LINKED ISSUE #761: Unmarshal non-bound environment variables
This is going to be an umbrella issue for other open issues for this problem.

These are:

- #1236
- #725
- #688
- #584
- #212
- #188
- #140

### Problem

Currently Viper cannot unmarshal automatically loaded environment variables.

Here is a failing test case reproducing the issue:

```go
package viper

import (
	"os"
	"testing"
)

func TestUnmarshal_AutomaticEnv(t *testing.T) {
	os.Setenv("MYKEY", "myvalue")

	type Config struct {
		MyKey string `mapstructure:"mykey"`
	}

	var c Config

	AutomaticEnv()

	if err := Unmarshal(&c); err != nil {
		t.Fatal(err)
	}

	if c.MyKey != "myvalue" {
		t.Error("failed to unmarshal automatically loaded environment variable")
	}

	os.Clearenv()
}
```

### Root cause

Automatic environment variable loading works as follows:

1. User requests a value from Viper: `viper.GetString("mykey")`
2. Viper runs the env key replacer on the key
3. Viper compares the key (case insensitively) with the list of environment variables
4. Returns the value if one is found

Unmarshal, however, doesn't work by requesting a key, quite the opposite: it tries to unmarshal existing keys onto the given structure. Since automatically loaded environment variables are not in the list of existing keys (unlike defaults and bind env vars), these variables won't be unmarshaled.

### Workaround

1. Use `BindEnv` to manually bind environment variables
2. Use `SetDefault` to set a default value (even an empty one), so Viper can find the key and try to find a matching env var

### Possible solutions

The linked issues list a number of workarounds to fix this issue.

Personally, I'm in favor of the following:

1. Unmarshal the given struct onto a `map[string]interface{}`. That will give use a nested structure of all the keys otherwise mapstructure expects.
2. Flatten the structure, so we get a list of expected keys
3. Check that every key is defined. If not, check if there is a matching environment variable.
4. Merge the available settings with the matching, automatically loaded environment variables.
5. Use mapstructure as usual

Although I think that this is currently the best available solution, it doesn't mean we can't do better, so please do suggest if you have an idea.

### LINKED ISSUE #1012: UnmarshalKey / Sub only uses read config, neglects environment variables

### LINKED ISSUE #1365: Viper does not unmarshal values passed from env or pflag
### Preflight Checklist

- [X] I am not looking for support or already pursued the available [support channels](https://github.com/spf13/viper/issues/new/choose) without success.
- [X] I have checked the [troubleshooting guide](https://github.com/spf13/viper/blob/master/TROUBLESHOOTING.md) for my problem, without success.

### Viper Version

1.11.0

### Go Version

1.17

### Config Source

Flags, Environment variables

### Format

_No response_

### Repl.it link

_No response_

```go
package main

import (
	"fmt"

	"github.com/spf13/cobra"
	"github.com/spf13/viper"
)

type Config struct {
	AFlag string `mapstructure:"A_FLAG"`
}

// rootCmd is the root cobra command of ormos-web
var rootCmd = &cobra.Command{
	Use:   "myapp",
	Short: "My app",
	Run: func(cmd *cobra.Command, args []string) {
		fmt.Println("running")
	},
}

func InitConfig(cmd *cobra.Command) {
	var conf Config
	v := viper.New()
	v.AddConfigPath(".")
	v.SetConfigName(".env")
	v.SetConfigType("env")
	v.SetEnvPrefix("FOO")
	v.AllowEmptyEnv(true)
	v.AutomaticEnv()
	v.BindPFlags(cmd.Flags())
	v.Unmarshal(&conf)
	fmt.Printf("%+v\n", conf)
}

func main() {
	rootCmd.PersistentFlags().StringP("a-flag", "", "", "a flag")
	InitConfig(rootCmd)
	rootCmd.Execute()
}
```

### Expected Behavior

after 

```
export FOO_A_FLAG=test
```

or either running as

```
go run main.go --a-flag=test
```

the program should be printing

```
{AFlag:test}
running
```

### Actual Behavior

the program just prints

```
{AFlag:}
running
```

### Steps To Reproduce

_No response_

### Additional Information

_No response_

### LINKED ISSUE #1522: AutomaticEnv function failing to read environment variables in Go code
### Preflight Checklist

- [X] I am not looking for support or already pursued the available [support channels](https://github.com/spf13/viper/issues/new/choose) without success.
- [X] I have checked the [troubleshooting guide](https://github.com/spf13/viper/blob/master/TROUBLESHOOTING.md) for my problem, without success.

### Viper Version

1.15.0

### Go Version

1.20.1

### Config Source

Environment variables

### Format

Other (specify below)

### Repl.it link

_No response_

```go
package main

import (
	"fmt"
	"log"

	"github.com/spf13/viper"
)

type Config struct {
	FbClientId string `mapstructure:"FB_CLIENT_ID"`
}

func LoadConfig() (config Config, err error) {
	viper.AutomaticEnv()

	err = viper.Unmarshal(&config)
	if err != nil {
		return
	}
	return
}

func main() {
	type Config struct {
		FbClientId string `mapstructure:"FB_CLIENT_ID"`
	}

	cfg, err := LoadConfig()
	if err != nil {
		log.Fatal(err)
	}
	fmt.Print(cfg.FbClientId)
}
```

### Expected Behavior

After running this command, I expect `fmt.Print(cfg.FbClientId)` to log `123`:
```sh
export FB_CLIENT_ID="123" && go run main.go
```

### Actual Behavior

There is no output in the console.

### Steps To Reproduce

_No response_

### Additional Information

_No response_

### LINKED ISSUE #1545: The env attribute cannot be specified as an attribute of a struct
### Preflight Checklist

- [X] I am not looking for support or already pursued the available [support channels](https://github.com/spf13/viper/issues/new/choose) without success.
- [X] I have checked the [troubleshooting guide](https://github.com/spf13/viper/blob/master/TROUBLESHOOTING.md) for my problem, without success.

### Viper Version

1.15.0

### Go Version

1.17.8

### Config Source

Environment variables

### Format

Other (specify below)

### Repl.it link

_No response_

```go
func main(){
        type C struct {
		HOME string `mapstructure:"HOME"`
	}

	c := C{}

	viper.AutomaticEnv()
	err := viper.Unmarshal(&c)
	if err != nil {
		panic(err)
	}
	fmt.Println(c.HOME)
	fmt.Println(viper.Get("HOME"))

}
```

### Expected Behavior

I want to get the value of home from the structure
output:
/Users/cy
/Users/cy

### Actual Behavior

output:

/Users/cy

### Steps To Reproduce

computer:
macOS Catalina version: 10.15.7(19h15)
MacBook Pro (15-inch, 2018)

### Additional Information

_No response_

### LINKED ISSUE #1587: Use environment variables with no need to have config file (yml,josn,etc...)
### Preflight Checklist

### Problem Description

As an avid user of the Viper package, I find it incredibly useful for managing configuration settings in my Go applications. However, I noticed that the current implementation requires explicitly declaring keys  in the YAML configuration files. This poses a limitation when it comes to utilizing environment variables efficiently, especially in containerized environments or cloud-based deployment scenarios.

### Proposed Solution

The suggested solution would involve extending Viper's capabilities to automatically look for environment variables corresponding to configuration keys when accessing the configuration values. If an environment variable with a matching name is found, Viper should fetch the value from the environment variable instead of the YAML file. This approach would provide more flexibility and better align with modern application deployment practices.

### Alternatives Considered

Simplified Configuration: Users can manage their application's configuration more efficiently, as there will be no need to update the YAML file every time a configuration value changes.

Containerization Support: With this enhancement, developers can easily configure their applications using environment variables, which is a common practice in containerized environments.

Enhanced Security: Storing sensitive information in environment variables instead of configuration files reduces the risk of accidental exposure.

Easier Cloud Integration: Cloud platforms often provide the ability to manage configuration via environment variables, making this feature essential for smooth integration with various cloud services.

### Additional Information

I understand that some users may prefer the current approach, where configuration keys are explicitly defined in YAML files. To maintain backward compatibility and ensure the flexibility of this feature, the solution could be implemented as an optional flag or a configuration option that users can enable or disable according to their specific requirements.

I noticed that other configuration libraries for Go, such as envconfig and godotenv, have implemented similar functionality. Studying their approaches might provide some insights and best practices for integrating environment variables seamlessly.

krakowski:
## Description

The current `viper` implementation does not allow to `viper.Unmarshal` environment variables into structs without binding them first (using `viper.BindEnv`). This behavior results in an "empty" struct (i.e no values filled) if no configuration file is provided with matching keys.

This pull request adds a new `viper.BindStruct` function which takes a struct pointer and binds each (nested) struct field to an environment variable using the struct field's name or the user-specified `mapstructure` tag. The implementation is quite simple, since I could reuse some functions:

```go
func (v *Viper) BindStruct(input interface{}) error {
	envKeysMap := map[string]interface{}{}
	if err := mapstructure.Decode(input, &envKeysMap); err != nil {
		return err
	}

	structKeys := v.flattenAndMergeMap(map[string]bool{}, envKeysMap, "")
	for key, _ := range structKeys {
		if err := v.BindEnv(key); err != nil {
			return err
		}
	}

	return nil
}
```

A simple example could look like this:

```go
type Configuration struct {
    Value int
}

os.Setenv("VALUE", 42)

var config Configuration
viper.BindStruct(&config)
viper.Unmarshal(&config)
```

Please see the [`TestBindStruct`](https://github.com/krakowski/viper/blob/feature/bind-struct/viper_test.go#L930-L1024) unit test for a more complete example.

## Related Issues
  * Fixes #188
  * Fixes #584
  * Fixes #688
  * Fixes #725 
  * Fixes #761
  * Fixes #1012 
  * Fixes #1365 
  * Fixes #1522 
  * Fixes #1545
  * Fixes #1587

##

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -2098,9 +2098,13 @@ outer:
 func AllSettings() map[string]any { return v.AllSettings() }
 
 func (v *Viper) AllSettings() map[string]any {
+	return v.getSettings(v.AllKeys())
+}
+
+func (v *Viper) getSettings(keys []string) map[string]any {
 	m := map[string]any{}
 	// start from the list of keys, and construct the map one value at a time
-	for _, k := range v.AllKeys() {
+	for _, k := range keys {
 		value := v.Get(k)
 		if value == nil {
 			// should not happen, since AllKeys() returns only keys holding a value,
--- a/viper.go
+++ b/viper.go
@@ -1111,7 +1111,32 @@ func Unmarshal(rawVal any, opts ...DecoderConfigOption) error {
 }
 
 func (v *Viper) Unmarshal(rawVal any, opts ...DecoderConfigOption) error {
-	return decode(v.AllSettings(), defaultDecoderConfig(rawVal, opts...))
+	// TODO: make this optional?
+	structKeys, err := v.decodeStructKeys(rawVal, opts...)
+	if err != nil {
+		return err
+	}
+
+	// TODO: struct keys should be enough?
+	return decode(v.getSettings(append(v.AllKeys(), structKeys...)), defaultDecoderConfig(rawVal, opts...))
+}
+
+func (v *Viper) decodeStructKeys(input any, opts ...DecoderConfigOption) ([]string, error) {
+	var structKeyMap map[string]any
+
+	err := decode(input, defaultDecoderConfig(&structKeyMap, opts...))
+	if err != nil {
+		return nil, err
+	}
+
+	flattenedStructKeyMap := v.flattenAndMergeMap(map[string]bool{}, structKeyMap, "")
+
+	r := make([]string, 0, len(flattenedStructKeyMap))
+	for v := range flattenedStructKeyMap {
+		r = append(r, v)
+	}
+
+	return r, nil
 }
 
 // defaultDecoderConfig returns default mapstructure.DecoderConfig with support
--- a/viper_test.go
+++ b/viper_test.go
@@ -948,6 +948,105 @@ func TestUnmarshalWithDecoderOptions(t *testing.T) {
 	}, &C)
 }
 
+func TestUnmarshalWithAutomaticEnv(t *testing.T) {
+	t.Setenv("PORT", "1313")
+	t.Setenv("NAME", "Steve")
+	t.Setenv("DURATION", "1s1ms")
+	t.Setenv("MODES", "1,2,3")
+	t.Setenv("SECRET", "42")
+	t.Setenv("FILESYSTEM_SIZE", "4096")
+
+	type AuthConfig struct {
+		Secret string `mapstructure:"secret"`
+	}
+
+	type StorageConfig struct {
+		Size int `mapstructure:"size"`
+	}
+
+	type Configuration struct {
+		Port     int           `mapstructure:"port"`
+		Name     string        `mapstructure:"name"`
+		Duration time.Duration `mapstructure:"duration"`
+
+		// Infer name from struct
+		Modes []int
+
+		// Squash nested struct (omit prefix)
+		Authentication AuthConfig `mapstructure:",squash"`
+
+		// Different key
+		Storage StorageConfig `mapstructure:"filesystem"`
+
+		// Omitted field
+		Flag bool `mapstructure:"flag"`
+	}
+
+	v := New()
+	v.SetEnvKeyReplacer(strings.NewReplacer(".", "_"))
+	v.AutomaticEnv()
+
+	t.Run("OK", func(t *testing.T) {
+		var config Configuration
+		if err := v.Unmarshal(&config); err != nil {
+			t.Fatalf("unable to decode into struct, %v", err)
+		}
+
+		assert.Equal(
+			t,
+			Configuration{
+				Name:     "Steve",
+				Port:     1313,
+				Duration: time.Second + time.Millisecond,
+				Modes:    []int{1, 2, 3},
+				Authentication: AuthConfig{
+					Secret: "42",
+				},
+				Storage: StorageConfig{
+					Size: 4096,
+				},
+			},
+			config,
+		)
+	})
+
+	t.Run("Precedence", func(t *testing.T) {
+		var config Configuration
+
+		v.Set("port", 1234)
+		if err := v.Unmarshal(&config); err != nil {
+			t.Fatalf("unable to decode into struct, %v", err)
+		}
+
+		assert.Equal(
+			t,
+			Configuration{
+				Name:     "Steve",
+				Port:     1234,
+				Duration: time.Second + time.Millisecond,
+				Modes:    []int{1, 2, 3},
+				Authentication: AuthConfig{
+					Secret: "42",
+				},
+				Storage: StorageConfig{
+					Size: 4096,
+				},
+			},
+			config,
+		)
+	})
+
+	t.Run("Unset", func(t *testing.T) {
+		var config Configuration
+
+		err := v.Unmarshal(&config, func(config *mapstructure.DecoderConfig) {
+			config.ErrorUnset = true
+		})
+
+		assert.Error(t, err, "expected viper.Unmarshal to return error due to unset field 'FLAG'")
+	})
+}
+
 func TestBindPFlags(t *testing.T) {
 	v := New() // create independent Viper object
 	flagSet := pflag.NewFlagSet("test", pflag.ContinueOnError)


sagikazarmark:
Thanks for sending this PR!

I'll try to prioritize reviewing it, especially since it resolves a long standing problem.

swengineer404:
Can we get an update on this?

bkubiak:
@sagikazarmark It would be great to get this PR merged. Can you check it, please?

na-ji:
Hello,

I tried this PR quickly on my personal configuration, and can tell that it does not work with slices/arrays. It might be a problem with mapstructure though.

krakowski:
Hi @na-ji,

There are test cases with slices included in this PR which work fine. See [here](https://github.com/spf13/viper/pull/1429/commits/0ac9227ed9045006258289ac2f7488c4ea2b537a#diff-cbedff684a75d0a2cfdc7d4234fb3573dd7af147878776e039d69e1694a18b34R980-R995) for example. What exactly does not work in your case?

maufranchini:
> Any update on this issue? I'm currently stuck with deployment due to this problem. Is there a potential workaround?

One approach that has worked for me is to utilize the `BindEnv` function provided by Viper.

For instance, if you have a configuration key `server.env` and you're setting an environment variable named `SERVER_ENV`, you can link them using:

```go
viper.BindEnv("server.env", "SERVER_ENV")
```
After this binding, you should be able to access the value of the SERVER_ENV environment variable using the corresponding configuration key in your application.

mathnogueira:
Any update on this PR? This would be extremely valuable, especially for those starting with the library as it's not clear that you have to use `BindEnv` to be able to unmarshal the env variable into the struct.

mathnogueira:
Any update on this PR? Having to add an extra `BindEnv` for each new config we add to the struct is error-prone. This new function would help a lot of us.

sagikazarmark:
Apologies for the late review everyone.

Using BindEnv as a workaround sounds OK, but maybe we could come up with a better solution specifically for Unmarshal.

The original reason why Unmarshal does not work with AutomaticEnv is this: internally, Unmarshall gets a list of keys already registered in Viper and builds the input for mapstructure from those keys.

Implementation: https://github.com/spf13/viper/blob/master/viper.go#L2098

Here is what I propose:

- BindStruct should register a list of keys in an internal key list (deduplicated)
- When Unmarshal calls AllSettings (and transitively GetKeys), it should merge the bound list with the available keys

Practically, we probably need an internal implementation of AllSettings that does that just for Unmarshal (we may be able to merge those back together later, but technically it could be a BC break since those keys don't necessarily have any values).

I'm a little weary to accept this PR as it is, because I don't think BindEnv is the right solution, it's rather a workaround.

I'm happy to give this a go in the next coming weeks, but if anyone wants to pick it up first, I'm happy to review PRs too.

mathnogueira:
Hey @sagikazarmark ! I just opened a PR to move the implementation out of a new method to the `viper.Unmarshal` function. I would love your review on that.

krakowski:
Hey @sagikazarmark,

thanks for taking a look at this! Just want to make sure I fully understand what you propose.

> * BindStruct should register a list of keys in an internal key list (deduplicated)
> * When Unmarshal calls AllSettings (and transitively GetKeys), it should merge the bound list with the available keys

Do you mean `BindStruct` should create a list of keys here?

https://github.com/spf13/viper/blob/c4dcd31f68e5d77ce447c0091dd1ca6d7e169807/viper.go#L210-L216

If so, this would mean those keys need to be merged with existing keys according to some priority within `AllKeys`.

https://github.com/spf13/viper/blob/c4dcd31f68e5d77ce447c0091dd1ca6d7e169807/viper.go#L2018-L2035

Since both `BindStruct` and `BindEnv` are used for binding environment variables, I'm not really sure which should be prioritized at this point. Do you have some thoughts on this?

Also, do you have any examples in mind where internally using `BindEnv` is disadvantageous? That might make it clearer to me why it is supposed to be a workaround.

Thanks!

mathnogueira:
Since @krakowski is taking a look at this. I'll close my PR for now.

krakowski:
> On second thought, I'm thinking maybe we should merge BindStruct's functionality into Unmarshal. Instead of a separate BindStruct call, we could just do all that within Unmarshal. We wouldn't have to store the keys in Viper either.

I will refactor this PR today to reflect this change making `BindStruct` a private function (`bindStruct`) inside `viper` and calling it from `Unmarshal` :+1:

That way, from the view of a user, only `Unmarshal` has to be called. This is what most people expect looking at the issues I referenced in this PR's description.

krakowski:
@sagikazarmark This is now implemented. Please have a look :slightly_smiling_face: 

One thing I noticed: There's also `UnmarshalKey`.
https://github.com/spf13/viper/blob/c4dcd31f68e5d77ce447c0091dd1ca6d7e169807/viper.go#L1103-L1105

Calling `bindStruct` there would traverse the whole struct every time `UnmarshalKey` is called. My proposal is to call `BindEnv` with the user-provided key like so:

```go
func (v *Viper) UnmarshalKey(key string, rawVal any, opts ...DecoderConfigOption) error {
	if err := v.BindEnv(key); err != nil {
		return err
	}
	
	return decode(v.Get(key), defaultDecoderConfig(rawVal, opts...))
}
```

What do you think?

sagikazarmark:
I don't think we need to do this for `UnmarshalKey` since the key in that case is already known and is passed to the `Get` call which will resolve the right env var. There is no need to call BindEnv.

viper.go
```diff
@@ -2034,6 +2038,24 @@ func (v *Viper) AllKeys() []string {
...
 
+// bindStruct binds a struct to the environment by extracting all Viper keys
+// from it and binding each to a ENV variable using BindEnv.
+func (v *Viper) bindStruct(input interface{}) error {
+	envKeysMap := map[string]interface{}{}
+	if err := mapstructure.Decode(input, &envKeysMap); err != nil {
+		return err
+	}
+
+	structKeys := v.flattenAndMergeMap(map[string]bool{}, envKeysMap, "")
```
sagikazarmark, line 2049: These keys should be merged with the list of AllKeys instead. No need to call BindEnv.

sagikazarmark:
@krakowski Thanks for updating the PR!

I thought we agreed not to call BindEnv, but to change the way AllSettings is called and merge the list of keys with AllKeys.

sagikazarmark:
I quickly made my own version: #1699

@krakowski I'd love to get your thoughts on it.

krakowski:
Hey @sagikazarmark,

> @krakowski Thanks for updating the PR!
> 
> I thought we agreed not to call BindEnv, but to change the way AllSettings is called and merge the list of keys with AllKeys.

I misunderstood the last part of your answer.

>> Do you mean BindStruct should create a list of keys here?

> On second thought, I'm thinking maybe we should merge BindStruct's functionality into Unmarshal. Instead of a separate BindStruct call, we could just do all that within Unmarshal. We wouldn't have to store the keys in Viper either.

This and the thumbs up on my plan led me to believe you dismissed the ideas posted in the original answer before making the edit and were okay with calling `bindStruct` internally. Looking at your changes I know see what you meant. I tried out your changes and they work as intended.

Can I pull your changes/commits into my branch (linked with this PR) and add myself as a co-author? Would be pretty demotivating having this open for 1 year just to get closed in the end without any attribution.

Again, thanks for taking a look into this! :slightly_smiling_face:

sagikazarmark:
> Can I pull your changes/commits into my branch (linked with this PR) and add myself as a co-author?

Of course, go ahead. 👍

krakowski:
Done, thank you :+1:

sagikazarmark (approved):
@krakowski appreciate your work on this, thank you!

andig:
@sagikazarmark @krakowski this still does not unmarshal something like `EVCC_LOADPOINTS_0_TITLE` into `[]map[string]interface{}`, right?

andig:
Really sorry for the noise, but I don't see how this works. Please consider:

```go
func TestEnv(t *testing.T) {
	v := viper.New()

	v.SetEnvPrefix("evcc")
	v.SetEnvKeyReplacer(strings.NewReplacer(".", "_"))
	v.AutomaticEnv()

	os.Setenv("EVCC_DATABASE_TYPE", "sqlite")
	os.Setenv("EVCC_DATABASE_DSN", "dsn")

	var conf struct {
		Database struct {
			Type string
			Dsn  string
		}
	}
	assert.NoError(t, v.UnmarshalExact(&conf))

	assert.Equal(t, "sqlite", conf.Database.Type)
	assert.Equal(t, "dsn", conf.Database.Dsn)
}
```

I expected this to work with 1.18, but it doesn't. Previously I was doing:

```go
// register all known config keys
flat, _ := flatten.Flatten(structs.Map(conf), "", flatten.DotStyle)
for _, k := range maps.Keys(flat) {
	_ = v.BindEnv(k)
}
```

...which does work.

Could you kindly clarify why above test wouldn't work?

andig:
@krakowski seems it cannot, that would actually be https://github.com/spf13/viper/issues/935.

krakowski:
@andig seems `UnmarshalExact` was missed in this PR. I opened #1704 to fix this. Your test case should work with the changes made there.

sagikazarmark:
Fix released in https://github.com/spf13/viper/releases/tag/v1.18.1

maxisam:
Hi, this doesn't work in this case. 

https://github.com/maxisam/mgob/actions/runs/7198427412/job/19607841688?pr=97#step:10:701

The config is like 

https://github.com/maxisam/mgob/blob/144a21f8275314f74b9ff3a279fbfea22cda06d3/pkg/config/plan.go#L14-L29

for property under target, it works, but not for property under Azure. 

Here is the test
https://github.com/maxisam/mgob/blob/144a21f8275314f74b9ff3a279fbfea22cda06d3/pkg/config/plan_test.go#L40-L69

Thanks!

xsteadfastx:
> > @sagikazarmark @krakowski this still does not unmarshal something like `EVCC_LOADPOINTS_0_TITLE` into `[]map[string]interface{}`, right?
> 
> You mean the map stored at index 0 should then contain a `TITLE` key with the value assigned to the `EVCC_LOADPOINTS_0_TITLE` environment variable? I can try it out. Does `BindEnv` support this kind of unmarshal operation?

tried it with the latest version with no luck.

sagikazarmark:
@xsteadfastx you should use `mapstructure` struct tag instead of `yaml`. Not sure if that's the only problem, but that is one.

xsteadfastx:
im using mapstructure.

andig:
> im using mapstructure.

You're using yaml tags, not mapstrucure. It shouldn't matter though as long as the member name is a straightforward version of the actual config. That seems to be the case afaikt.

That said, https://github.com/spf13/viper/pull/1429#issuecomment-1855405306 is actually OT as was my post since it referred to slices which were/are not supported via ENV.

krakowski:
@maxisam It seems `mapstructure` does not follow pointer types when decoding fields.

```go
type StorageConfig struct {
  Size int `mapstructure:"size"`
}

type Configuration struct {
  Storage StorageConfig `mapstructure:"filesystem"`
}
```

Decoding the `Configuration` struct using `mapstructure` will yield a `filesystem.size` key. If you change the `Storage` field's type to `*StorageConfig` (pointer type) it yields a `filesystem` key. The code you posted does not work, because in your case `test` is used as a key instead of `test.env`.

A simple workaround you can use is to bind this key manually using `BindEnv`.

```go
...
v.AutomaticEnv()

os.Setenv("EVCC_DATABASE_TYPE", "sqlite")
os.Setenv("EVCC_DATABASE_DSN", "dsn")
os.Setenv("EVCC_TEST_ENV", "test-env")

v.BindEnv("test.env")
...
```

Another option is to not use pointer fields inside the configuration structs.

maxisam:
@krakowski Thanks for explaining, I knew it must be pointer type. I think using pointer type should be pretty common, but maybe I am wrong. (I only use Golang in this project)

xsteadfastx:
> > im using mapstructure.
> 
> You're using yaml tags, not mapstrucure. It shouldn't matter though as long as the member name is a straightforward version of the actual config. That seems to be the case afaikt.
> 
> That said, [#1429 (comment)](https://github.com/spf13/viper/pull/1429#issuecomment-1855405306) is actually OT as was my post since it referred to slices which were/are not supported via ENV.

why do you think im using yaml? i mean i didnt event posted any code.

sagikazarmark:
@maxisam to be precise: mapstructure doesn't support multiple indirections. In fact, it requires you to pass a pointer to your desired output, but in case of the input it only dereferences a pointer once which results in the error seen before.

I submitted a PR to mapstructure to fix that. No idea when or if it' going to be accepted.

I have another PR in Viper as a workaround: #1707

@xsteadfastx this is what makes me believe you use `yaml` as struct tags: https://github.com/maxisam/mgob/blob/144a21f8275314f74b9ff3a279fbfea22cda06d3/pkg/config/plan.go#L14-L29

It works, as long as you tell mapstructure to use that instead of `mapstructure`.

But as @andig pointed out, slices may be another issue.

krakowski:
> @krakowski Thanks for explaining, I knew it must be pointer type. I think using pointer type should be pretty common, but maybe I am wrong. (I only use Golang in this project)
> 
> I also wanna point it out this is actually a breaking change. After this, we can't no longer use pointer type with environment variable. But it was fine before.
> 
> I like the idea of this change but I think the pointer type is a big gotcha and limit the usage a lot

@maxisam The code you posted did not work before this change (or ever) either. Please see the following replit based on viper `v1.17.0`: https://replit.com/@krakowski1/Viper-Automatic-Env#main.go

Before this change you needed to call `BindEnv` for every key. Now it is required for pointer type fields only because of the way `mapstructure` decoding works (which is what I mentioned as a workaround).

maxisam:
it did work before. I had an unit test for that. I load config first and load env variable

maxisam:
@sagikazarmark Thanks for explaining. Also you points to my repo not @xsteadfastx

preslavrachev:
> Released v1.18.0 with this change.

Has this been made publicly available? In the code, I am seeing a check, which in my case defaults to `false`:

```go
const BindStruct = false
```

How do I turn this on within my project?

preslavrachev:
Whoever arrives here looking for how to enable this feature, you need to add the `tags=viper_bind_struct` option when running `go build`, `go run`, or `go test`. This is rather unfortunate, as it now needs me to create makefile to ensure that my app is started correctly (unfortunately, Go does not have default build options, as to my knowledge).

You can still make your life easier by enabling those build flags in your VS Code configuration too:

```json
{
  "gopls": {
    "build.buildFlags": ["-tags=viper_bind_struct"]
  },
  "go.buildFlags": ["-tags=viper_bind_struct"],
  "go.testFlags": ["-tags=viper_bind_struct"]
}

@sagikazarmark any idea on whether this feature will be enabled by default in future releases?
```

poundifdef:
@preslavrachev There is much more context for why this was put behind a build flag here: https://github.com/spf13/viper/issues/1706

Basically, that code caused regressions in specific cases and so it was put behind a flag until it is fully backwards-compatible.

alfa-alex:
I have a question about the expected behavior, I hope it's fine to ask this here (not sure yet if this is a bug or intended): Is it correct that fields tagged with `omitempty` cannot be set by environment variable only?

For example, when I change [this line](https://github.com/spf13/viper/blob/ea35b92596da4253bd5e5514a7ce363de374bd63/viper_test.go#L974) to be
```
		Secret string `mapstructure:"secret,omitempty"`
```
instead, the test fails.

andig:
Is `omitempty` even a valid mapstructure tag?

PR #1430, opened 2022-09-06
Use "go install"

guettli:


sagikazarmark:
Are you sure this is necessary? I don't know the difference between get and install anymore.

guettli:
I don't know all the details, but AFAIK `go install` is recommended.

See: https://go.dev/doc/go1.18#go-get

sagikazarmark:
As far as I understand, go get should be used to install a dependency, go install is for downloading and installing executables.

guettli:
@sagikazarmark yes, I guess you are right. Sorry for the noise.

PR #1433, opened 2022-09-18
Allow ConfigParseError to unwrap

andig:
Currently, viper errors cannot be unwrapped and hence do not give access to inner errors. Together with e.g. https://github.com/go-yaml/yaml/pull/760 this could be used to get access to the source file position producing the error.

### DIFF
--- a/util.go
+++ b/util.go
@@ -27,12 +27,12 @@ type ConfigParseError struct {
 }
 
 // Error returns the formatted configuration error.
-func (pe *ConfigParseError) Error() string {
+func (pe ConfigParseError) Error() string {
 	return fmt.Sprintf("While parsing config: %s", pe.err.Error())
 }
 
 // Unwrap returns the wrapped error.
-func (pe *ConfigParseError) Unwrap() error {
+func (pe ConfigParseError) Unwrap() error {
 	return pe.err
 }
 
--- a/util.go
+++ b/util.go
@@ -27,8 +27,13 @@ type ConfigParseError struct {
 }
 
 // Error returns the formatted configuration error.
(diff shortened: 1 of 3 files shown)


andig:
ping @sagikazarmark anything blocking this change I could help with?

andig:
friendly ping :)

andig:
Refs https://github.com/go-yaml/yaml/pull/901 now. Response would be nice 👍🏻

util.go
```diff
@@ -27,10 +27,15 @@ type ConfigParseError struct {
 }
 
 // Error returns the formatted configuration error.
-func (pe ConfigParseError) Error() string {
+func (pe *ConfigParseError) Error() string {
```
sagikazarmark, line 30: Why make it a pointer?

sagikazarmark (changes requested):

util.go
andig: Honestly I don't remember right now. I came from yaml issue referenced above and had some restriction with non-pointer values. Started discussion in https://groups.google.com/u/1/g/golang-nuts/c/3M3QDkochcs/m/uo-x6jA0BgAJ which concluded in 

> however i think it is a good default position to use pointer values to 
implement interfaces generally, if there is no other good reason not to. 

If you're not happy with pointer type I'll need to recreate the entire case and try to find out what the root cause issue with non-pointer types was.

sagikazarmark: Honestly, I'm not happy doing it unless I understand why it's necessary.

andig: I have reverted the pointer type and added a test- it does unwrap as expected.

PR #1439, opened 2022-09-28
Merge defaults when Get()'ing maps

### LINKED ISSUE #747: viper.Sub() not extracting nested default values
I noticed an issue where viper.Sub() fails to extract default values of a nested key. I attached demo code to reproduce the issue using latest viper and Go 1.12.6 under x64_linux.

`main.go`:
~~~go
package main

import (
	"fmt"

	"github.com/spf13/viper"
)

func main() {
	viper.SetDefault("config.value2.internal", 3)
	viper.SetConfigFile("./demo.yaml")
	err := viper.ReadInConfig()
	if err != nil {
		panic(fmt.Errorf("Fatal error config file: %s ", err))
	}
	// Until now, everything works as expected
	fmt.Printf("config.value1: %d\n", viper.GetInt("config.value1"))
	fmt.Printf("config.value2: %d\n", viper.GetInt("config.value2"))
	fmt.Printf("config.value2.internal: %d\n", viper.GetInt("config.value2.internal"))

	// Extract the config subtree
	v := viper.Sub("config")
	// This doesn't work if value2 node is missing in configuration
	fmt.Printf("value2.internal: %d\n", v.GetInt("value2.internal"))

	spew.Dump(v.AllSettings())
}
~~~

`demo.yaml` (works as inteded, but defaults are useless since everything is explicitly set):
~~~yaml
config:
  value1: 1
  value2: 
    internal: 3
~~~

`demo.yaml` (broken):
~~~yaml
config:
  value1: 1
  #value2: 
  #  internal: 3
~~~

The issue seems somehow related / similar to issue #71 as well as PR #195

setrofim:
When fetching a configuration sub-tree (either via Get() that results in a map, or through Sub()), ensure that any defaults for keys under that tree are propagated into the returned result.

Fixes #747

Signed-off-by: setrofim <setrofim@gmail.com>


setrofim:
Note: this is only a partial solution for defaults. Similar issues exist across other layers (flags, overrides).

setrofim:
I've updated the pull with a more complete solution that considers all layers.

PR #1435, opened 2022-09-24
extend stringToString pflag binding to stringToInt pflag

vorishirne:
Support for pflag's StringToString was added long back, but Viper Still couldn't bind to pflag StringToInt. Addid that support.

PR #1462, opened 2022-11-08
 extend v.find() to handle stringToInt value types with stringToInt pflag #1435 

vorishirne:
Support for pflag's StringToString was added long back, but Viper Still couldn't bind to pflag StringToInt. Addid that support.

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -1283,7 +1283,8 @@ func (v *Viper) find(lcaseKey string, flagDefault bool) interface{} {
 			return cast.ToDurationSlice(slice)
 		case "stringToString":
 			return stringToStringConv(flag.ValueString())
-
+		case "stringToInt":
+			return stringToIntConv(flag.ValueString())
 		default:
 			return flag.ValueString()
 		}
@@ -1363,6 +1364,8 @@ func (v *Viper) find(lcaseKey string, flagDefault bool) interface{} {
 				return cast.ToIntSlice(res)
 			case "stringToString":
 				return stringToStringConv(flag.ValueString())
+			case "stringToInt":
+				return stringToIntConv(flag.ValueString())
 			case "durationSlice":
 				s := strings.TrimPrefix(flag.ValueString(), "[")
 				s = strings.TrimSuffix(s, "]")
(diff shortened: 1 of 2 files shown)


viper.go
```diff
@@ -1371,7 +1371,7 @@ func readAsCSV(val string) ([]string, error) {
 
 // mostly copied from pflag's implementation of this operation here https://github.com/spf13/pflag/blob/master/string_to_string.go#L79
 // alterations are: errors are swallowed, map[string]interface{} is returned in order to enable cast.ToStringMap
-func stringToStringConv(val string) interface{} {
+func stringToInterfaceCong(val string) interface{} {
```
sagikazarmark, line 1374: Typo: `Cong` vs `Conv`

viper_test.go
```diff
@@ -1166,16 +1166,22 @@ func TestBindPFlagDetectNilFlag(t *testing.T) {
...
 		Expected map[string]string
 		Value    string
 	}{
+		// stringToString
 		{map[string]string{}, ""},
 		{map[string]string{"yo": "hi"}, "yo=hi"},
 		{map[string]string{"yo": "hi", "oh": "hi=there"}, "yo=hi,oh=hi=there"},
 		{map[string]string{"yo": ""}, "yo="},
 		{map[string]string{"yo": "", "oh": "hi=there"}, "yo=,oh=hi=there"},
+		//stringToInt
```
sagikazarmark, line 1180:
```suggestion
		// stringToInt
```

sagikazarmark: I wonder what would have happened with this test cases without adding `stringToInt` above.

sagikazarmark (changes requested):

viper_test.go
vorishirne: before this, int values were merely being treated as integers

viper.go
vorishirne: done

viper_test.go
vorishirne: done above suggested change

sagikazarmark (approved):
Thank you!

viper.go
sagikazarmark: Sorry, I'm not sure I was clear: it should be `Conv`

viper.go
```diff
@@ -1348,8 +1348,8 @@ func (v *Viper) find(lcaseKey string, flagDefault bool) interface{} {
 				s = strings.TrimSuffix(s, "]")
 				res, _ := readAsCSV(s)
 				return cast.ToIntSlice(res)
-			case "stringToString":
-				return stringToStringConv(flag.ValueString())
+			case "stringToString", "stringToInt":
+				return stringToInterfaceCong(flag.ValueString())
```
sagikazarmark, line 1352: `Conv` here as well

sagikazarmark (changes requested):
Would you mind also rebasing your branch? CI is not triggered for some reason.

viper.go
vorishirne: sorry I mistakenly didn't pushed the changes

viper.go
vorishirne: sorry I mistakenly didn't pushed the changes

vorishirne:
@sagikazarmark stringToInt should be handeled separately, as per this https://github.com/spf13/pflag/blob/d5e0c0615acee7028e1e2740a11102313be88de1/string_to_int.go#L68

Hence I have updated the pull request with that respect. Sorry for going in wrong direction earlier

sagikazarmark (approved):

sagikazarmark:
Thanks @vorishirne !

PR #1483, opened 2023-01-03
add:viper: allow to disable internal log messages

burgesQ:
Thanks 🙏🏻

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -433,16 +433,14 @@ func (v *Viper) WatchConfig() {
 	go func() {
 		watcher, err := newWatcher()
 		if err != nil {
-			v.logger.Error("failure to create watcher",
-				"msg", err.Error())
+			v.logger.Error(fmt.Sprintf("failed to create watcher: %s", err))
 			os.Exit(1)
 		}
 		defer watcher.Close()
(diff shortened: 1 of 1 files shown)


sagikazarmark:
Thanks for the PR @burgesQ !

I believe a better solution would be setting the output for the logger to something like `ioutil.Discard`.

Viper v1 uses jww: https://github.com/spf13/jwalterweatherman#setting-a-log-file

burgesQ:
> Viper v1 uses jww: [spf13/jwalterweatherman#setting-a-log-file](https://github.com/spf13/jwalterweatherman#setting-a-log-file)

This MR is about discarding message that came from the std' package (log) - not the one embed into the viper struct (jww). 

I've already hooked my syslog logger to the viper' struct Logger interface (jww) - but still some messages are outputted to stdout (log).

Turn out those messages are emitted by another logger than the Logger interface. Otherwise, I can change the viper code, so they're emitted through jww

sagikazarmark:
Oh yeah, that would be great. We shouldn't be using multiple loggers at the same time. So please change it to use the existing logger.

You may have to add a return statement after, because the new logger interface does not enforce calling `os.Exit`.

burgesQ:
👐🏻

PR #1532, opened 2023-04-03
Update main image to better handle dark background

Deleplace:
Added a thin white border around the black letters.

Same fix as https://github.com/spf13/cobra/issues/1880

The same image now looks good both in light theme as well as dark theme.

Before:
image

After:
image


sagikazarmark (approved):
Thanks!

PR #1538, opened 2023-04-13
feat: Sub uses key as env prefix

KnutZuidema:
This change implements usage of the key given to Sub as an additional env prefix if the value at the given key is not available.
This let's users provide environment variables in the form of `PREFIX_SUB_VALUE` and the call `v.Sub("sub")` and access the value via `v.Get("value")`.

This changes the existing behavior of `Sub` in the sense that we won't return `nil` if a key is not found with a map value.


KnutZuidema:
The CLA page is not working for me, all I see is a blank version selector that I cannot interact with.

PR #1541, opened 2023-04-15
[FEATURE] it's option to set case sensitive

Nothing-no:
anyway, I think it's better to offer user an option to choose case sensitive or not.

### DIFF
--- a/README.md
+++ b/README.md
@@ -862,7 +862,16 @@ application foundation needs.
 Is there a better name for a [commander](http://en.wikipedia.org/wiki/Cobra_Commander)?
 
 ### Does Viper support case sensitive keys?
+#### [FEATURE] surport case sensitive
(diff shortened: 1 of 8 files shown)


Nothing-no:
> anyway, I think it's better to offer user an option to choose case sensitive or not.
### add MapTo
- source file

```yaml
service:
    ip: 127.0.0.1
    port: 1234
version: 1.0.01
```

- code
```go
type Service struct {
	Port int `viper:"port"`
	IP string `viper:"ip"`
}
//your prepare code ...

var service Service
var version string

if err := viper.MapTo("service",&service); err != nil {
	//error handler...
}
if err := viper.MapTo("version",&version); err != nil {
	//error handler...
}



log.Println(service,version)

//....
```

PR #1565, opened 2023-06-13
feat: add func GetEnvPrefix

### LINKED ISSUE #802: add func: GetEnvPrefix() string
we have the `func SetEnvPrefix(in string)`, can we also add `func GetEnvPrefix() string`?

My use case for this is to find keys related to the application which viper cannot detect on it's own.
I don't mind creating the PR if needed.

baruchiro:
Close #802

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -523,6 +523,12 @@ func (v *Viper) SetEnvPrefix(in string) {
 	}
 }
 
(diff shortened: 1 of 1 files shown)


sagikazarmark:
Well, I guess it doesn't hurt.

PR #1387, opened 2022-06-23
Recurse into arrays when converting keys to lowercase

awrichar:
Fixes #1386

PR #1577, opened 2023-06-29
feat: partially revert #1387

### LINKED ISSUE #1576: Allow reverting slice lowercasing to the previous behavior
### Preflight Checklist

### Problem Description

#1387 fixed a bug around how case insensitivity works in Viper. Unfortunately it causes issues for a number of key players in the field.

### Proposed Solution

Allow reverting back to the previous behavior using a go tag. This is going to be a Viper v1 thing only.

Viper v2 will either be completely case insensitive or will allow partial case sensitivity.

### Alternatives Considered

_No response_

### Additional Information

_No response_

sagikazarmark:
1387 introduced a rather serious breaking change that's causing pain throughout the community. Even though it fixed a buggy behavior the previous one was still useful to a lot of people.

This change introduces a way to revert back to that old behavior.

Fixes #1576


sagikazarmark:
@ldez @jerome-laforge @b-repo @deynekas Is this something you could work with to get back the previous behavior?

ldez:
Hello,

the build tag is a not real solution for us because golangci-lint can be installed with `go install` or use with the "tools" pattern.
So the current PR will not fix our problem.

PR #1581, opened 2023-07-05
Provide SetKeyDelim

mfine:
Allow the key delimiter to be set on viper objects.

Currently it's not possible to change the key delimiter of viper objects after construction - allow a method to update viper objects after construction similar to `SetEnvKeyReplacer`.

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -1478,11 +1478,11 @@ func (v *Viper) SetEnvKeyReplacer(r *strings.Replacer) {
 	v.envKeyReplacer = r
 }
 
-// SetKeyDelim sets the delimiter used for determining key parts on the viper
+// SetKeyDelimiter sets the delimiter used for determining key parts on the viper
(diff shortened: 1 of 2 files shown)


sagikazarmark:
@mfine thanks for opening a PR!

Can you tell me your use case for changing the delimiter?

It was actually a very conscious decision not to allow changing it on the fly as it introduces as kinds of issues with keys created prior to the change.

I'd love to hear your use case though, so we can figure out a solution together.

Thanks!

mfine:
@sagikazarmark thanks for the response. Our use case is around existing usage built around the default/global viper object where the `.` delimiter is problematic: other languages are using a different delimiter and the `.` is problematic with environment variable names in a lot of contexts. Hence, we were looking for a way to change the delimiter on the default/global viper object without having to change a lot of the existing code.

Issue #1404, opened 2022-07-18
API to stop the file watcher


skidder:
### Preflight Checklist

- [X] I have searched the [issue tracker](https://www.github.com/spf13/viper/issues) for an issue that matches the one I want to file, without success.

### Problem Description

I'd like to be able to stop the file watcher created by calling `Viper.WatchFile()`. Being able to gracefully tear down the filewatcher helps to ensure a clean shutdown without possible invocations of the `onConfigChange` function.

### Proposed Solution

Add a Golang context and cancel-function to the `Viper` struct type. Move the events waitgroup (`eventsWG`) from the watcher go-routine to the Viper struct type. Initialize these fields in the `WatchConfig()` function. Add a `Viper.Stop()` function that calls the cancel function to stop the watcher.

lvyonghuan:
I think this feature should be there, it has its uses

PR #1583, opened 2023-07-07
feat: add a cancel function return from WatchConfig() to stop watchin…

chowyi:
I need to stop the WatchConfig() function in certain scenarios. I found someone with the same need in an open Issue: https://github.com/spf13/viper/issues/1404. I have attempted to implement it.

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -21,6 +21,7 @@ package viper
 
 import (
 	"bytes"
+	"context"
 	"encoding/csv"
 	"errors"
 	"fmt"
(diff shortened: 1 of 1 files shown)


sagikazarmark:
Thanks @chowyi!

Unfortunately, this is a breaking change which is not allowed on the v1 release track at the moment.

I'd be happy to accept a change that introduces a new function implementing the same functionality.

chowyi:
Thanks for your suggestion. 
I reverted the previous modification, added a new function `StopWatching()` implementing the same functionality and the unitTest.

PR #1584, opened 2023-07-12
feat: lazy initialization codec

xieyuschen:
# What I have changed
I have applied the [`TODO: make this lazy initialization instead`](https://github.com/spf13/viper/blob/097e0d888f8c9aa9270a15dce08e170cffc19f36/viper.go#L307) left by @sagikazarmark . The main changes are listed as following:

- lazily initialize the codec
- merge encoder and decoder to `codec`. 
This change is caused as I try to create a new codec by calling a method of a nil pointer with type. I want the decoder and encoder to share the same object constructed by the method.
- create `constructor.Codec` interface to avoid cycle import.

# Benchmark
I use this command to run the benchmark across commits:
```shell
git checkout 214f242 
go test -v -bench=. -count=10 > old.out
git checkout 60d2632
go test -v -bench=. -count=10  > new.out
benchstat old.out new.out
```
The output is:
```
goos: darwin
goarch: amd64
pkg: github.com/spf13/viper
cpu: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
                    │    old.out    │               new.out               │
                    │    sec/op     │   sec/op     vs base                │
New-12                2972.5n ± 77%   431.8n ± 3%  -85.48% (p=0.000 n=10)
_unmarshalReader-12    148.4µ ±  7%   139.1µ ± 3%   -6.25% (p=0.019 n=10)
GetBool-12             186.6n ± 18%   180.6n ± 2%   -3.22% (p=0.017 n=10)
Get-12                 157.5n ±  7%   156.6n ± 1%        ~ (p=0.436 n=10)
GetBoolFromMap-12      3.944n ± 28%   3.878n ± 2%        ~ (p=0.052 n=10)
geomean                551.7n         366.2n       -33.62%
```
For me, I think this change doesn't reduce the efficiency of `viper`.


internal/encoding/codec.go
```diff
@@ -0,0 +1,135 @@
...
+	"toml":       (*toml.Codec)(nil),
+	"hcl":        (*hcl.Codec)(nil),
+	"tfvars":     (*hcl.Codec)(nil),
+	"ini":        (*ini.Codec)(nil),
+	"properties": (*javaproperties.Codec)(nil),
+	"props":      (*javaproperties.Codec)(nil),
+	"prop":       (*javaproperties.Codec)(nil),
+	"dotenv":     (*dotenv.Codec)(nil),
+	"env":        (*dotenv.Codec)(nil),
+}
```
xieyuschen, line 54: For the code here, I could use a map to store all constructors, which is more go style. However, this way is slightly faster. I prefer this way.

My benchmark code i listed below and it's inside the `viper/internal/encoding/codec_bench_test.go`.

```go
package encoding

import (
	"testing"

	"github.com/spf13/viper/internal/encoding/yaml"
)

func BenchmarkNilType(b *testing.B) {
	m := map[string]Codec{
		"test": (*yaml.Codec)(nil),
	}
	c := m["test"]
	for i := 0; i < b.N; i++ {
		c.Construct()
	}
}
func BenchmarkWithMapFn(b *testing.B) {
	m := map[string]func() Codec{
		"test": func() Codec {
			return &yaml.Codec{}
		},
	}
	fn := m["test"]
	for i := 0; i < b.N; i++ {
		fn()
	}
}
```

## bench result
```
goos: darwin
goarch: amd64
pkg: github.com/spf13/viper/internal/encoding
cpu: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
BenchmarkNilType
BenchmarkNilType-12      	693284240	         1.484 ns/op
BenchmarkWithMapFn
BenchmarkWithMapFn-12    	789944808	         1.528 ns/op
PASS
```

sagikazarmark:
Wow, thanks @xieyuschen !

Give me some time to review this PR, but I really appreciate the contribution.

One note off the bat: performance generally is not a huge concern in case of Viper as it's mostly initialized/used at the beginning of an application run, so readability is generally preferred over rigorous performance optimization.

xieyuschen:
> Wow, thanks @xieyuschen !
> 
> Give me some time to review this PR, but I really appreciate the contribution.
> 
> One note off the bat: performance generally is not a huge concern in case of Viper as it's mostly initialized/used at the beginning of an application run, so readability is generally preferred over rigorous performance optimization.

I agree with you, and I have changed the implementation in a go style. @sagikazarmark Could you kindly check it if you have time?

Regards

PR #1585, opened 2023-07-13
fix isPathShadowedInFlatMap type cast bug

linuxsong:
The `cast.ToStringMap()` method calls `ToStringMapE()` for type conversion, but `ToStringMapE()` does not support conversion from `map[string]string` and `map[string]FlagValue` to `map[string]interface{}`, so when calling the `isPathShadowedInFlatMap()` method, ToStringMapE will always return a " unable to cast %#v of type %T to `map[string]interface{}`" error, so the `isPathShadowedInFlatMap()` method does nothing. 

So in the `isPathShadowedInFlatMap` method, should to call `castMapStringToMapInterface` and `castMapFlagToMapInterface()` respectively to convert `map[string]string` and `map[string]FlagValue` into `map[string]interface{}`.

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -827,9 +827,11 @@ func (v *Viper) isPathShadowedInDeepMap(path []string, m map[string]interface{})
 func (v *Viper) isPathShadowedInFlatMap(path []string, mi interface{}) string {
 	// unify input map
 	var m map[string]interface{}
-	switch mi.(type) {
-	case map[string]string, map[string]FlagValue:
-		m = cast.ToStringMap(mi)
+	switch miv := mi.(type) {
+	case map[string]string:
+		m = castMapStringToMapInterface(miv)
+	case map[string]FlagValue:
+		m = castMapFlagToMapInterface(miv)
 	default:
 		return ""
 	}
--- a/viper_test.go
+++ b/viper_test.go
@@ -2679,6 +2679,31 @@ func TestSliceIndexAccess(t *testing.T) {
 	assert.Equal(t, "Static", v.GetString("tv.0.episodes.1.2"))
 }
 
+func TestIsPathShadowedInFlatMap(t *testing.T) {
+	v := New()
+
+	stringMap := map[string]string{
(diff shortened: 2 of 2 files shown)


sagikazarmark:
@linuxsong thank you for your contribution!

Could you provide a test that fails without this change, so we can make sure we don't reintroduce this bug later?

Thanks!

linuxsong:
>

linuxsong:
> @linuxsong thank you for your contribution!
> 
> Could you provide a test that fails without this change, so we can make sure we don't reintroduce this bug later?
> 
> Thanks!

I added a unit test. It  will fails without this change. Please help to review.

sagikazarmark:
@linuxsong thank you for providing a test.

Is this issue reproducible using the public API of Viper?

For example, can you reproduce it using flags manually passed in?

If so it would be better to write a blackbox test (as well) that doesn't rely on internals, so if any refactoring happens we can safely make changes around the internal code without breaking existing features.

Here is what I would do (although I'm not sure how to reproduce the problem):

Write a test that:

- configures a flag value
- reads in config or sets a default value
- asserts that shadowing happens (or not) by reading from Viper

linuxsong:
> @linuxsong thank you for providing a test.
> 
> Is this issue reproducible using the public API of Viper?
> 
> For example, can you reproduce it using flags manually passed in?
> 
> If so it would be better to write a blackbox test (as well) that doesn't rely on internals, so if any refactoring happens we can safely make changes around the internal code without breaking existing features.
> 
> Here is what I would do (although I'm not sure how to reproduce the problem):
> 
> Write a test that:
> 
> * configures a flag value
> * reads in config or sets a default value
> * asserts that shadowing happens (or not) by reading from Viper

@sagikazarmark  thank you for your review.
I use public API of Viper write a new unit test  reproduce the issue. Without this fix the test case would fail.
Please help to review again.

sagikazarmark (approved):
Thanks!

PR #1590, opened 2023-07-21
Add NATS support

hooksie1:
This adds support for NATS as a remote configuration storage. This is the continuation of the crypt PR @sagikazarmark.


hooksie1:
ok now that @sagikazarmark has merged in the fix for crypt, I think this is ready to go.

sagikazarmark (approved):
Thanks!

Issue #1566, opened 2023-06-13
Regression bug: sub tree configs failing to use environment variables


redkite127:
### Preflight Checklist

- [X] I have searched the [issue tracker](https://www.github.com/spf13/viper/issues) for an issue that matches the one I want to file, without success.
- [X] I am not looking for support or already pursued the available [support channels](https://github.com/spf13/viper/issues/new/choose) without success.
- [X] I have checked the [troubleshooting guide](https://github.com/spf13/viper/blob/master/TROUBLESHOOTING.md) for my problem, without success.

### Viper Version

1.16.0

### Go Version

1.19.4

### Config Source

Environment variables, Files

### Format

YAML

### Code reproducing the issue

```go
package main

import (
	"fmt"
	"log"
	"os"

	"github.com/spf13/viper"
)

type Config struct {
	WorkerCount int            `mapstructure:"workercount"`
	Database    DatabaseConfig `mapstructure:"database"`
}

type DatabaseConfig struct {
	Host     string `mapstructure:"host"`
	Port     int    `mapstructure:"port"`
	Username string `mapstructure:"username"`
	Password string `mapstructure:"password"`
	Database string `mapstructure:"database"`
}

func main() {
	os.Setenv("WORKERCOUNT", "7")
	os.Setenv("DATABASE_PASSWORD", "MyPassword")

	viper.SetConfigName("config")
	viper.AddConfigPath(".")
	err := viper.ReadInConfig()
	if err != nil {
		panic(fmt.Errorf("reading config: %w", err))
	}

	viper.AutomaticEnv()
	log.Println("workercount:", viper.Get("workercount")) // expecting 7, got 7

	subViper := viper.Sub("database")
	subViper.SetEnvPrefix("database")
	subViper.AutomaticEnv()
	log.Println("database_password:", subViper.Get("password")) // expecting "MyPassword", got "1234"
}
```

### Expected Behavior

The value of the password of the database is supposed to be taken from the environment variable, set at the beginning of the `main()`. I should have `MyPassword`.

### Actual Behavior

I have the value from the configuration file and not the one from the environment variable: `admin`

### Steps To Reproduce

1. I used this `config.yaml`:

```
workercount: 4
database:
  host: localhost
  port: 3306
  username: admin
  password: admin
  database: 1234
```

2. Simply executed the code above in v1.16.0.

### Additional Information

It works with `viper:v1.15.0`, but it doesn't with `viper:v1.16.0`.

[...]

PR #1617, opened 2023-09-06
remove parents to use env prefix instead to avoid mixing both concepts

JLugagne:
In the commit [3f4449054d043e278c83a837648d882823a0a3c7](https://github.com/spf13/viper/commit/3f4449054d043e278c83a837648d882823a0a3c7) the concept of parents was added to keep track of the ancestors but led to some regressions (cf: issue [1566](https://github.com/spf13/viper/issues/1566)).

Instead of adding this new concept, why not using the envPrefix which seems to be made for this purpose? It may change Sub's behavior but makes it more intuitive I believe. And keeps the behavior defined by the TestEnvSubConfig.

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -205,7 +205,6 @@ type Viper struct {
 	envKeyReplacer      StringReplacer
 	allowEmptyEnv       bool
 
-	parents        []string
 	config         map[string]interface{}
 	override       map[string]interface{}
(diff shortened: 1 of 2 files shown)


robdesbois:
@TaylorOno @sagikazarmark you co-authored the original commit - any comments on this resolution to #1566?

PR #1624, opened 2023-09-09
feat: use new finder library

sagikazarmark:
This PR introduces a new finder library based file search using `io/fs`.


sagikazarmark:
Closed in favor of #1625

PR #1628, opened 2023-09-12
Make deps fixes

bersace:
Diving into viper development environment, I hit some rough edges. Here are some fixes for this.


sagikazarmark (approved):
Thanks!

Issue #1546, opened 2023-04-25
Let user determine strict prefix for env


bersace:
### Preflight Checklist

- [X] I have searched the [issue tracker](https://www.github.com/spf13/viper/issues) for an issue that matches the one I want to file, without success.

### Problem Description

Hi,

Some apps don't want the `_` between prefix and variable name in environment variables name. For example, standard env prefix for openldap is `LDAP` as in `LDAPNOINIT`, `LDAPRC`, etc. The same goes for libpq with the `PG` prefix as in `PGHOST`, `PGUSER`, etc.

### Proposed Solution

I suggest that viper v2 don't imposes the `_` in the prefix and let the user set an explicit prefix. Migrating from v1 will just require to add a single underscore. That's seems fair. What do you think of this ?

### Additional Information

This is somewhat related to case sensitivity. Env vars are case sensitive, think of `http_proxy`. viper should let user handle this. I understand viper can be opinionated. Still that would be great to bypass viper's opinion and keep the user in control.

Thanks for viper. It's awesome. And pflag too !

Regards,
Étienne

sagikazarmark:
@bersace thank you for the suggestion!

I believe it should be possible to implement this feature in v1 by using a feature gate. We've introduced options a while ago, we could add a new one for this.

On case sensitivity: I understand that in isolation it makes sense, but as soon as you try to merge env vars, case sensitivity just makes the problem infinitely more complex.

For example:

```go
os.Setenv("http_proxy", "foo")
os.Setenv("http_PROXY", "bar")

viper.SetDefault("HTTP_PROXY", "baz")

viper.GetString("HTTP_proxy") // ????
```

There is a chance we can make it optional in certain cases, but due to the incredible complexity of Viper v1, it's certainly not going to happen over night.

bersace:
Cool, I'll try to implement this. Thanks for the feedback.

PR #1629, opened 2023-09-12
feat: Strict env prefix

bersace:
cf. #1546

I didn't get to test with a sub parser. Hint welcome.

cc @sagikazarmark

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -197,6 +197,7 @@ type Viper struct {
 	configType        string
 	configPermissions os.FileMode
 	envPrefix         string
+	envPrefixStrict   bool
 
 	// Specific commands for ini parsing
 	iniLoadOptions ini.LoadOptions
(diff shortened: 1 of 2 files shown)


viper.go
```diff
@@ -519,19 +529,24 @@ func SetEnvPrefix(in string) { v.SetEnvPrefix(in) }
 
 func (v *Viper) SetEnvPrefix(in string) {
 	if in != "" {
-		v.envPrefix = in
+		v.envPrefix = in + "_"
```
sagikazarmark, line 532: Why append the underscore here if there is a separate variable tracking whether it should be?

viper.go
```diff
@@ -282,6 +283,15 @@ func EnvKeyReplacer(r StringReplacer) Option {
 	})
 }
 
+func StrictEnvPrefix(in string) Option {
```
sagikazarmark, line 286: I'm not sure I like the name `StrictEnvPrefix`.

Something like `RawEnvPrefix` or `BareEnvPrefix` might sound better.

An alternative I can imagine is a separate option called EnvPrefixWithDelimiter (deprecating any current options).

sagikazarmark (changes requested):

viper.go
bersace: The variable tracks whether the underscore is added by viper (legacy behaviour) or by user.
The rest of the code does not bother whether the envPrefix was modified by viper or not.

PR #1664, opened 2023-10-18
Update to go1.21.3 and use log/slog instead of slog-shim

Brandon-M-Hartman:
Simple changes, just moving up to go1.21.3 and log/slog instead of using the slog-shim that was used previously. Also removed the deprecated Logger interface as nothing in the code uses it anymore.


sagikazarmark:
@Brandon-M-Hartman could you please rebase your branch to the latest master? Thanks!

sagikazarmark:
Nevermind, I ended up doing it myself. Thanks for the contribution!

Issue #1014, opened 2020-11-03
Case sensitive keys


sagikazarmark:
This issue will track (probably) the most requested feature for Viper: case sensitivity.

Every other issue will be closed in favor of this one.

We know this is something that the community wants (although based on feedbacks, it might not be a majority), so it's being considered for Viper v2. **It doesn't mean it will be implemented though.**

Please see for more details https://github.com/spf13/viper#does-viper-support-case-sensitive-keys

aaronbuchwald:
What is the timeline for a decision on incorporating case sensitivity into v2 as well as the release of v2?

li1234yun:
Maybe key case sensitive can be viper config param, maybe like `WithCaseSensitive` etc. When `WithCaseSensitive == true`, all viper operation is case sensitive, or not.

li1234yun:
I want to mention a little bit about the benefits of case sensitivity. Because the structure public fields of golang start with uppercase, when the read configuration is consistent with the structure fields, we can directly use the configuration structure for deserialization, which saves a lot of time and avoids some unnecessary intermediate conversion.

sagikazarmark:
Unmarshal is case insensitive, so it should work either way.

li1234yun:
> Unmarshal is case insensitive, so it should work either way.

Thanks!

WeiChieh-Chen:
Maybe I'm wrong, but because of the automatic function here, I have to take on additional processing work. And usually, the configuration files are set by themselves, which means that these format designs are useful to users. Is this really right ?

boatrainlsz:
it's been 2 years

link-duan:
fix this please

dreamlu:
https://github.com/spf13/viper/commit/2cb5cbf43cc6d8b777a6f1baae164e3853c8430d
why delete it?

jim3ma:
There are many issues about case sensitive, why it is not end till now ?

meln5674:
If this is not going to be fixed, then viper.BindPFlag() on a pflag.StringToString[Var] or calling viper.Unmarshal() on a struct with map fields should both panic, because those are both guaranteed not to do what the user intended.

sngyai:
Any progress?

chayuto:
i want this fix too!

when unmarshal map[string]string
```
this_is_cap_key:TEST VAL
```

kayvonkhosrowpour:
[...]

PR #1673, opened 2023-10-31
Add CaseSensitiveKeys option

travisnewhouse:
Case-sensitive keys are one of the most request features of Viper.  See #1014.

When reading configuration from sources with case-sensitive keys, such as YAML, TOML, and JSON, a user may wish to preserve the case of keys that appear in maps.  For example, consider when the value of a setting is an (opaque) map with string keys that are case-sensitive. Ideally, if the map value is not going to be indexed by a Viper lookup key, then the map value should be treated as an opaque value by Viper, and its keys should not be modified.

Viper's default behavior is that keys are case-sensitive, and this behavior is implemented by converting all keys to lower-case.  For users that wish to preserve the case of keys, this commit introduces an Option `CaseSensitiveKeys()` that can be used to configure Viper to use case-sensitive keys.  When CaseSensitiveKeys is enabled, all keys retain the original case, and lookups become case-sensitive (except for lookups of values bound to environment variables).

The behavior of Viper could become hard to understand if a user could change the CaseSensitiveKeys setting after values have been stored.  For this reason, the setting may only be set when creating a Viper instance, and it cannot be set on the "global" Viper.  This merge request implements case-sensitive keys in a manner that respects the existing behavior of the "global" Viper, and gives users an option to create a Viper that uses case-sensitive keys behavior.

### DIFF
--- a/util.go
+++ b/util.go
@@ -37,37 +37,40 @@ func (pe ConfigParseError) Unwrap() error {
 	return pe.err
 }
 
-// toCaseInsensitiveValue checks if the value is a  map;
-// if so, create a copy and lower-case the keys recursively.
-func toCaseInsensitiveValue(value any) any {
+// CopyMap returns a deep copy of a map[any]any or map[string]any.  If value is
+// not one of those map types, then it is returned as-is.  If preserveCase is
+// false, then all keys will be converted to lower-case in the copy that is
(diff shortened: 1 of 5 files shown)


Noksa:
Any chance to get this merged?

PR #1725, opened 2023-12-21
feat: make env variables accessible

cfabianski:
Make Env Variables publicly accessible.

I used this new function to include the environment variables automatically in the doc https://github.com/Bearer/bearer/pull/1446


fmontorsi-equinix:
I'd love to have this merged btw.
I think it would be so useful to provide automatic, always in-sync documentation about all environment variables being used by applications...

PR #1735, opened 2024-01-09
feat: add errors.Is support to all errors

gwvandesteeg:
Add functionality to support errors.Is on all generated errors to keep in line with best practice on checking whether an error is of the specified type as per changes to error handling in go1.13.

This allows code to go from

```
if _, ok := err.(viper.ConfigFileNotFoundError); !ok {
  // do things
}
```
to
```
if !errors.Is(err, viper.ConfigFileNotFoundError{}) {
  // do things
}
```

### DIFF
--- a/internal/encoding/error_test.go
+++ b/internal/encoding/error_test.go
@@ -1,0 +1,17 @@
+package encoding
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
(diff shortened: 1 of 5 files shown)


gwvandesteeg:
This resolves #1736

andig:
Imho `errors.Is()` already does this. The real discrepancy here is checking `*Error` vs `Error`.

Issue #1752, opened 2024-02-06
feat: the order of processing aliases.


EwvwGeN:
### Preflight Checklist

- [X] I have searched the [issue tracker](https://www.github.com/spf13/viper/issues) for an issue that matches the one I want to file, without success.

### Problem Description

If I need to read data first from one field and only then read it from an alias, then there is no such functionality. Although it makes sense that if, for example, we have an Employee named Bob, I first want to search for all information about Bob from him and only then refer to his Employee object.

I asked a question on stackoverflow, but then I found out that there is no native solution in the package. (https://stackoverflow.com/questions/77934046/golang-viper-how-to-take-a-value-from-another-field-if-the-first-one-does-not-e)

### Proposed Solution

Add settings that will allow you to first search by the key itself, and only then access the aliases.

PR #1753, opened 2024-02-06
feat: new settings for viper, edit processing of an aliases

EwvwGeN:
Solution for https://github.com/spf13/viper/issues/1752


Pigolitsyn (approved):
good job

PR #1754, opened 2024-02-07
fix: search config file error

amone-bit:
There have two config files that have same file name and different extention.
image
And their contents follow.
image
image
This is my code.That will produce a problem.I expect the file named "key_map.toml", but viper read the file named "key_map.json"
image
 And that were some error information that viper gived.
image
So I update the "file.go" to fix the problem.And now work good.
image

### DIFF
--- a/file.go
+++ b/file.go
@@ -27,8 +27,9 @@ func (v *Viper) findConfigFile() (string, error) {
 func (v *Viper) searchInPath(in string) (filename string) {
 	v.logger.Debug("searching for config in path", "path", in)
 	if v.configType != "" {
-		if b, _ := exists(v.fs, filepath.Join(in, v.configName)); b {
-			return filepath.Join(in, v.configName)
+		if b, _ := exists(v.fs, filepath.Join(in, v.configName+"."+v.configType)); b {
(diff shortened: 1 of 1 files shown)


sagikazarmark:
There is a lot of confusion around how file searching works in Viper right now.

Technically, config type was not supposed to be used when looking for a file, only for encoding.

The problem with this PR is that it breaks loading files without extensions.

Here is a test that breaks with this PR:

```go
	t.Run("find file without extension", func(t *testing.T) {
		fs := afero.NewMemMapFs()

		err := fs.Mkdir(testutil.AbsFilePath(t, "/etc/viper"), 0o777)
		require.NoError(t, err)

		file, err := fs.Create(testutil.AbsFilePath(t, "/etc/viper/config"))
		require.NoError(t, err)

		_, err = file.WriteString(`key: value`)
		require.NoError(t, err)

		file.Close()

		v := New()

		v.SetFs(fs)
		v.AddConfigPath("/etc/viper")
		v.SetConfigType("yaml")

		err = v.ReadInConfig()
		require.NoError(t, err)

		assert.Equal(t, "value", v.Get("key"))
	})
```

I opened #1795 to start brainstorming on a better solution.

PR #1758, opened 2024-02-11
Update links to Golang Modules documentation

tobb10001:
I noticed that the link to the Golang Modules documentation leads to a site that says

> The Go wiki on GitHub has moved to go.dev (#61940).

See: https://github.com/golang/go/wiki/Modules

This PR replaces the link in the README and TROUBLESHOOTING files with the one that's suggested on the old site.


sagikazarmark (approved):
Thanks!

Issue #1732, opened 2024-01-06
AutomaticEnv does not override Unmarshaled config slice values


lwlee2608:
### Preflight Checklist

- [X] I have searched the [issue tracker](https://www.github.com/spf13/viper/issues) for an issue that matches the one I want to file, without success.
- [X] I am not looking for support or already pursued the available [support channels](https://github.com/spf13/viper/issues/new/choose) without success.
- [X] I have checked the [troubleshooting guide](https://github.com/spf13/viper/blob/master/TROUBLESHOOTING.md) for my problem, without success.

### Viper Version

1.18.2

### Go Version

1.21.1

### Config Source

Files

### Format

YAML

### Code reproducing the issue

```go
var yamlSimpleSlice = []byte(`
name: Steve
port: 8080
modes:
  - 1
  - 2
  - 3
clients:
  - name: foo
  - name: bar
`)

func TestSliceIndexAutomaticEnv(t *testing.T) {
	v.SetConfigType("yaml")
	r := strings.NewReader(string(yamlSimpleSlice))

	type ClientConfig struct {
		Name string
	}

	type Configuration struct {
		Port    int
		Name    string
		Modes   []int
		Clients []ClientConfig
	}

	// Read yaml as default value
	err := v.unmarshalReader(r, v.config)
	require.NoError(t, err)

	assert.Equal(t, "Steve", v.GetString("name"))
	assert.Equal(t, 8080, v.GetInt("port"))
	assert.Equal(t, "foo", v.GetString("clients.0.name"))
	assert.Equal(t, "bar", v.GetString("clients.1.name"))
	assert.Equal(t, []int{1, 2, 3}, v.GetIntSlice("modes"))

	// Override with env variable
	t.Setenv("NAME", "Steven")
	t.Setenv("MODES_2", "300")
	t.Setenv("CLIENTS_1_NAME", "baz")

	SetEnvKeyReplacer(strings.NewReplacer(".", "_"))
	AutomaticEnv()

	// Unmarshal into struct
	var config Configuration
	v.Unmarshal(&config)

	assert.Equal(t, "Steven", config.Name)
	assert.Equal(t, 8080, config.Port)
	assert.Equal(t, []int{1, 2, 300}, config.Modes)
	assert.Equal(t, "foo", config.Clients[0].Name)
	assert.Equal(t, "baz", config.Clients[1].Name)
}
```

### Expected Behavior

[...]

PR #1785, opened 2024-03-18
Add support for slices when using environement variables

TheoBrigitte:
Fixes: https://github.com/spf13/viper/issues/1732

Following up on @lwlee2608 great work from https://github.com/spf13/viper/pull/1733, here is a PR which adds support for using slices with environment variables and provide an additional fix to ensure we are not limited by the original config slice length.

Here is an example yaml config file :
```yaml
clients:
  - name: foo
  - name: bar
  - name: baz
```

Values can now be overriden using
```
CLIENTS_0_NAME=one
CLIENTS_1_NAME=two
CLIENTS_2_NAME=three
```

Note that for this solution to work the config must be initialized as folow:
```go

	var yamlConfig = []byte(`
clients:
  - name: foo
  - name: bar
  - name: baz
`)

	viper.SetConfigType("yaml")
	r := strings.NewReader(string(yamlConfig))
	viper.ReadConfig(r)
	SetEnvKeyReplacer(strings.NewReplacer(".", "_"))
	AutomaticEnv()
	// Configuration must have the same structure as yamlConfig
	var config Configuration
	v.Unmarshal(&config)

```

### DIFF
--- a/viper.go
+++ b/viper.go
@@ -1149,19 +1149,11 @@ func unmarshalPostProcess(input any, opts ...DecoderConfigOption) error {
 		return err
 	}
 
-	v.postProcessingSliceFields(map[string]bool{}, structKeyMap, "")
+	v.postProcessingSliceFields(structKeyMap, "")
 	return nil
 }
 
-// TODO remove shadow
-func (v *Viper) postProcessingSliceFields(shadow map[string]bool, m map[string]any, prefix string) map[string]bool {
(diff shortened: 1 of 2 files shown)


sagikazarmark:
Thanks @TheoBrigitte ! I'm gonna need some time to review this, but it looks good.

kadirakinkorkunc:
any news on this topic? @sagikazarmark

janekolszak:
@sagikazarmark Maybe it's time to merge?

PR #1819, opened 2024-04-30
fix: typo

hakusai22:
fix: typo


sagikazarmark:
AFAIK both spelling (one or two ‘l’s) are correct.

PR #1840, opened 2024-05-24
fix: search file by supported extension use manually specific config …

AuZhoomLee:
this PR is aimed to resolve issue #1754 and #1401, with compatible changes about `searchInPath`

test all imaginable situation passed and statement coverage is 100%:

```
=== RUN   Test_searchInPath
=== RUN   Test_searchInPath/find_file_without_extension
=== RUN   Test_searchInPath/cannot_find_file_with_extension
=== RUN   Test_searchInPath/find_file_with_supported_extension
=== RUN   Test_searchInPath/find_file_with_specific_extension_from_multiple_supported_files
=== RUN   Test_searchInPath/find_file_with_unsupported_extension
--- PASS: Test_searchInPath (0.00s)
    --- PASS: Test_searchInPath/find_file_without_extension (0.00s)
    --- PASS: Test_searchInPath/cannot_find_file_with_extension (0.00s)
    --- PASS: Test_searchInPath/find_file_with_supported_extension (0.00s)
    --- PASS: Test_searchInPath/find_file_with_specific_extension_from_multiple_supported_files (0.00s)
    --- PASS: Test_searchInPath/find_file_with_unsupported_extension (0.00s)
PASS
coverage: 17.1% of statements in ./...
```

### DIFF
--- a/file.go
+++ b/file.go
@@ -29,9 +29,22 @@ func (v *Viper) searchInPath(in string) (filename string) {
 	for _, ext := range SupportedExts {
 		v.logger.Debug("checking if file exists", "file", filepath.Join(in, v.configName+"."+ext))
 		if b, _ := exists(v.fs, filepath.Join(in, v.configName+"."+ext)); b {
-			v.logger.Debug("found file", "file", filepath.Join(in, v.configName+"."+ext))
-			return filepath.Join(in, v.configName+"."+ext)
+			// record the first found
+			if len(filename) < 1 {
+				filename = filepath.Join(in, v.configName+"."+ext)
+			}
+			// if specific configType are same with the current extension type, then return current
+			if v.configType == ext {
+				filename = filepath.Join(in, v.configName+"."+ext)
+				break
+			}
 		}
+	}
+
+	// return only if file exists
+	if len(filename) > 0 {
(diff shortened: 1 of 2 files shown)


AuZhoomLee:
#1849 is perfer good design, BUT the orginal function of `SetConfigType` means specific the type of file config, it's not functional as the name and comments mention, which is:
```
// SetConfigType sets the type of the configuration returned by the
// remote source, e.g. "json".
``` 

why not let it work as expectly? it's now works in some way, like if no other supported file name exist in the current file path.

em... , I think it's more confused maintainer this issue of this function.

SO, I want to 2 points:

1.  the change I committed realise worked seems no other issue, or can u give me another unexpectd example instance?
2.  if the function have certainly issue about this, why don't mark it Depracated, and supply more poper way to replace this.

sagikazarmark:
`SetConfigType` was supposed to override the detected file type (or set it for remote kv stores). It wasn't supposed to be used for configuring file search.

So when you say it's the expected functionality, it's not exactly true, although I understand why people would expect that.

I don't have an exact example that would break, but given the complexity (and past experiences) it could easily break for someone out there. Then what?

Let me think about it. But right now, I'd prefer deprecating the old API and drive people towards the new one.

PR #1854, opened 2024-06-04
Add experimental flag for bind struct

### LINKED ISSUE #1721: Enable experimental features runtime
### Preflight Checklist

### Problem Description

Consumers may want to enable experimental features runtime. Although it may not always be possible, it should be doable for the most part.

### Proposed Solution

Introduce a new API to enable feature flags in Viper.

Something like:

```go
v := viper.New(viper.Enable("bind-struct"))
```

The nice thing about this API is that it's generic and it doesn't necessarily break once an experiment is rolled out as a fully functional feature.

The counterpart of `Enable` should be `Disable`, so users can go back to the old behavior while it's still available.

### Alternatives Considered

Other potential names:

- `EnableFlag`/`DisableFlag`
- `EnableExperiment`/`DisableExperiment`
- `Enable(Experimental)Feature`/`Disable(Experimental)Feature`

### Additional Information

Came up in https://github.com/spf13/viper/issues/1706#issuecomment-1858515373

### LINKED ISSUE #1851: Make features.BindStruct a Variable or an Option, Instead of Build Constraints
### Preflight Checklist

### Problem Description

The features.BindStruct feature flag (which is also the only one feature flag in this package) is currently implemented as a build constraint here:
https://github.com/spf13/viper/blob/272344e4263af5a4463828db9e1862b6a35d812d/internal/features/bind_struct.go#L1-L5

Being implemented as a build constraint, users have to add a whole new `viper_bind_struct` build tag to their build scripts, in order to enable this feature.
This was quite surprising (to me, at least), and I think it's making harder for people to use this feature.

### Proposed Solution

Option 1: Make this feature flag a dynamically configurable variable instead of a const regulated by build constraints, and move the `features` package outside of the `internal` package.

```go
package features

var BindStruct = false
```

This way, users of viper can dynamically enable this feature from normal golang source code, like as follows:

```go
import "github.com/spf13/viper/features" // Just an example new package path

func main() {
    features.BindStruct = true // Enable the feature

    // Use the feature...
    var c MyConfig
    viper.AutomaticEnv()
    viper.Unmarshal(&c)
}
```

Option 2: Make this to an option that you can pass to `viper.NewWithOptions`, or something that you can configure later with viper instances.
I believe this was the intention of this original TODO comment here: https://github.com/spf13/viper/blob/272344e4263af5a4463828db9e1862b6a35d812d/viper.go#L979

```go
func main() {
    v := viper.NewWithOptions(viper.BindStruct()) // Enable the feature (just an example option name)

    // Use the feature...
    var c MyConfig
    v.AutomaticEnv()
    v.Unmarshal(&c)
}
```

or something similar to `AutomaticEnv()` usage:

```go
func main() {
    var c MyConfig
    viper.BindStruct() // Enable the feature (just an example option name)
    viper.AutomaticEnv()
    viper.Unmarshal(&c)

    // or to a specific viper instance
    v := viper.New()
    v.BindStruct()
    v.AutomaticEnv()
    v.Unmarshal(&c)
}
```

### Alternatives Considered

_No response_

### Additional Information

Related issues and PRs:
- https://github.com/spf13/viper/pull/1429
- https://github.com/spf13/viper/issues/1797
- https://github.com/spf13/viper/pull/1720 It seems this build constraint was introduced to address https://github.com/spf13/viper/issues/1706.
- https://github.com/spf13/viper/issues/1721 If this issue is resolved, then issue/1721 may not be needed (as suggested in the first comment)

sagikazarmark:
Fixes #1851
Fixes #1721

### DIFF
--- a/experimental.go
+++ b/experimental.go
@@ -3,6 +3,6 @@ package viper
 // ExperimentalBindStruct tells Viper to use the new bind struct feature.
 func ExperimentalBindStruct() Option {
 	return optionFunc(func(v *Viper) {
-		v.experimentalFinder = true
+		v.experimentalBindStruct = true
 	})
 }
--- a/experimental.go
+++ b/experimental.go
@@ -1,0 +1,8 @@
+package viper
+
+// ExperimentalBindStruct tells Viper to use the new bind struct feature.
+func ExperimentalBindStruct() Option {
+	return optionFunc(func(v *Viper) {
+		v.experimentalFinder = true
+	})
+}
--- a/viper.go
+++ b/viper.go
@@ -195,7 +195,8 @@ type Viper struct {
 	encoderRegistry *encoding.EncoderRegistry
 	decoderRegistry *encoding.DecoderRegistry
 
-	experimentalFinder bool
+	experimentalFinder     bool
+	experimentalBindStruct bool
 }
 
 // New returns an initialized Viper instance.
@@ -219,6 +220,7 @@ func New() *Viper {
 	v.resetEncoding()
 
 	v.experimentalFinder = features.Finder
+	v.experimentalBindStruct = features.BindStruct
 
 	return v
 }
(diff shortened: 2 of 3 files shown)


andig:
Love it- much simpler to use than the compile switches. Doesn't this also replace https://github.com/spf13/viper/pull/1715? Wondering why that's not visible in the diff?

sagikazarmark:
The build tag didn't go away. You can still set it which means it's going to be enabled by default for every Viper instance. Or you can enable it on a case-by-case basis using the option.

haoming29:
Is there a way to enable this feature with the global instance other than doing the build flag? Thanks!

PR #1861, opened 2024-06-06
Revert to go-difflib and go-spew releases

skitt:
Now that consul/api is gone from the module dependency tree, these dependencies can be reverted to their latest releases; the changes since then are only CI-related or documentation updates, which aren't relevant for downstream users.

(This is very minor but avoids questions like
https://github.com/kubernetes/kubernetes/pull/123339#discussion_r1618125911 about the use of untagged dependencies.)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -19,9 +19,9 @@ require (
 )
 
 require (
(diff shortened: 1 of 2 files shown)


sagikazarmark (approved):

sagikazarmark:
Thanks!

PR #1867, opened 2024-06-20
Encoding

sagikazarmark:
Use external encoding library


sagikazarmark:
We should probably merge #1869 first

PR #1869, opened 2024-06-21
New encoding layer

sagikazarmark:
Expose a new encoding abstraction layer. It allows users to bring their own encoding logic into Viper.

It also allows moving some of the less used formats (HCL, INI, Java properties) out of the core.

### DIFF
--- a/encoding.go
+++ b/encoding.go
@@ -1,6 +1,9 @@
 package viper
 
 import (
+	"strings"
+	"sync"
+
 	"github.com/spf13/viper/internal/encoding/dotenv"
 	"github.com/spf13/viper/internal/encoding/hcl"
 	"github.com/spf13/viper/internal/encoding/ini"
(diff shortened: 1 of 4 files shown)


sagikazarmark:
@skitt @andig could use a pair of extra eyes on this.

encoding.go
```diff
@@ -0,0 +1,113 @@
...
+
+// Codec combines [Encoder] and [Decoder] interfaces.
+type Codec interface {
+	Encoder
+	Decoder
+}
+
+// EncoderRegistry returns an [Encoder] for a given format.
+// The second return value is false if no [Encoder] is registered for the format.
+type EncoderRegistry interface {
```
sagikazarmark, line 33: Alternate names:

- `Encoders`/`Decoders`/`Codecs`
-

andig: The name sounds fine when the registries are separate.

viper.go
```diff
@@ -1715,7 +1723,12 @@ func (v *Viper) unmarshalReader(in io.Reader, c map[string]any) error {
 
 	switch format := strings.ToLower(v.getConfigType()); format {
 	case "yaml", "yml", "json", "toml", "hcl", "tfvars", "ini", "properties", "props", "prop", "dotenv", "env":
-		err := v.decoderRegistry.Decode(format, buf.Bytes(), c)
+		decoder, ok := v.decoderRegistry2.Decoder(format)
+		if !ok {
+			return ConfigParseError{errors.New("decoder not found")}
+		}
+
+		err := decoder.Decode(buf.Bytes(), c)
```
andig, line 1731:
```suggestion
		if err := decoder.Decode(buf.Bytes(), c); err !=nil {
```

andig:
> It also allows moving some of the less used formats (HCL, INI, Java properties) out of the core.

@sagikazarmark this still does not remove the go.mod dependencies, or does it?

sagikazarmark:
@andig not yet. This is just the first step to still allow using those libraries once they are removed.

encoding.go
```diff
@@ -0,0 +1,139 @@
...
+
+type encodingError string
+
+func (e encodingError) Error() string {
+	return string(e)
+}
+
+const (
+	// ErrEncoderNotFound is returned when there is no encoder registered for a format.
+	ErrEncoderNotFound = encodingError("encoder not found for this format")
```
sagikazarmark, line 39: I wonder if this should be an interface instead. That way, implementations wouldn't have to rely on Viper as a dependency.

andig: I wouldn't have an issue depending on Viper if I implemented an encoder for it?

sagikazarmark: No, the encoder interface can be implemented implicitly.

andig: Of course. But isn‘t the error on the consumer side that would always be depending on Viper? Maybe I misunderstood the question, sorry for the noise.

sagikazarmark: I'm talking about the _implementation_.

sagikazarmark:
I think this is ready

PR #1870, opened 2024-06-24
Drop encoding formats

sagikazarmark:
This PR is a followup for #1869

Drop HCL, Java properties and INI formats from the core.

Use `NewCodecRegistry` to register custom encoders supporting those format.

Moved to this repo: https://github.com/go-viper/encoding

Fixes #1867


sagikazarmark:
For reference, all three encoding formats are below 10%.

Screenshot 2024-06-24 at 15 29 10

Issue #871, opened 2020-03-26
Viper.Sub() does not preserve the Key delimiter (options in the constructor)


bogdandrutu:
See https://github.com/spf13/viper/blob/master/viper.go#L775

Arguably "Reset()" should preserve the options from the constructor.

PR #1887, opened 2024-07-22
feat: copy keydelim from parent chart in viper.Sub()

obs-gh-alexlew:
Addresses https://github.com/spf13/viper/issues/871

Simply copies the `keyDelim` from the parent chart to the subchart when calling `viper.Sub()`. Since there's currently no way to change the `keyDelim` of a viper instance post-creation, this is the only way to ensure that the `sub` viper instance maintains the `keyDelim`.


sagikazarmark (approved):

NathanNam:
Great! When could we expect to see a new version with this change?

PR #1982, opened 2025-02-15
docs: added a link to the releases widget

operagxsasha:


sagikazarmark:
Thanks @operagxsasha , but I'm not sure this makes sense. The Go version refers to the minimum Go version required for the project, not the Viper version.

operagxsasha:
maybe it is better to remove it altog

Issue #1988, opened 2025-03-01
Bug in File Search


rishavch2104:
### Preflight Checklist

- [x] I have searched the [issue tracker](https://www.github.com/spf13/viper/issues) for an issue that matches the one I want to file, without success.
- [x] I am not looking for support or already pursued the available [support channels](https://github.com/spf13/viper/issues/new/choose) without success.
- [x] I have checked the [troubleshooting guide](https://github.com/spf13/viper/blob/master/TROUBLESHOOTING.md) for my problem, without success.

### Viper Version

1.19.0

### Go Version

1.23.5

### Config Source

Flags

### Format

YAML

### Code reproducing the issue

```go

```

### Expected Behavior

When searching if file exists, file path should only be searched with list of available extension and custom extension if provided in config.

### Actual Behavior

In line 86 in file.go , extension is not added while checking exists so incorrect files will be matched.

rishavch2104:
Fixed in #1874

rishavch2104:
@sagikazarmark can you please check this?

sagikazarmark:
Admittedly the "old" file search API is weird and not trivial to use. `SetConfigType` was never intended to be used as a file extension, rather an override and a way to set type information in cases where there is no file name at all.

The proposed fix would interfere with other supported scenarios though. For example, some people set the config name to `.env` and the type to `dotenv` to support reading from env files.

I would suggest checking out the newly released file search API in 1.20.0. At allows you to customize the entire file search process, while also gives you a much better default implementation.

PR #1990, opened 2025-03-01
Fix: checking for custom file extension was happening without file type

rishavch2104:
Fix for https://github.com/spf13/viper/issues/1988


sagikazarmark:
Setting the config type is not exactly for this.

Can you please try the new Finder API instead?

Issue #2041, opened 2025-07-30
viper can not find my config file correctly


27149chen:
### Preflight Checklist

- [x] I have searched the [issue tracker](https://www.github.com/spf13/viper/issues) for an issue that matches the one I want to file, without success.
- [x] I am not looking for support or already pursued the available [support channels](https://github.com/spf13/viper/issues/new/choose) without success.
- [x] I have checked the [troubleshooting guide](https://github.com/spf13/viper/blob/master/TROUBLESHOOTING.md) for my problem, without success.

### Viper Version

1.19.0

### Go Version

1.24

### Config Source

Files

### Format

YAML

### Code reproducing the issue

```go
viper.SetConfigName("agent")
viper.SetConfigType("yaml")
viper.AddConfigPath(".")
viper.AddConfigPath("/config/path")
if err := viper.ReadInConfig(); err != nil {
    return err
}
```

### Expected Behavior

viper can find my config file `/config/path/agent.yaml` successfully

### Actual Behavior

an error occurred:
`While parsing config: yaml: control characters are not allowed`

### Steps To Reproduce

I have a command tool called `agent`, and a config file `/config/path/agent.yaml`
1. use viper to load config file as code shown above
2. start it by running `./agent`
3. an error occurred

### Additional Information

My binary called `agent`, and my config file called `agent.yaml`, viper does not search the config file in the secondary path /config/path, instead, it treats the binary `agent` as a config file and try to parse it

related to https://github.com/spf13/viper/issues/1988

rentiansheng:
I also encountered this problem,  I think is bug , this code  https://github.com/spf13/viper/blob/master/file.go#L86 without type params.

messikiller:
> I also encountered this problem, I think is bug , this code https://github.com/spf13/viper/blob/master/file.go#L86 without type params.

It seems unnecessary to fix this issue.
```go
func (v *Viper) findConfigFile() (string, error) {
                ...
		if v.configType != "" {
			names = locafero.NameWithOptionalExtensions(v.configName, SupportedExts...)
		} else {
			names = locafero.NameWithExtensions(v.configName, SupportedExts...)
		}
                ...
}
```
The `v.configName` will always be appended with the default supported extensions to generate a new list of files to search for. In the end, Viper will still iterate through and search all supported extension names.

The only confusing part is that Viper prioritizes searching through the default list of supported extensions first, and only afterwards will it look for the `configName` with extension suffixes:

```go
[...]

PR #2043, opened 2025-08-08
fix: find config file without type params #2041

rentiansheng:


sagikazarmark:
The problem is, config type isn't just used for extensions, but unmarshaling as well. Having files without extension, but unmarshaling as a certain type is perfectly valid.

Can you take a look at the new finder API? I believe you would be able to reach the same results.

rentiansheng:
I know

PR #2050, opened 2025-08-23
chore: add a stale issue scheduled action

spacez320:
This uses the [stale GHA](https://github.com/actions/stale) to schedule automatic stale marking and closure of issues.

As-is, this is marked in "debug" mode to see what it would be doing for us to verify before turning it on for real.

Open to ideas about how to test this, but it looks like it will only ever execute off of the default branch (per their docs).


sagikazarmark (approved):
Thanks! ❤️

PR #2059, opened 2025-09-15
feat: turn stale issue management on

spacez320:
Looks like it's doing what it's supposed to and will start making issues as stale, somewhat conservatively, so I think we can turn it on: https://github.com/spf13/viper/actions/runs/17721172394/job/50353800256


spacez320:
@sagikazarmark Good to go.

PR #2084, opened 2025-10-14
Bug fix: Key delimiters are now case insensitive

### LINKED ISSUE #2069: key delimiter can't be uppercase
### Preflight Checklist

- [x] I have searched the [issue tracker](https://www.github.com/spf13/viper/issues) for an issue that matches the one I want to file, without success.
- [x] I am not looking for support or already pursued the available [support channels](https://github.com/spf13/viper/issues/new/choose) without success.
- [x] I have checked the [troubleshooting guide](https://github.com/spf13/viper/blob/master/TROUBLESHOOTING.md) for my problem, without success.

### Viper Version

1.18.2

### Go Version

1.24.6

### Config Source

Files

### Format

YAML

### Repl.it link

_No response_

```go
v := viper.NewWithOptions(viper.KeyDelimiter("<SEPARATOR>"))
```

### Expected Behavior

For map values, they will come in blank because of this line:

https://github.com/spf13/viper/blob/0525738e5e7fcf38e40258475d46c96b4648c263/viper.go#L1153

This code:
```
		path   = strings.Split(lcaseKey, v.keyDelim)
```
should be
```
		path   = strings.Split(lcaseKey, strings.ToLower(v.keyDelim)
```

I don't think I can open a PR against this repo, or I would just send one. Hopefully this is a simple enough patch to describe here.

### Actual Behavior

When I read the config file, and I read this:

```
severity:
  foo: warning
```
I don't get a value for `severity` from v.Unmarshal, because the split is splitting a lowercased string on a possibly-upper-case keyDelim.

See fix above: call `strings.ToLower()` on the `v.keyDelim` before splitting on it.

### Steps To Reproduce

_No response_

### Additional Information

I need to support map-valued config keys whose subkeys contain dots even though that map is flat.

My workaround is that I'm using `·` (U+00B7) as a keyDelim instead of `<SEPARATOR>`. Because my map is flat, I just need something that nobody will ever actually want to use.

maxBRT:
Simply added the ``` strings.ToLower(v.keyDelim)) ``` as suggested and wrote a small test for it.

Fixes #2069

### DIFF
--- a/README.md
+++ b/README.md
@@ -602,6 +602,7 @@ if err != nil {
 
 If you want to unmarshal configuration where the keys themselves contain `.`
 (the default key delimiter), you can change the delimiter.
+The delimiter is case-insensitive when using alphabetic characters.
 
 ```go
 v := viper.NewWithOptions(viper.KeyDelimiter("::"))


viper.go
```diff
@@ -1464,10 +1464,12 @@ func Set(key string, value any) { v.Set(key, value) }
 
 func (v *Viper) Set(key string, value any) {
 	// If alias passed in, then set the proper override
+
 	key = v.realKey(strings.ToLower(key))
 	value = toCaseInsensitiveValue(value)
 
-	path := strings.Split(key, v.keyDelim)
+	path := strings.Split(key, strings.ToLower(v.keyDelim))
```
spacez320: I believe we will need to consider more cases where the key delimiter is used to build a key path, such as in [SetDefault](https://github.com/spf13/viper/blob/master/viper.go#L1545) and possibly other places that try to ultimately add a value to config.

viper_test.go
```diff
@@ -2718,3 +2718,14 @@ func skipWindows(t *testing.T) {
 		t.Skip("Skip test on Windows")
 	}
 }
+
+// Test the Set method with key delimiter for case insenitivty
+func TestUpperCaseKeyDelimiter(t *testing.T) {
```
spacez320: Thanks for adding a test. Perhaps this may be moved closer to, or possibly combined with, the existing delimiter test functions:
- https://github.com/spf13/viper/blob/master/viper_test.go#L2535
- https://github.com/spf13/viper/blob/master/viper_test.go#L1749

It would be also nice to make sure that `v.Set("foozbar", "Foo Bar Baz")` as well, to stress the expectation that delimiters are lower-cased just like keys.

spacez320:
I plan on reviewing this soon. cc @sagikazarmark

viper.go
```diff
@@ -1917,11 +1918,13 @@ func (v *Viper) mergeFlatMap(shadow map[string]bool, m map[string]any) map[strin
 	// scan keys
 outer:
 	for k := range m {
-		path := strings.Split(k, v.keyDelim)
+		fmt.Println(k)
```
spacez320: Did you mean to leave this in?

spacez320:
Apologies for the delay, I do still mean to review this. For now I'll probably lay down some superficial comments, but I intend to do a proper thorough review next week.

viper.go
```diff
@@ -1917,11 +1918,13 @@ func (v *Viper) mergeFlatMap(shadow map[string]bool, m map[string]any) map[strin
...
+		path := strings.Split(k, strings.ToLower(v.keyDelim))
+		fmt.Println(path)
 		// scan intermediate paths
 		var parentKey string
 		for i := 1; i < len(path); i++ {
-			parentKey = strings.Join(path[0:i], v.keyDelim)
+			parentKey = strings.Join(path[0:i], strings.ToLower(v.keyDelim))
```
spacez320: We do this specific processing on the key delimeter in so many places. I know it's very simple, but it would be great to have a shared place in case we need to use the key delimieter other places (without remembering to lower the string).

EIther a shared function or literally lower-casing `v.keyDelim` as part of its definition.

viper_test.go
```diff
@@ -2533,50 +2533,132 @@ func TestUnmarshal_DotSeparatorBackwardCompatibility(t *testing.T) {
 // `)
 
 func TestKeyDelimiter(t *testing.T) {
-	v := NewWithOptions(KeyDelimiter("::"))
-	v.SetConfigType("yaml")
-	r := strings.NewReader(string(yamlExampleWithDot))
+	t.Run("KeyDelimiterYAMLAndUnmarshal", func(t *testing.T) {
```
spacez320, line 2536: Just for my own understanding, what is the purpose of wrapping under `t.Run`?

viper.go
maxBRT: I did not thank you.

viper.go
maxBRT: Thank you for taking the time to review this.

Turns out this was a simple oversight on my part.
By lowercasing the key delielimiter at its definition, we do not require any additional processing.

viper_test.go
maxBRT: Since I added subtests to func TestKeyDelimiter(t *testing.T), I decided to make the previously single test its own subtest as well.
I based this decision on how other test functions with multiple subtests are structured for consistency.
https://github.com/spf13/viper/blob/7de42545181ddc051c7463cacb800b76ab266422/viper_test.go#L174-L356

viper_test.go
```diff
@@ -2533,50 +2533,132 @@ func TestUnmarshal_DotSeparatorBackwardCompatibility(t *testing.T) {
...
+				},
+			},
+		}
 
-	v.SetDefault("charts::values", values)
+		v.SetDefault("charts::values", values)
 
-	assert.Equal(t, "leather", v.GetString("clothing::jacket"))
-	assert.Equal(t, "01/02/03", v.GetString("emails::steve@hacker.com::created"))
+		assert.Equal(t, "leather", v.GetString("clothing::jacket"))
```
spacez320, line 2560:
```suggestion
        // It retrieves existing values using the custom delimeter
		assert.Equal(t, "leather", v.GetString("clothing::jacket"))
```

viper_test.go
```diff
@@ -2533,50 +2533,132 @@ func TestUnmarshal_DotSeparatorBackwardCompatibility(t *testing.T) {
...
+		}
 
-	var actual config
+		var actual config
 
-	require.NoError(t, v.Unmarshal(&actual))
+		require.NoError(t, v.Unmarshal(&actual))
 
-	assert.Equal(t, expected, actual)
+		assert.Equal(t, expected, actual)
```
spacez320, line 2582:
```suggestion
        // It sets and unmarshals new values with the custom delimiter
		assert.Equal(t, expected, actual)
```

viper_test.go
```diff
@@ -2533,50 +2533,132 @@ func TestUnmarshal_DotSeparatorBackwardCompatibility(t *testing.T) {
...
+
+		got := v.Get("foo")
+		want := map[string]any{"bar": "Foo Bar Baz"}
+		assert.Equal(t, want, got)
+
+		v.Set("foozbar", "Foo Bar Baz")
+		got = v.Get("foo")
+		assert.Equal(t, want, got)
+
+		v.Set("baz", "Bazzz")
```
spacez320, line 2599: What are we testing here? That the delimiter is not used in some cases? Maybe we can add a comment of intent.

viper_test.go
```diff
@@ -2533,50 +2533,132 @@ func TestUnmarshal_DotSeparatorBackwardCompatibility(t *testing.T) {
...
+				"bar": 123,
+				"baz": 456,
+			},
+		}
+
+		v := NewWithOptions(KeyDelimiter("Z"))
+		shadow := make(map[string]bool)
+		shadow = v.flattenAndMergeMap(shadow, config, "")
+
+		assert.True(t, shadow["foozbar"])
```
spacez320:
```suggestion
		assert.True(t, shadow["fooZbar"])
```

viper_test.go
spacez320: You might need to fix spacing with these, GitHub inline editor didn't seem to do the right thing, sorry.

spacez320 (approved):
This looks good to me.

spacez320:
@sagikazarmark I believe this is ready for your consideration.

PR #2087, opened 2025-10-14
Use StringReplacer interface instead of strings.replacer

thespags:
There exists StringReplacer already, but it's not available on a specific viper instance, only the singleton instance.


thespags:
Hi @sagikazarmark, is this something non-controversial to merge? There's a `StringReplacer` interface created for viper, that's used for the singleton instance, but on the viper call itself it requires the struct `*strings.Replacer`, which is harder to implement a matcher for.

thespags:
Thanks for your other approval. Hi again @sagikazarmark , I was wondering if we could get this change in as well.

thespags:
closing for https://github.com/spf13/viper/pull/2112

PR #2099, opened 2025-12-30
chore: upgrade golangci-lint to 2.7.2, add modernize, address findings

scop:


internal/encoding/json/codec_test.go
```diff
@@ -1,10 +1,12 @@
-package json
+package json_test
```
scop, line 1: This could be hushed with a linter disable, but using a separate test package is considered a good practice, see the [testpackage linter](https://golangci-lint.run/docs/linters/configuration/#testpackage) (not currently enabled, but anyway).

PR #2112, opened 2026-03-27
Use StringReplacer interface instead of strings.replacer 

thespags:


thespags:
any chance for this to get in ?

